PR #1, opened 2026-07-12
ES-brand audit: canonicalize Spanish makes across all 6 kinds (192 records, find→verify→critic pass)

aaguiare:
Full canonicalization pass over every Spanish-origin make in the catalog — all 6 kinds, 192 model records audited record-by-record (coverage 192/192), plus two horizontal cross-make sweeps. Every fix flows through the curated input layer (`overrides/` + `spotchecks.yml`); `catalog/` and `dist/` are untouched, per AGENTS.md. **497 added lines, zero deletions of existing curation. `ruby scripts/lint_overrides.rb` green.**

---

## 1. Scope

**Vertical slices** (make × kind): SEAT + Cupra (car, van), Ebro (car), Santana (van), Derbi / Bultaco / Montesa / OSSA / Gas Gas / TRS / Sherco / Rieju / Silence / Velca / Nerva / Torrot / Stark / Ebroh / MITT (motorcycle), Derbi / Bultaco / Rieju / Sherco / Silence / Scutum / Motorhispania / Scoobic / Ebroh (moped), Irizar / Ayats / Beulas / Ferqui / Indcar / UNVI / Car-bus.net (bus), plus Spanish motorhome coachbuilders leaked into the car kind (Moncayo, PLA, Ilusion).

**Horizontal sweeps**: (a) ES model names hiding under wrong makes across the entire catalog (found `scania/irizar`, `iveco/sunrise`, `iveco/wing`); (b) make-level near-dupes + casing audit (found the SCUTUM/Silence and E-BROH/Ebroh unmerged rebrands, five casing pins).

Out-of-scope finds were **recorded but deliberately not fixed** to keep this PR reviewable — see §8.

## 2. Methodology (how these lines were produced and checked)

1. **Find** — 10 auditor agents, one per make-group/sweep, each with the full model records (availability, popularity, `xrefs.tan`), the repo's altitude rules, and web access for brand facts.
2. **Adversarial verify** — every finding re-derived by an independent skeptic agent instructed to refute: fact-check against primary sources, recompute data-loss from the catalog, validate YAML shape against each file's contract, and grep blast radius for every regex/styling token. Outcome over 194 findings: **178 confirmed, 15 revised** (comment corrections, mechanism swaps, one blast-radius catch), **1 refuted** (a proposed Velca fold — Tramontana and Tramontana S are genuinely separate models).
3. **Apply** — single writer, only confirmed/revised findings at high/medium confidence.
4. **Two independent critics on the final diff**:
[...]

PR #2, opened 2026-07-25
PR #0: parallel-curation guardrails — duplicate-key lint, ownership map, dataset shape audit

rameerez:
Foundation PR for two-maintainer curation. **This is what lifts the override freeze.** No data changes: every override value is byte-identical, verified by round-trip.

## The blocking guard: duplicate YAML keys destroy work silently

```ruby
YAML.safe_load("Honda:\n  A: X\nHonda:\n  B: Y\n")  # => {"Honda"=>{"B"=>"Y"}}
```

Two appended `Honda:` blocks means one block **vanishes** — lint green, build green, and renames fail silently by design (a key matching nothing is simply inert). `scripts/lint_curation.rb` walks the Psych AST, which is the only way to see a duplicate: by the time `safe_load` returns, the loser is already gone.

It also caught a live bug class on `main`: **a make block keyed by anything other than the make's canonical DISPLAY NAME is inert forever**, because the pipeline looks renames up by display name (`normalizer.rb`'s `@o.model_renames[make]`), not by slug. `Mercedes Benz:` slugifies correctly and still never matches. The check knows the catalog *lags* the override layer, so PR #1's pending casing pins (MITT/TRS/EBRO/UNVI) are accepted rather than flagged.

## What's here

| file | purpose |
|---|---|
| `OWNERSHIP.yml` + `scripts/gen_ownership.rb` | 860 makes, zero overlap, one make = one owner across all six kinds. Ties (`adler` 1\|1, `garia` 2\|2) are **pinned in the script, never computed** — Hash order from the catalog JSON could otherwise flip an owner between releases and both sides would edit one block |
| `scripts/lint_dataset.rb` + `data/name_shapes.yml` | name-shape audit of the published catalog, `legit`/`debt` buckets. `debt` counts may only go down; spotless ≡ empty. Two buckets because one lets you reach zero by relabelling junk as legitimate |
| `scripts/reorg_make_blocks.rb` | make blocks alphabetized so concurrent appends land in different places. **Refuses to write unless the parsed YAML is identical before and after** |
| `scripts/lint_curation.rb` | duplicate keys, make-key reachability, added-line provenance |
| `NAMING.md` | the naming canon: identity, evidence standards, kind hygiene, mechanism choice, definition of clean. Two-wheel sections stubbed for the other maintainer |

## The shape audit already earns its keep

Reusing the dataset's own publish rule as a shape test — of **1,061 bare-integer names, 890 are corroborated** by ≥2 sources in ≥2 countries (Peugeot 208, Abarth 595, Mercedes 1824 trucks), and the **171 thin ones are almost exactly the KBA shared-string leak**. That single evidence condition replaces hundreds of hand-written allowlist lines.

The drop-list effectiveness rule (no published make may match a drop entry under **either** transliteration convention) replaces ~69 hand-written absence spotchecks and covers entries nobody has written yet. It immediately confirms three silent failures older than the published catalog:

```
PÖSSL             → poessl             48 models  *** SILENT DROP FAILURE ***
BÜRSTNER          → buerstner          20 models  *** SILENT DROP FAILURE ***
NIESMANN+BISCHOFF → niesmann-bischoff  12 models  *** SILENT DROP FAILURE ***
MONCAYO           → moncayo             3 models  (pending a build — not a failure)
```

That pending/failure split uses the commit that last touched `dist/` as the cutoff, because `catalog/` is a monthly build output and legitimately lags merged overrides.

## Also

19 pre-existing uncommented rename lines annotated with documented reasons — AGENTS.md requires every override line to say why, but `renames.yml`, `models/aliases.yml` and `styling.yml` were parse-only, so the rule was unenforced on exactly the three files carrying the most decisions.

CI runs all five checks per PR (`fetch-depth: 0` for the diff-based provenance check). `lint_dataset.rb` is report-only until `debt` is empty.


rameerez:
Superseded by #4, which contains every commit from this branch plus the curation stacked on top. Verified before closing: #4's tree includes all of `s4w/pr0-lint-reorg` (`git log origin/s4w/pr0-lint-reorg ^s4w/de-heal-curation` is empty), and #4 merged green with build + lint passing. Closing rather than merging both to avoid a redundant merge commit on a file where ordering is load-bearing.

PR #3, opened 2026-07-25
Re-pin licenses (build fix), 2W rename rekey, former_ids, CI cache save-always

rameerez:
Data-repo half of the build-outage fix and the two-wheeler nameplate correction. Pairs with **vehiclesdb/vehiclesdb-pipeline#2** — merge the pipeline PR first or alongside; the rekeyed rename keys assume the new spacing rule.

## `data/licenses/` — the 13-day outage

Re-pinned via `rake licenses:pin` after migrating `ie_cso`, `my_jpj`, `ca_nrcan` (the three that broke the build) plus `uk_dft` and `nz_nzta` from `whole_file` to requester-invariant `phrases`. **Terms were diffed before re-pinning and are substantively unchanged** — still CC BY 4.0 / OGL-Canada, zero SA/NC/ND clauses.

The pinned `.txt` artifacts shrink by ~2,545 lines because they now hold the load-bearing sentences instead of whole government pages. `my_jpj` is the instructive one: its page prints `Data as of <month>`, so a `whole_file` pin there breaks **every month by construction**.

## `overrides/models/renames.yml` — 16 keys the spacing change orphaned

The caser-before-renames trap, in reverse. 10 rekeyed to their post-spacing form (`COTA4RT260` → `"Cota 4RT260"`, `MULHACEN125` → `"Mulhacen 125"`, `AVENTURA500` → deleted, …); 6 deleted because the spacing rule now produces the target natively — **with their source URLs preserved in a comment** so the facts survive the deletion.

A new pipeline test (`test_override_key_reachability.rb`) asserts key reachability every build, so this class cannot regress silently again. It also found a pre-existing dead key: `Mercedes-Benz "280 SE"` has never matched anything.

## `overrides/models/former_ids.yml` (NEW) — 1,050 migration aliases

SCHEMA.md §Versioning calls renaming a published id the one breaking change, needing "a major schema bump plus a migration alias". This is that alias, so ~1,050 corrections stay inside the *additive* growth contract.

Generated by `scripts/propose_former_ids.rb`, which proposes only mechanical pairs and **refuses to map an id that still exists**: `car/audi/89` survives (the Netherlands genuinely registers the B3 Audi 80 as "89") and must never be claimed as an alias of `car/audi/a3`.

The 97 ids that vanished with no successor are deliberately **not** here — they are the Bürstner/Pössl/Niesmann+Bischoff motorhome floorplan codes that only ever published because the drop matcher couldn't see through German transliteration. Those are removals, not renames, and belong in CHANGELOG.md.

## `.github/workflows/monthly-build.yml` — `save-always` on the source cache

During the outage the repo held exactly **one** cache entry (319.57 MiB, created 07-06) that every failing run restored but none refreshed — `actions/cache` skips its save step on job failure. So each retry re-downloaded everything published after 07-06, while the entry drifted toward GitHub's 7-day eviction with the next weekly run sitting exactly on that boundary.


rameerez:
## Context: this PR unblocked everything else, and one of its findings is still open

This is the PR that ended a **13-day red build**. Root cause: three `whole_file`
licence pins tripping on cosmetic page churn (`ie_cso`, `my_jpj`, `ca_nrcan`). Terms
were diffed **before** any pin was touched — substantively unchanged, **zero
SA/NC/ND clauses**, so the CC-BY composite was never at risk.

**The generalizable finding:** `my_jpj`'s page prints `Data as of <month>`, so a
`whole_file` pin there **breaks every single month by construction**. Any pin whose
target page carries a date is a time bomb on a monthly cadence. And phrases must be
**tested**, not assumed — OGL v3's *"You are free to: copy, publish, distribute and
transmit the Information"* does not match as a substring on the National Archives
page, because it is split across markup.

**⚠️ STILL OPEN, from this PR's own investigation:** `be_fps` is pinned but **not
ingested**, and its page contains **no licence text at all** — "licence", "Creative
Commons" and "CC0" are absent from 22,607 chars (it is a JS shell). That pin guards
nothing, and a green gate there is **not** evidence Belgian terms were ever checked.
Re-point it before Belgium is ingested.

Also from this PR: the `pipeline-failure` label did not exist, so
`gh issue create --label pipeline-failure` errored and **three failed builds paged
nobody**. If alerting ever goes quiet again, check the label first.

**Full brief:** PR #4, first comment (data side) and `vehiclesdb-pipeline` PR #1,
first comment (code side).

rameerez:
## Onboarding pointer

This PR is the **data-repo half of the build-outage fix**. Full context for everything in it — and every gotcha it created — is in the S2W long-form brief:

What a future maintainer most needs from this PR specifically:

- **The `extract` mode lives in `data/licenses/pins.json` (this repo), not the Rakefile.** The Rakefile's `PINS` only feeds `rake licenses:pin`. So a licence-gate fix is a *data-repo* change — which is why any data PR lacking it fails the build, and why PR #2 showed a red build that was nothing to do with its own contents.
- **`my_jpj` on `whole_file` mode breaks EVERY MONTH by construction** — the page prints the dataset's own freshness date (`Data as of 31 May 2026` → `30 Jun 2026`). That is the canonical argument for `phrases` mode.
- **Verify every phrase against a real fetch before pinning it.** OGL v3's headline sentence does *not* match as a substring on the National Archives page — it is split across markup, and `license_extract` collapses whitespace but does not strip HTML. A phrase spanning a tag boundary silently never matches and the pin becomes vacuous.
- **Always diff the terms before re-pinning.** All three were verified substantively unchanged with zero SA/NC/ND clauses first. One ShareAlike ingredient infects the whole CC-BY composite (AGENTS.md), so this step is not optional even when the diff "looks cosmetic".
- The 16 rekeyed rename keys here are the **caser-before-renames trap in reverse**: a normalizer change orphaned keys that had been working. Renames fail silently, so nothing complained. `pipeline/tests/test_override_key_reachability.rb` now guards the class.
- `save-always: true` on the source cache: `actions/cache` does **not** save on a failed job, so during the outage the repo held one cache entry from 07-06 that every failing run restored but none refreshed.

PR #4, opened 2026-07-25
Heal the German data + 390 duplicate-spelling merges + kind hygiene (stacked on #2)

rameerez:
Stacked on #2 (`s4w/pr0-lint-reorg`) — review that first. Companion to **vehiclesdb-pipeline#1** (the parser fix); this makes the recovered names *correct*.

## `moves.yml` — a new mechanism, and why it was unavoidable

Registers file sub-brands under the **type-approval holder**, not the marque, because "make" comes from a manufacturer key on the registration document. KBA assigns `Marke` from the HSN ([kba.de](https://www.kba.de/DE/Statistik/Fahrzeuge/Marken_Hersteller/markenHersteller_node.html)):

- Cupra has no HSN → every Cupra is filed under **HSN 7593 "SEAT"**. FZ 6.1 shows `Handelsname = 'CUPRA ATECA,ATECA'` under that key.
- Genesis has no HSN → GV60/GV70/G80/G90 ride **Hyundai's HSN 8252**, whose rows read `'GV60, GENESIS GV60'`.

Renaming can't fix that (the target lives under another make) and dropping throws away real evidence — **26,512 German registrations** for the Cupras alone. PR #1 chose dropping, for want of this mechanism.

## Retiring PR #1's SEAT block

`"468"`–`"474"` were dropped as "KBA numeric type-codes" on the Tesla-512 precedent. They were SEAT's entire current German lineup: `si[468..474]` = ATECA, BORN, FORMENTOR, IBIZA, LEON, TAVASCAN, TERRAMAR. The stated reason — *"seat/468 held DE rank 33 and polluted popularity"* — described **correct data**. A `null` rename beats a move by design, so these had to go in the same change as `moves.yml`.

## KBA naming corrections, each verified against primary sources

| KBA string | → | why |
|---|---|---|
| `"GLK, GLC"` | GLC | temporal merge; 16,653 registrations were landing on a nameplate dead since 2015 |
| `"ML-KLASSE, GLE"` / `"GL-KLASSE, GLS"` | GLE / GLS | same class |
| `RX6` | MG HS | **not a model** — KBA's type code (`'MG RX6;-HS;-EHS;-EHS PHEV'`) |
| `BZ4` | bZ4X | truncation, present since the 2023 workbook |
| `CX-6` | CX-6e | a **real new model**, not a CX-60 truncation — KBA lists CX-6/CX-60/CX-80 as three rows |
| `ES8` | EL8 | NIO renamed ES6/ES8 → EL6/EL8 for Europe |
| `1`/`3`/`5` (smart) | #1/#3/#5 | the hash is official |

Temporal merges are fixed **source-side**, because `normalizer.rb` splits on comma and keeps the *first* part — post-split, KBA's row is indistinguishable from a genuine historic GLK.

## 390 duplicate spellings merged

The same car published twice because spellings slug differently:

```
mercedes-benz/280-se  "280 Se"   4 sources
mercedes-benz/280se   "280SE"    4 sources    ← the same W108
```

632 such groups exist in this half (1,283 records), each splitting a model's own evidence. This applies the safe 388: one existing spelling is already correct, others merge into it. **632 groups / 1,283 records → 249 / 503.** The remaining 244 need a *new* canonical string (Mercedes writes "280 SE" — neither variant is right) and are out for source verification before they mint ids.

Surfaced two pre-existing errors, both corrected: Ford targeted `"F 150"` (Ford writes **F-150**), Toyota targeted `"Hi Ace"` (it's one word).

## Kind hygiene: 161 non-vehicles out of the car kind

14 motorhome builders, each **measured** against RDW's CC0 register rather than inferred: `KABE 79/79 kampeerwagen`, `CI 166/166`, `JOINT 254/254`, `SEA 186/186`, `BAVARIA 217/218`, `DREAMER 147/148`, `FORSTER 148/148`. Plus three compound forms that escaped existing entries (`KNAUS TABBERT` beside `KNAUS`) — the same escape route that kept PÖSSL published. 49 make-as-model rows dropped; `van/uaz/uaz` deliberately kept and tracked, because UAZ's naming is numeric type designations and the row's identity is unrecoverable.

`FENDT` is annotated rather than assumed: RDW's `merk=FENDT` rows are agricultural trailers, not caravans. Neither is a car, so the car-kind drop holds either way.

## Verification

Offline through the real normalizer: 399 KBA rows parse, **319 land on existing catalog ids**, top sellers heal to `volkswagen/golf` (36,508 YTD), `t-roc`, `tiguan`, `opel/corsa`, `skoda/octavia`. 18 spotcheck rows pin it, including the pair that states the whole fix: `seat/468` must NOT exist **and** `seat/ateca` must carry `de`.

Cumulative on this half: 10,561 → 10,400 records (490 merged, 107 non-cars, 54 junk).


rameerez:
# ONBOARDING BRIEF — the 2026-07-25 correction pass, in full

**Read this before touching the dataset.** It is written to be self-contained: it
assumes you have the repos and nothing else. No prior conversation, no other
document, no institutional memory. Where something lives in a file, the path is
given. Where a claim rests on evidence, the evidence is quoted or linked.

This documents a single day's work by **two parallel agent sessions** that
corrected identity errors across the dataset. It is deliberately long, because
the expensive part of this work was not the fixes — it was discovering *why* the
obvious fix was wrong, three separate times.

---

## PART 0 — THE ONE-PARAGRAPH VERSION

Two parser bugs were fabricating data. One published Germany's model names as
XLSX shared-string *indices* (`volkswagen/552` **was the Golf**); the other glued
two-wheeler names shut (`BANDIT1200`). A curation pass then began **deleting the
real records the bugs had produced**, on the reasonable-looking theory that
`seat/468`–`474` were registry junk — they were SEAT's entire current German
lineup. Separately, Germany was absent from **three entire kinds** (van/truck/bus)
because a drop rule promised a routing step that was never built. All of this is
now fixed, guarded by lint rules and spotchecks, and the reasoning is below.

---

## PART 1 — THE SYSTEM, FROM ZERO

### 1.1 The repos

| repo | visibility | what it is |
|---|---|---|
| `vehiclesdb/vehiclesdb` | public | the dataset. `catalog/` + `dist/` are BUILD OUTPUTS. `overrides/` + `spotchecks.yml` are the hand-curated INPUTS. |
| `vehiclesdb/vehiclesdb-pipeline` | private | the factory. Fetches 14 national registers, normalizes, reconciles, validates, emits into the public repo. |
| `vehiclesdb/vehicles` | public | the Ruby gem that bundles a `dist/` snapshot. |

**Critical mental model:** `catalog/` **lags** `overrides/`. The build runs monthly.
So an override merged today does not change the published data for weeks. Almost
every confusing measurement in this work traces back to that lag — see §5.1.

### 1.2 The data flow

```
14 official registers  →  per-source parser (pipeline/sources/*.rb)
  → normalizer (make aliases, drops, family rules, CASING, RENAMES, MOVES)
  → reconciler (publish when ≥2 sources agree, or 1 source above a count threshold)
  → validate (6 gates + the spotcheck panel)
  → emit into vehiclesdb/{catalog,dist,manifest.json}
```

### 1.3 Pipeline order, and why it decides what is even possible

```
make aliases/drops → model drop_patterns → normalization (family rules, prefix
strip) → CASING (styling.yml) → RENAMES (renames.yml) → MOVES (moves.yml)
→ body types, aliases, popularity
```

Consequences you will trip over if you forget them:

- **The caser runs BEFORE renames.** A rename key must match the *post-cased*
  string. Adding an acronym token silently re-cases other records' rename keys and
  orphans them.
- **Renames run before moves.** A `null` rename therefore BEATS a move (see §4.10).
- **Renames are make-scoped but KIND-BLIND.** A key written for `car` fires in
  every kind of that make.
- **Renames fail SILENTLY.** A key that matches nothing is simply inert. Nothing
  logs it. This is why absence spotchecks and the reachability test exist.

### 1.4 Who owns what (two-session split)

Curation is split by **MAKE, not by kind**: one make = one owner across all six
kinds. The owner is the side owning that make's dominant kind by record count.

- **s4w** — car, van, truck, bus (427 makes)
- **s2w** — motorcycle, moped (433 makes)

The map is **generated** into `OWNERSHIP.yml` by `scripts/gen_ownership.rb`, 860
makes, zero overlap. Ties (`adler` 1|1, `garia` 2|2) are **pinned in the script**,
never computed — Hash order from the catalog JSON could otherwise flip a make's
owner between releases, and two people would then edit one block.

**Consequence people get wrong:** each owner curates their makes **in every kind**,
so each side holds records inside the other's kinds (measured: 427 four-wheel
records under two-wheel-owned makes, 81 the other way). **"Done" is defined per
make-set, never per kind-set.**

### 1.5 Working protocol (adopt this if you run parallel sessions)

1. **Never work in the primary clones.** `~/GitHub/vehiclesdb` and
   `~/GitHub/vehiclesdb-pipeline` are read-only, for reading merged state. Each
   session works in its own `git worktree` on `s4w/…` / `s2w/…` branches. Without
   this, one session's `git checkout` changes the other's tree mid-edit.
2. **Append-only inside your own region; never reorder another's lines.** Make
   blocks are kept alphabetical by `scripts/reorg_make_blocks.rb` so concurrent
   edits land in different places.
3. **No force-push anywhere**, including `main`.
4. **Cross-session adversarial review before merge.** This caught a build-killing
   raise, a data-losing truncation, and a wrong diagnosis that would have
   discarded 79,569 registrations. None of the three was found by its author.
5. **Pipeline repo is partitioned by FILE**, not by make. Rough split used:
   `xlsx_lite.rb`/`overrides.rb`/`validate.rb`/car-only sources = s4w;
   `normalizer.rb`/`emit.rb` = s2w; `reconciler.rb`/`emit.rb`/multi-kind sources =
   announce before editing. `pipeline/tests/` = one new file per side.

---

## PART 2 — THE INCIDENTS, WITH MECHANISM AND LESSON

Every one of these shipped or nearly shipped. They are ordered by how much damage
they did or would have done.

### 2.1 The XLSX cell regex that published shared-string indices as model names

**Mechanism.** `pipeline/lib/xlsx_lite.rb` scanned cells with:

```ruby
%r{<c r="([A-Z]+)\d+"([^>]*)>(?:.*?<v>(.*?)</v>)?}m
```

Two flaws compounding: `([^>]*)>` cannot distinguish a **self-closing** (empty)
cell `<c r="B23" s="73"/>` from an opening tag, and the value group is lazy +
multiline with **no boundary**, so it walked forward past `</c>` into the next
cell and stole its `<v>`. Because the empty cell carries no `t="s"`, the
shared-string **index** was then emitted as literal text instead of being resolved.

KBA FZ 10 is the only xlsx source with empty cells mid-row. Its real layout on
continuation rows is `B` empty + `C`=Modellreihe — **the inverse of what the
source file's own header comment claimed**, which is a large part of why this
survived review for a month.

```xml
<c r="B22" s="71" t="s"><v>87</v></c><c r="C22" s="42" t="s"><v>79</v></c>  ← AUDI / A1
<c r="B23" s="73"/>                  <c r="C23" s="41" t="s"><v>89</v></c>  ← ∅    / A3
```

**Damage.** 148 records across 29 makes carried indices instead of names.
`volkswagen/552` was the **Golf** — Germany's best-selling car — while
`volkswagen/golf` carried twelve countries and no `de`. Every German popularity
figure for the country's best sellers sat on a garbage id, and the real nameplates
were ranked as if Germany did not exist. Germany's published top-10 was 8/10 junk,
and its #1 was `zeekr/neuzulassungen-insgesamt` — the table's **grand-total row**.

**Fix.** Cell-bounded scan (`/>` alternative tried before `>`, `<v>` read only
inside its own cell), inline-string support, and `parse_sheet` /
`parse_shared_strings` split out as pure functions so the reader is testable
without building a zip. Plus fail-loud `audit!` gates in `de_kba_fz10.rb` on make
count, row count, and **bare-integer share** — the exact regression signature.

**There was zero xlsx test coverage.** There are now 21 tests, hermetic (XML built
inline) plus 2 integration tests against a cached snapshot.

### 2.2 The cleanup that deleted SEAT's German lineup

A previous merged PR dropped `seat/468`–`474` as *"KBA numeric type-codes, not
nameplates (Tesla 512 precedent)"*, with the stated reason *"seat/468 held DE rank
33 and polluted popularity"*.

They were not type-codes. `si[468..474]` = **ATECA, BORN, FORMENTOR, IBIZA, LEON,
TAVASCAN, TERRAMAR** — SEAT's entire current German lineup. The "pollution" was
correct data: the Ateca really is a German top-40 seller.

**A second pass then proposed deleting the other 148 for the same reason.** It was
caught only by reading the parser.

**The rule this produced** (now `NAMING.md` §2.1, and the most important paragraph
in this brief):

> A model name that is a bare integer, single-source, is a **PARSER** suspect
> before it is a **DATA** suspect. Resolve it against that source's raw snapshot
> before proposing any drop. Consecutive integers under one make are presumptively
> a shared-string/index leak — i.e. evidence the records are REAL.
>
> **Recover the true name by RE-PARSING THE ROW, never by resolving the integer as
> a table index: shared-string tables are workbook-wide and will hand you a
> confident wrong answer from another sheet.**

The second half is paid for too. Index-resolution produced three confident wrong
answers: `si[350]="ACTROS"` (a truck, in a cars-only sheet), `si[500]="SUZUKI"` (a
make), `si[100]="Q6"` (while Audi 100 is also a real nameplate).

**Structural corroboration is a cheap triage filter, not proof.** A contiguous run
bracketed by structure — make header, alphabetical order, `ZUSAMMEN` terminator —
is good reason to go re-parse. It is never a reason to conclude.

### 2.3 Germany was absent from three entire kinds (79,534 registrations discarded)

**Measured on the published catalog:**

```
van   records with `de` evidence:   0 / 1,120
truck records with `de` evidence:   0 / 1,045
bus   records with `de` evidence:   0 /   385
car   records with `de` evidence: 216 / 8,363
```

**Mechanism.** `overrides/models/drop_patterns.yml` drops `CADDY`, `TRANSPORTER`,
`KANGOO` etc. from the `car` kind with the comment *"they belong to kind van, whose
build ingests them via kind_maps instead"* — but **no kind_map existed for
`de_kba_fz10` and the source declared only `car`.** The rows were dropped and
nothing picked them up. The comment described an intent the build never
implemented.

**Why those rows are in a passenger-car table at all** (this is the load-bearing
insight, and it generalizes): **"Personenkraftwagen" means EU class M1, and M1 is
defined purely by seat count (≤8 besides the driver) with NO mass ceiling.** So
FZ 10 legitimately contains:

- the **M1 passenger versions of vans** (Caddy Life, Multivan, Transit Custom
  Kombi), whose type approvals are *disjoint* from their N1 siblings — provable by
  comparing KBA FZ 4.2 (Pkw by Handelsname) against FZ 4.8 (Lkw);
- **motorhomes up to 26 tonnes**, as M1 special-purpose `SA/Wohnmobil` — 71,575
  units = 2.5% of German Pkw registrations in 2025.

Sources: KBA methodology
`https://www.kba.de/DE/Statistik/Fahrzeuge/fz_methodik/fz_methodische_erlaueterungen_202402_pdf.pdf`,
2007/46/EC Annex II (now Reg. (EU) 2018/858 Annex I), KBA SV 1 systematics.

**Fix.** `overrides/kind_maps/de_kba_fz10.yml` (new) routes 30 nameplates to `van`
via a `by_model:` map; `de_kba_fz10.rb` declares `kinds [car van]` and stamps each
Row's kind from it. Result: 369 car rows + 30 van rows carrying **79,534
registrations** — Ducato 16,040 · Transporter 15,042 · Transit Custom 8,438 ·
Caddy 7,914.

**NOT routed, deliberately:** rows that are ~100% motorhome (Actros, Arocs, Atego,
Unimog, Volvo FH, MAN TGL/TGM/TGS/TGX, Iveco Stralis/Eurocargo, DAF XF). **FZ 11.1
— the same population cut by SEGMENT instead of make — places 100% of those
registrations in segment `WOHNMOBILE`, matching FZ 10.1 to the unit.** They are
campers on a truck chassis, not trucks sold as trucks; routing them to `truck`
would assert a market that does not exist.

**Verification method worth reusing:** FZ 10.1 and FZ 11.1 reconcile 399/399
Modellreihen exactly, which makes the segment cut a reliable cross-check. Its
`UTILITIES` segment is literally *"the M1 versions of nameplates that also have N1
versions"* — i.e. KBA's own car/van boundary, available free.

### 2.4 KBA "temporal merges" landing 22k registrations on nameplates dead since 2015

KBA splices a nameplate to its renamed successor inside ONE Modellreihe cell to
keep its time series continuous. `Normalizer` splits on commas and keeps the
**FIRST** part (`normalizer.rb`, a rule written for RDW strings like
`"SX4, SUZUKI SX4, S"`), so:

```
"GLK, GLC"       → GLK   (GLK ended 2015; 16,653 registrations Jan–May 2026)
"ML-KLASSE, GLE" → ML    (M-Class renamed GLE April 2015; 4,921)
"GL-KLASSE, GLS" → Gl    (GL renamed GLS MY2016; 755)
```

**A rename in `overrides/` cannot fix this**: post-split, the KBA row is
indistinguishable from a genuine historic GLK row from another register, so a
`GLK → GLC` rename would rewrite real GLK cars. Hence a **source-level**
`COMMA_MERGED` table in `de_kba_fz10.rb`, plus a loud warning for any *unmapped*
comma row (KBA adds them over time).

`"ID.4, ID.5"` is **not** a temporal merge — both are on sale. It maps to `ID.4`
(the parent, higher-volume nameplate, and the one VW Group's own reporting names)
and is documented as a **lossy compromise**, not a canonicalization.

**Caveat before you "correct" a low count:** a German *Neuzulassung* can be the
first German registration of an **imported used car**, so trace volumes of dead
nameplates are real. Mercedes' `CLK` row shows 1 unit in May 2026 though the CLK
ended in 2010.

### 2.5 Registers file sub-brands under the TYPE-APPROVAL HOLDER → `moves.yml`

**Mechanism.** A register's make column is often derived from a manufacturer key on
the registration document, not from marketing. KBA assigns `Marke` from the **HSN**
(Herstellerschlüsselnummer):
`https://www.kba.de/DE/Statistik/Fahrzeuge/Marken_Hersteller/markenHersteller_node.html`

- **Cupra has no HSN** → every Cupra is filed under **HSN 7593 "SEAT"**. KBA FZ 6.1
  shows `Handelsname = 'CUPRA ATECA,ATECA'`, `'CUPRA LEON'` under that key.
- **Genesis has no HSN** → GV60/GV70/G80/G90 ride **Hyundai's HSN 8252**, whose
  rows read `'GV60, GENESIS GV60'`.
- **RDW does the same**: `merk=PIAGGIO` carries 130 distinct Vespa-badged names and
  **79,569 registrations**, because Piaggio is the approval holder.

Renaming cannot fix it (the target lives under a different make) and dropping
throws away real evidence. Hence `overrides/models/moves.yml`, a new mechanism:
`"Make|Model": "TargetMake|TargetModel"`. Now 58 entries — 46 Piaggio→Vespa, 5
SEAT→Cupra, 2 more SEAT→Cupra (Leon/Ateca), 4 Hyundai→Genesis, 1 Auto Union→Audi.

**Rules that each cost someone a red test:**

1. Keys are **POST-rename, POST-casing** pairs. `"Hyundai|GV70"`, never
   `"Hyundai|Gv70"`. A wrong-form key is silently inert.
2. A `null` rename **BEATS** a move by design. Retiring a null and adding a move
   must happen in the same change.
3. Only move when the register is unambiguously filing ONE marque's model under
   another. **Do NOT move a POOLED row.** KBA's `SEAT/LEON` and `SEAT/ATECA` pool
   the SEAT *and* Cupra versions of one nameplate; moving them would hand SEAT's
   own volume to Cupra. They stay with SEAT.

**Generalize this:** any register whose make column is approval-derived will file
sub-brands under the holder. Candidates to check before merging any "duplicate"
makes: Husqvarna/GasGas under KTM, Vespa/Aprilia/Moto Guzzi/Derbi under Piaggio.
**A "duplicate make" may actually be an approval holder, and merging it is the same
class of error as the SEAT deletions.**

### 2.6 The same car published twice — 632 duplicate-spelling groups

```
mercedes-benz/280-se  "280 Se"   4 sources
mercedes-benz/280se   "280SE"    4 sources   ← the same W108
```

Two registers spell it differently, the spellings slug apart, and the model's own
evidence splits across two ids so both look weaker than the car is. 632 groups /
1,283 records in the four-wheel half. Detector:
`scripts/find_duplicate_spellings.rb`.

Two risk classes come out of it:

- **canonical already exists** as one of the variants → a safe merge (388 groups).
- **canonical is a NEW string** (`"280 SE"` — neither variant is right) → mints an
  id, so verify against the marque's own archive first (244 groups).

Now down to ~140 groups (the long tail of one-off registry spellings).

### 2.7 The verified marque naming conventions

Each was checked against the marque's own archive, not encyclopaedias. **Cite these
rather than re-deriving them.**

**Mercedes-Benz** — cars are `<number> <UPPERCASE letters>`, with a **lowercase
series letter** where the archive uses one and a **hyphenated** valve-count suffix.
Trucks/buses invert to `<letter(s)> <number>`.

```
220 SEb   (the b is the W111/3 SERIES letter — lowercase)
300 SEL · 380 SEC · 300 TE · 300 TE 4MATIC (all caps, no hyphen)
300 CE-24 (hyphen, NOT "300 CE 24V")   ·   L 312 (truck: letter, space, number)
```
Source: `https://mercedes-benz-publicarchive.com/marsClassic/` — Mercedes' own
heritage archive. **Operational note:** it 403s ordinary fetchers but serves fine
to a browser User-Agent; `mercedes-benz.com` blocks both.
Two corrections to common belief: **L = lang and E = Einspritzmotor are
confirmed**, but **"S = Sonderklasse" is NOT what Mercedes says** — their material
attributes the S to 1949 and glosses it "Super or Special"; "S-Class" as a name
arrived in 1972. Do not put "Sonderklasse" in the dataset as fact.

**Lexus** — `<CODE> <NUMBER><suffix attached>`; powertrain suffixes **lowercase**
(h/e/d), body suffixes **uppercase** (C/L). `RX 450h`, `CT 200h`, `IS 250C`. **A
space before the suffix letter never occurs in any official Lexus material.**
Regional divergence is JDM-vs-everyone: lexus.com/.co.uk/.de all use the spaced
form; only lexus.jp closes it (`RX450h`). **The NAMEPLATE, at our altitude, is the
LINE** (exactly like BMW's "3 Series") — so 162 engine/powertrain variants fold and
**Lexus went from 180 published "models" to 25 nameplates**.
Traps: US spec-page SEO metadata contains `IS250C` — that is keyword tagging, not a
name. Trim casing diverges (`F SPORT` US vs `F Sport` UK).

**Volvo** — three separate findings:
- Volvo **dropped the P and never wrote "P1800E"**: P1800 → 1800S → 1800E → 1800ES.
- The trailing letter on **PV444K is a PRODUCTION SERIES** (K = 1956), not a trim —
  15 series across the PV444/PV544. Folds to PV444.
- **242/244/245 are cylinder/door digits, not models.** Volvo consolidated the
  badge to plain **"240"** for MY1983, so for over half the production run "240" is
  the *only* correct name. Same for 140 and 260. **`262C` stays separate** (Bertone
  body).

**Jaguar** — `XJ-S` is the launch name (Sept 1975); the hyphen left the *badge* only
at the 1991 facelift (boundary is a VIN: JDHT records 179733 as the last
pre-facelift). Jaguar's own heritage copy still writes XJ-S. 11 spellings plus
HE/V12/Auto and the XJ-SC cabriolet fold to `XJ-S`; **XJR-S stays separate** (a
distinct TWR-developed model). `XJ6`/`XJ12`/`XK8` are closed up, never spaced.
`E-type` is capital-E lowercase-t.

**⚠️ THE CASE TRAP — read this before writing any dedup.** Jaguar `S-type`
(1963–68) and `S-TYPE` (X200, 1999–2008) are **different cars distinguished ONLY by
case.** A case-folding dedup merges them silently. We hold only the X200 today, so
no damage — but the shape recurs with any ALL-CAPS modern revival. Dedupe
case-sensitively, or disambiguate on years.

**Discriminator for case-only pairs, from the two-wheel half:** registry
case-inconsistency is heavily **SKEWED** (one form dominant 10–100×, because one
data-entry convention prevails). Two genuinely different products sharing a
case-folded name show **comparable volumes and disjoint year ranges**. All 515
case-only raw pairs measured in the 2W aggregate were skewed → collapsing correct.

### 2.8 KBA `Modellreihe` is a normalized `Handelsname`, not a model name

Confirmed distortion classes — do **not** use it as a canonical name:

| class | example |
|---|---|
| marque-token stripping | `3` → MG3; `1`/`3`/`5` → smart `#1`/`#3`/`#5` |
| internal type codes | `RX6` is not a model — it is MG HS/EHS (`'MG RX6;-HS;-EHS;-EHS PHEV'`) |
| truncation | `BZ4` → Toyota **bZ4X** (present since the 2023 workbook, so never a real name); `STARRAY` → Starray EM-i |
| temporal merge | `"GLK, GLC"` (§2.4) |
| silent family absorption | `VW TRANSPORTER` absorbs Multivan + Caravelle + California — **nothing in the string signals it**, and FZ 10.1 has no Multivan row at all |
| China-market names | `ES8` → **EL8** for Europe (NIO renamed ES6/ES8 → EL6/EL8) |
| platform-family numbers | `VOLVO 60` = S60+V60, **not** the 60-series platform (XC60 is its own row) |

**`VOLVO 60` deserves its own warning.** It is currently published as
`car/volvo/60` with `body_types: ["hatchback"]`, single-sourced, and it sits in
dropdowns beside genuine bare-number nameplates (`audi/90`, `rover/60`,
`land-rover/90`) **with nothing to tell a consumer which is real.** It also breaks
an id-stability invariant: `volvo/60` meant {S60,V60,V60CC} in 2021 and ~{V60} in
2026. Recommended resolution is to map it to `volvo/v60` (the surviving nameplate),
which needs the FZ 11.1 segment join to justify. **Still open.**

`CX-6` is the opposite case and a good example of why you measure: it looks like a
CX-60 truncation but is a **real new model, the Mazda CX-6e** — KBA lists CX-6,
CX-60 and CX-80 as three separate rows, and CX-6 first appears in the 2026
workbooks.

### 2.9 Inline flow-style YAML silently ate 21 entries

```yaml
Honda:      { Honda: null }        # motorcycle fi|nl AND moped nl|nz
```

Valid YAML, parses identically to a block mapping — and **invisible to every
line-based tool either session writes** (alphabetizer, provenance checker,
collision merger, union merger). A union merge could not see 21 such entries, so
they vanished from a branch with **no conflict and no lint failure**.

Guarded: `lint_curation.rb` now rejects flow mappings in the curation files. One
entry per line, each with its own `#` reason.

### 2.10 A `null` rename silently killing a move — three times

A `null` rename beats a move **by design** (an explicit drop is a stronger
statement than a routing rule). So `SEAT: {Formentor: null}` alongside
`"SEAT|Formentor": "Cupra|Formentor"` means the record **vanishes instead of
moving**, and nothing says so.

It re-broke three times, because **a union merge cannot tell a deliberate DELETION
from an entry the other side simply lacks** — so every merge from `main`
resurrected the retired null.

Guarded: `lint_curation.rb` fails when a move key has a blocking null, naming both
remedies.

### 2.11 A bare integer as a rename target crashing the build

```yaml
"244": 240      # ← YAML parses the VALUE as Integer
```

The pipeline slugifies rename targets, `slugify` calls `.downcase`, and the build
dies with `undefined method 'downcase' for an instance of Integer`. Nine entries
from a Volvo body-variant fold. **Local unit tests could not catch it** — they
exercise `classify()` with string literals, so a YAML *type* error in the data
never reaches them. Only a real build over real rows hits it.

Guarded: `lint_curation.rb` requires rename/move values to be String or null, with
the crash named in the error message.

### 2.12 A cross-repo contract mismatch where no move ever fired

`normalizer.rb` looks up `moves["#{make}|#{nameplate}"]` — **string** key — and
splits the value itself. The loader in `overrides.rb` normalized both sides into
**arrays**. Every lookup missed, so **no move ever fired**: no error, no warning,
just SEAT-badged Cupras staying under SEAT.

Caught by two spotchecks (`cupra/formentor` and `genesis/gv60` missing `de`) — not
by any test. **That is the argument for writing tripwires before trusting a new
mechanism.** The contract is now stated in the loader so neither side can change
one half alone.

### 2.13 A test that disproved correct data (and the reasoning failure around it)

`test_override_key_reachability.rb` ran `classify()` with **moves ACTIVE** while
measuring "what does the pipeline produce for this key". The move under test
therefore applied, and the helper reported its own **target**:

```
"Piaggio|Vespa Sprint 125"  → helper says "Sprint 125"   (the POST-move model)
                            → key says   "Vespa Sprint 125"   → "unreachable"
```

All 48 correct move keys looked dead. On that basis one session wrote a confident
wrong diagnosis ("the prefix strip eats the badge") and handed back 46 correct
moves — **after its own direct run had already printed `PIAGGIO VESPA SPRINT →
Vespa/Sprint`.**

The other session's rebuttal needed no experiment: *a rename key matches the
post-strip nameplate; `SEAT: Cupra Leon: null` demonstrably works; therefore the
post-strip nameplate IS `"Cupra Leon"`; therefore a move keyed on it cannot fail.*

Fixed with a `without_moves` seam mirroring the existing `without_renames`.
**Lesson: the test is the derived artifact; the pipeline is the fact.** When a test
and a direct observation disagree, suspect the harness.

### 2.14 Accent-stripping merged two different cars

A duplicate detector normalized names with `gsub(/[^a-z0-9]/,"")`, which **deletes**
non-ASCII letters instead of folding them. So `ë-C3` — Citroën's **electric** C3, a
distinct model — normalized to `c3` and was merged into the petrol C3. Same for
`ë-C4` and `ë-Jumper`. 11 false merges.

**Always fold with NFKD (`unicode_normalize(:nfkd)` then strip marks), never
delete.**

### 2.15 A "correct" rule over-applied across marques

Fixing Audi `A3` (registers split it to `"A 3"`) with a rule "fuse a short letter
prefix to digits" also fused marques that genuinely space their names, turning the
researched `DS 3`/`DS N°4` into `DS3` and Gas Gas's `EC 300` into `EC300`.

**A rule that is right for one marque is not right for all of them.** Whole-file
scripts need an ownership filter *and* a per-marque exception list.

### 2.16 Whole-file scripts crossed the ownership boundary

The same scripts rewrote 64 entries under the other session's makes, some wrong for
their conventions (Kawasaki writes `ZX-10R` **with** the hyphen; the rule produced
`ZX10R`). All 64 were reverted to exactly what `main` had. **A correct rule plus an
unfiltered file walk is still a boundary violation.**

### 2.17 A silent push failure

`git push … 2>&1 | tail -1` swallowed a non-fast-forward rejection. The fix was
believed landed for ~20 minutes while `origin` did not have it. **Verify the remote
state (`git show origin/main:path | grep`), never the "pushed" echo.** Note also
that piping a command into `tail` makes the pipeline's exit status `tail`'s — which
is how a `&&` chain continued past two lint failures.

### 2.18 Licence pins: the 13-day outage, and one pin that guards nothing

The monthly build was red for 13 days. Root cause: three `whole_file` licence pins
tripping on cosmetic page churn (`ie_cso`, `my_jpj`, `ca_nrcan`). All three terms
were diffed first and were **substantively unchanged** — zero SA/NC/ND clauses, so
the CC-BY composite was never at risk.

**`my_jpj` is the one to remember:** its page prints `Data as of <month>`, so a
`whole_file` pin there **breaks every single month by construction**. Migrated to
`phrases` mode along with `uk_dft` and `nz_nzta`. Worth knowing: OGL v3's *"You are
free to: copy, publish, distribute and transmit the Information"* does **not**
match as a substring on the National Archives page (split across markup) — so
phrases must be *tested*, not assumed.

**⚠️ `be_fps` is pinned but NOT ingested, and its page contains no licence text at
all** — "licence", "Creative Commons" and "CC0" are all absent from 22,607 chars
(it is a JS shell). **That pin guards nothing, and a green gate there is not
evidence Belgian terms were ever checked.** It must be re-pointed before Belgium is
ingested. **Still open.**

Also: `actions/cache` never saves on a failed job and `save-always` was unset, so
the cache was restored-but-never-refreshed through every failure, drifting toward
7-day eviction. Fixed.

### 2.19 A stale-month payload, and why zip magic is not enough

KBA sometimes serves a **different month** than the URL asks for — a request for
`fz10_2026_05.xlsx` returned the **March** workbook (174 KB) on one fetch and the
correct May one (178 KB) on the next. Zip magic cannot catch that, and a silently
stale month means silently stale popularity for a whole country. The source now
asserts the sheet's own `"FZ 10.1 … im <Monat> <Jahr>"` title matches the request.

(Pre-existing related quirk, already handled: KBA answers missing files with HTTP
200 + an HTML 404 page, so `curl --fail` cannot see it.)

---

## PART 3 — THE TOOLING (what exists, what it catches, how to run it)

All stdlib-only, all seconds-fast, all wired into CI on every PR.

| script | catches |
|---|---|
| `scripts/lint_overrides.rb` | shape/vocabulary of the override files (pre-existing) |
| `scripts/lint_curation.rb` | **duplicate YAML keys** (safe_load keeps the LAST and discards the other block), flow-style mappings, non-String rename values, null-vs-move contradictions, make keys that match no display name, `kind_maps` shape, and added-line provenance |
| `scripts/lint_dataset.rb` | name-SHAPE defects in the published catalog + **drop-list effectiveness**; `legit`/`debt` buckets in `data/name_shapes.yml` |
| `scripts/reorg_make_blocks.rb` | keeps make blocks alphabetical; **refuses to write unless the parsed YAML is identical before and after** |
| `scripts/gen_ownership.rb` | regenerates `OWNERSHIP.yml` from the catalog (ties pinned in-script) |
| `scripts/find_duplicate_spellings.rb` | the same car published under two spellings |
| `pipeline/tools/verify_source_landing.rb` | does a source's parsed rows land on the RIGHT records? existing / new / dropped |
| `pipeline/tests/test_override_key_reachability.rb` | rename/move/styling keys that can never match, and drop entries that fold together |

### 3.1 The two mechanisms inside `lint_dataset.rb` worth understanding

**Corroboration as a shape test.** The dataset publishes when ≥2 independent
sources agree (`DECISIONS.md`). The same bar separates a genuinely odd nameplate
from a parser artifact: **of 1,061 bare-integer names, 890 are corroborated**
(Peugeot 208, Abarth 595, Mercedes 1824 trucks) **and the 171 thin ones were almost
exactly the KBA leak.** One evidence condition replaces hundreds of allowlist lines.

**`legit` vs `debt`, and why one bucket would rot.** `legit` = the shape is correct
and permanent (`MAZDA3` really is the nameplate); entries must be shape-general and
evidenced. `debt` = the shape is wrong and unfixed, with a `count` that may only go
DOWN. **Spotless ≡ `debt` is empty.** With a single bucket you reach zero by
relabelling junk as legitimate, which is how allowlists become landfills.

**The drop-coverage rule** replaces ~69 hand-written absence spotchecks: *no
published make may match any drop entry, under either transliteration convention*
(`ü→u` and `ü→ue`). It immediately confirmed three silent failures — `PÖSSL`,
`BÜRSTNER`, `NIESMANN+BISCHOFF` were listed as dropped yet 80 records stayed
published, because NL/FI registers transliterate umlauts and exact matching could
never work.

### 3.2 The pending-vs-failure distinction (§5.1's practical form)

Because `catalog/` lags `overrides/`, a drop merged today is legitimately still
published. `lint_dataset.rb` therefore only FAILS for entries **older than the
published catalog**, using **the commit that last touched `dist/`** as the cutoff:

```bash
git log -1 --format=%ct -- dist/vehicles.csv
```

**Not "the last green run"** — a validate-only run publishes nothing, so those two
diverge on almost every run once PRs are landing.

---

## PART 4 — MEASURED RESULTS

```
                                     before        after
records (four-wheel half)            10,561        9,823 distinct
  non-vehicles removed                              107  (14 motorhome builders)
  junk strings dropped                               66
  merged onto a canonical                         1,020
duplicate-spelling groups               632          ~140
Lexus nameplates                        180           25
Mercedes-Benz car nameplates            481          378
Germany present in van/truck/bus      0 recs   79,534 registrations routed
rename entries                          146        1,203  across 148 makes
moves                                     0           58
spotcheck rows                           30          120
pipeline tests                    0 for xlsx     65 runs / 274 assertions
```

Two-wheel half (other session): 821 fabricated names replaced with forms that
exist upstream, 77 escaped motorhome floorplan codes removed, 1,052 `former_ids`
migration aliases.

---

## PART 5 — THINGS THAT WILL MISLEAD YOU

### 5.1 `catalog/` lags `overrides/` — the single most common confusion

A curation PR does not change the published catalog until a build runs. So:

- `find_duplicate_spellings.rb` counts will **not drop** after your merge.
- A drop you just merged still shows its make as published.
- A rename block keyed on a **pending** casing pin looks inert against today's
  catalog but is merely waiting (`lint_curation.rb` knows this: valid key set =
  catalog display names **∪** `makes/aliases.yml` values).

### 5.2 A shape test is a hypothesis, not a finding

A negative result worth preserving: a "UK annual model-year code" class was filed
from a shape sweep (`AN 400 AL4`, `AM5`…). Investigated: the suffixes exist in the
raw VEH0120 strings, but every published record matching the shape is a **real
model** — Aprilia RS4, Piaggio ET2, Rieju RS2/RS3, Vmoto TC1, e-max LB1. **Six
records, zero defects.**

### 5.3 Corroboration does not clear every class

It separates real numeric nameplates from parser artifacts (§3.1). It proves
**nothing** for `make_as_model` and `embedded_make`: all 51 make-as-model rows were
multi-source, because registers **share** the same fallback habit when a model field
is blank. Those need per-record judgement against the manufacturer's own naming.

### 5.4 Local tests cannot catch data-type errors, and unit tests cannot catch landing

- A YAML type error in the data crashes only a real build (§2.11).
- A parser fix can look perfect at row level and still put everything in the wrong
  place — use `verify_source_landing.rb` (§2.3 was found that way).

### 5.5 A make with only one model in a kind

Dropping a make-as-model row that is a make's **only** model erases the make from
that kind. Where the true identity is unrecoverable, **track it as debt rather than
guess**. Live example: `van/uaz/uaz` — UAZ's own convention is numeric type
designations (452 / 469 / 3741 / 2206), so the row is one of those and an aggregate
cannot say which. Guessing at an unresolvable row is what produced the SEAT
incident.

### 5.6 A test fixture can be invalidated by legitimate curation

Twice, a test asserting a *rule* broke because it used live curation as its
example: `MAZDA/3` as "an uncurated single digit" (then curated), and
`SEAT: Formentor: null` as "a null that beats a move" (then legitimately retired so
the move could fire). **A test about a rule should stub, not depend on data the
rule permits to change.**

---

## PART 6 — OPEN ITEMS, WITH ENOUGH CONTEXT TO ACT

1. **~140 duplicate-spelling groups** (long tail). Run
   `ruby scripts/find_duplicate_spellings.rb --make=<make>`. Safe class = canonical
   already exists; new-canonical class needs the marque's archive (§2.7).
2. **`car/volvo/60`** — a statistical bucket published as a nameplate (§2.8).
   Blocked on the FZ 11.1 join for justification. `volvo/90` is *not* published,
   purely because 988 units missed the count threshold while `60`'s 4,211 cleared
   it — that inconsistency is itself the argument for making the rule explicit.
3. **Add FZ 11.1 as a source.** Same publisher, same DL-DE/BY-2.0 licence, no new
   pin category. Unlocks: the `WOHNMOBILE` filter, the `UTILITIES` car/van
   boundary, and the Volvo/aggregate question — converting a hand-enumerated
   `by_model` map into a principled join. **Highest-leverage remaining item.**
4. **Re-point `be_fps`** (§2.18). Must happen before Belgium is ingested.
5. **`validate.rb` gate for `former_ids`** — assert it never names an id that is
   live in the same kind, and that every rename target that changed an id has an
   alias. The generator now reads intent from the override layer rather than
   inferring from an id diff; the gate is the enforcement half.
6. **`mercedes-benz/atego` and `mercedes-benz/unimog` are in the `car` kind**, via
   AR/FI/NL/NZ rather than KBA — the same motorhome-base phenomenon, proving the
   count threshold is not a sufficient guard.
7. **`former_ids` coverage gap.** It models *renames* (old id gone, new id
   appeared) and cannot express a *merge* into an id that already existed — which
   is most of this pass (`volvo/244 → volvo/240`, `lexus/rx-450h → lexus/rx`). A
   consumer holding `volvo/244` still 404s. Deriving from `renames.yml` +
   `moves.yml` fixes it; also allow an explicit `accepted_loss:` for the three
   documented evidence losses that the country-superset rule would otherwise
   reject forever.
8. **Two unreachable move keys, filed not guessed:** `"SEAT|Cupra Leon"` and
   `"SEAT|Cupra Ateca"` were *retired as nulls and re-added as moves* and now work.
   But the **pooled** `SEAT/LEON` and `SEAT/ATECA` rows still cannot be split — a
   post-strip Cupra-badged row is indistinguishable from the SEAT version.
   Recovering those needs the prefix strip to preserve what it removed.
9. **`kind_maps` `by_model` for `de_kba_fz10` is enumerated, not derived** — 30
   nameplates hand-listed. Item 3 replaces it.

---

## PART 7 — HOW TO VERIFY ANY OF THIS YOURSELF

```bash
# lints (seconds, no network)
ruby scripts/lint_overrides.rb
ruby scripts/lint_curation.rb
ruby scripts/reorg_make_blocks.rb --check
ruby scripts/lint_dataset.rb --report          # add --owner=s4w to scope

# pipeline tests (needs the data repo path from a worktree)
VDB_DATA_REPO=~/GitHub/vehiclesdb rake test

# does a source land on the right records? (offline, cached snapshot)
VDB_DATA_REPO=~/GitHub/vehiclesdb VDB_CACHE_DIR=~/GitHub/vehiclesdb-pipeline/cache \
  ruby pipeline/tools/verify_source_landing.rb

# prove the xlsx bug for yourself
unzip -p cache/de_fz10_202605.xlsx xl/worksheets/sheet4.xml | less   # look at B23/C23

# what does the pipeline actually produce for a raw row?
VDB_DATA_REPO=~/GitHub/vehiclesdb ruby -e '
  require_relative "pipeline/lib/normalizer"
  p VDB::Normalizer.new.classify("SEAT", "FORMENTOR", kind: :car)'   # => ["Cupra","Formentor"]

# measure a register directly (RDW is CC0)
curl -s --get "https://opendata.rdw.nl/resource/m9d7-ebf2.json" \
  --data-urlencode '$select=voertuigsoort,inrichting,count(*) AS n' \
  --data-urlencode "\$where=merk='KABE'" \
  --data-urlencode '$group=voertuigsoort,inrichting'
```

**Release safety:** merging never publishes. `monthly-build.yml` gates
commit/tag/release, the HF mirror and the jsDelivr purge on
`(schedule == day-12 cron) || (dispatch && inputs.publish)`. A validate-only
dispatch is always safe. **No scheduled publish had ever succeeded as of this
work** — the autopilot's publish path is unproven, not merely once-broken.

---

## PART 8 — THE PROCESS FINDING, WHICH OUTLIVES THE DATA FINDINGS

Three times in one day, one session produced a **confident, well-reasoned, wrong**
conclusion, and each time the other session's check caught it:

1. A merged PR deleted SEAT's German lineup as "junk type-codes" — caught by
   reading the parser instead of the data.
2. A second pass proposed deleting the other 148 for the same reason — caught by
   the same read.
3. A session declared 48 moves unfixable — caught by a proof needing no
   experiment, *after its own output had already contradicted it*.

**None of the three was found by its author.** Both sessions also shipped a tool to
fix a silent-loss class, and **each tool had a different silent-loss bug of its
own** (one regex skipped commented lines; the other stripped comments from lines it
moved). Each was caught by the other side.

If you are a single agent maintaining this: **you do not have that check.**
Compensate by (a) preferring the parser/raw source over the published data whenever
they disagree, (b) treating any test that contradicts a direct observation as
suspect until proven, and (c) writing the tripwire *before* trusting the mechanism —
the two spotchecks that caught the moves contract mismatch were written before the
mechanism was believed to work, which is the only reason the failure was visible.

PR #5, opened 2026-07-25
Promote the duplicate-spelling detector to tracked tooling

rameerez:
This found the largest quality defect in the four-wheel half (632 groups / 1,283 records where the same car publishes twice). It was a throwaway script in /tmp.

Header documents the accent-folding gotcha that cost real damage: stripping non-ASCII made `ë-C3` (Citroën's **electric** C3, a distinct model) normalize to `c3` and merge into the petrol C3.


rameerez:
## Context for this PR (and where the full brief lives)

This PR did three things, each with a nuance worth carrying forward:

**1. Landed all 58 cross-make moves.** 46 Piaggio→Vespa (authored by the two-wheel
session), 12 four-wheel (SEAT→Cupra ×7 incl. Leon/Ateca, Hyundai→Genesis ×4, Auto
Union→Audi). Registers file sub-brands under the **type-approval holder** — KBA's
`Marke` comes from the HSN, so Cupra rides HSN 7593 "SEAT"; RDW files 130
Vespa-badged names (79,569 registrations) under `merk=PIAGGIO`.

**2. Retired two stale nulls.** `SEAT: Cupra Leon: null` / `Cupra Ateca: null` said
*"no cross-make move exists"* — true when written, false once `moves.yml` existed.
The gain is **popularity**, not evidence: `cupra/leon` already carried those
countries, so the nulls were lossless on evidence but discarded the NL/NZ rows'
registration counts. A move carries them.

**3. Promoted `scripts/find_duplicate_spellings.rb` from a `/tmp` script to tracked
tooling.** It found the largest quality class in the four-wheel half — 632 groups /
1,283 records where the same car publishes twice because two registers spell it
differently and the spellings slug apart.

### Two traps documented in that script's header

- **Normalize with NFKD accent FOLDING, never by deleting non-ASCII.** Stripping
  accents made `ë-C3` (Citroën's **electric** C3, a distinct model) normalize to
  `c3` and merge into the petrol C3. 11 false merges before it was caught.
- **It reads the PUBLISHED catalog**, which is a monthly build output — so counts do
  **not** drop after a curation merge. Expect them flat, then a step change at the
  next release.

### The immediate history behind this PR, which matters

The 46 Vespa moves were handed **back** an hour earlier on a confident wrong
diagnosis ("the prefix strip eats the badge"), then landed once the other session
disproved it. The actual fault was a test harness measuring with moves **active**,
so each move reported its own target as "what the pipeline produces" and every
correct key looked dead. See `vehiclesdb-pipeline` PR #6.

**Full brief:** PR #4, first comment.

PR #6, opened 2026-07-25
Kind-boundary DECISION (researched) + land 46 Piaggio→Vespa moves

rameerez:
Two related pieces of marque/kind correctness. Owner asked for the kind-boundary questions to be **researched and decided**, not escalated.

---

## 1. The kind-boundary decision

**Q1 — `kind` is a LEGAL-category axis.** Not a judgement call; the current approach is demonstrably incoherent. `renault/twizy` is published in **three kinds at once** — `car` (ar\|gb\|my\|ua), `motorcycle` (nl), `moped` (es\|fi\|nz). Citroën Ami spans car+moped+van; Aixam spans car(9)+moped(5)+van(1). Registers disagree about the same vehicle because each one's vehicle-type word is *administrative*.

**Q2 — L6e/L7e go to `car` with `body_types: ["quadricycle"]`, not a new kind.** Decided by binding internal precedent, DECISIONS.md line 23: *"Three-wheelers fold into `motorcycle` with `body_types: ["trike"]` — no kind explosion."* L5e is an EU category in its own right and was folded into the kind it physically resembles.

The counter-case was researched, not dismissed: Citroën markets the Ami as *"technically not a car"*, and L6e runs on an **AM moped licence** (14 in FR, 16 in UK) while L7e needs B1. But that licence axis splits L6e from L7e — honouring it would put an enclosed four-wheel Ami in with Vespas and split the Aixam range. Fold by physical form; keep the category in the body type.

**Q3 — additive via `former_ids`, but NOT implementable as a curation change.** I built it and reverted rather than half-ship. Only `nl_rdw` plumbs `eu_category` into a `Row`; `es_dgt`, `fi_traficom`, `lu_snca`, `ua_mvs`, `nz_nzta` also feed `moped` and don't. Routing one source reproduced the incoherence, dropped `silence/s04` and `opel/rocks-e` on the car-kind threshold, and labelled moved rows `hatchback` instead of `quadricycle`. Prerequisite named, dependency-ordered plan in the doc. **Recommend its own release.**

Sources: EUR-Lex 32013R0168 · GOV.UK Annex B L-category table · Stellantis/Citroën Ami press material.

---

## 2. The 46 Vespa moves — and why "unreachable" was wrong

The other session reported all 46 as unreachable, concluding the prefix strip eats the badge. **It strips the MAKE (`PIAGGIO`), not the badge (`VESPA`).** 46/46 verified reachable against `origin/main`, then stubbed and watched fire. Their two SEAT failures were the `null` rename winning — the behaviour `test_null_rename_beats_a_move_by_design` already pins.

| | before | after |
|---|---|---|
| `piaggio` records carrying a Vespa badge | 47 | **1** |
| `vespa` records | 88 | **114** |

Evidence **unions**: `vespa/gts` gains `nz`, `px125e` gains `gb|nz`. The one row left behind is deliberate — the bare brand-as-model row is unresolvable to a single model, like PR #1's `SEAT: Cupra: null`, and it's tripwired so it reads as a decision.

+25 `former_ids` generated by reading `moves.yml` as **intent** — an id-diff generator could not have proposed these because most targets already existed under `vespa`, which is exactly the gap their review identified. 124 spotcheck rows. **ALL GATES GREEN.**


rameerez:
## Heads-up: the 46 Piaggio→Vespa moves in this PR are already on `main`

They were landed in #5 (build green) exactly as authored here, together with 12
four-wheel moves and the two retired SEAT nulls. `overrides/models/moves.yml` on
`main` now has **58 entries**, and `MOVES-S2W-staging.yml` is deleted.

```bash
git show origin/main:overrides/models/moves.yml | grep -c '^"Piaggio'   # → 46
```

So the moves half of this PR will either conflict on `moves.yml` or duplicate keys
into it — and a duplicate key is exactly what `scripts/lint_curation.rb` now fails
on, so CI will tell you rather than YAML silently keeping the last one.

Suggest keeping this PR to the **kind-boundary decision**, which is untouched and
entirely yours.

Apologies for the ordering: they were staged here, handed back on a wrong diagnosis
(a test harness measuring with moves active — see `vehiclesdb-pipeline` #6), then
landed once that was corrected, all inside an hour.

**Full brief:** PR #4, first comment.

rameerez:
## Onboarding pointer

Two things landed here: the **`kind`-boundary decision** and the **46 Piaggio→Vespa moves**. Full context in the S2W brief:

Most load-bearing points from this PR:

- **`kind` is a LEGAL-category axis, decided on evidence:** `renault/twizy` was published in **three kinds at once** — `car` (ar\|gb\|my\|ua), `motorcycle` (nl), `moped` (es\|fi\|nz). Registers disagree about the same vehicle because each one's vehicle-type word is administrative.
- **L6e/L7e belong in `car` + `body_types: ["quadricycle"]`, not a new kind** — binding precedent is DECISIONS.md folding L5e trikes into `motorcycle` + `["trike"]` to avoid kind explosion. The counter-case (Citroën markets the Ami as "technically not a car"; L6e runs on an AM *moped* licence) was researched and loses because the licence axis splits L6e from L7e.
- **The migration was later BUILT AND REVERTED** — see `PROPOSAL-kind-boundary.md` and PR #9. It deletes ~35 real microcars unless DGT's national asterisk codes are first mapped to EU categories. **Do not retry without that step.**
- **Why the Vespa moves exist:** RDW files Vespas under `merk=PIAGGIO` because Piaggio holds the type approval and the register's make column is approval-derived — the same mechanism as KBA's HSN putting Cupras under SEAT.
- **A `null` rename BEATS a move by design.** Retiring a null and adding a move must happen in the same change; this cost both sessions an hour of wrong diagnosis.
- **`strip_make_prefix` strips the MAKE, not the BADGE.** It removes `PIAGGIO`/`SEAT`, never `VESPA`/`CUPRA`.
- `motorcycle/piaggio/vespa` is **deliberately** left under Piaggio — a bare brand-as-model row is unresolvable to one model, exactly like PR #1's `SEAT: Cupra: null`. Tripwired so it reads as a decision.

PR #7, opened 2026-07-25
Reunite 17 Vespa nameplates my own moves had split

rameerez:
A defect in **my own work from an hour ago**, found by reading `build/candidates/` — which I should have done before landing the first 46 moves.

## What happened

Registers disagree on whether the badge appears in the model column at all:

```
fi_traficom  merk=Piaggio  handelsbenaming="VESPA ET4 150"   ← my move matched
nl_rdw       merk=PIAGGIO  handelsbenaming="ET4 150"          ← it did not
```

Moving only the badge-prefixed spelling **split each nameplate into two single-source candidates**, and neither then cleared the ≥2-source publication threshold. Records that previously published silently vanished into the candidate queue — **and the build stayed green throughout**, because nothing asserts that a move must not orphan its own twin.

## Recovered, evidence unioned

| id | now | note |
|---|---|---|
| `vespa/et4-150` | motorcycle[fi\|nl] | had vanished entirely |
| `vespa/pk50xl` | motorcycle[gb\|nl] moped[nl\|nz] | gb 25 regs + nl 27 |
| `vespa/150-super` | motorcycle[gb\|nl\|nz] | gb 50 regs |
| `vespa/lx125` | motorcycle[es\|fi\|nl\|ua] | gained ua |
| `vespa/primavera` | moped[es\|fi\|lu\|nl\|nz] | nz 11 regs |

## Safety

Every plate is a genuine Vespa nameplate with no Piaggio-branded namesake (Primavera, Sprint, PK, ET4, LX, GTS, GTV, PX, 150 Super). Candidates matching a real Piaggio line (Beverly, Liberty, Zip, Fly, Ciao, Runner, Velofax, Typhoon, Ape) were excluded by filter. Bare `"50"` was excluded **by hand** — Piaggio has its own 50cc lines, so 3 Dutch registrations aren't worth a mis-attribution.

No new `former_ids` are needed, and that's correct rather than an omission: the un-prefixed ids were *candidates*, never published, so rule 1 of the generator (only alias an id that actually shipped) excludes them.

+2 regression tripwires (126 rows). **ALL GATES GREEN.**

**Lesson worth generalising:** after any cross-make move, read the candidate queue. A move that relocates one spelling of a nameplate can strand the other below the threshold, and every gate stays green while it happens.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -1146,3 +1146,10 @@
 "motorcycle/piaggio/vespa-px200e":             "motorcycle/vespa/px200e"                      # moves.yml Piaggio|Vespa PX200E -> Vespa|PX200E  [nl,nz]
 "motorcycle/piaggio/vespa-sei-giorni":         "motorcycle/vespa/sei-giorni"                  # moves.yml Piaggio|Vespa Sei Giorni -> Vespa|Sei Giorni  [fi,nl]
 "motorcycle/piaggio/vespa-seigiorni":          "motorcycle/vespa/seigiorni"                   # moves.yml Piaggio|Vespa Seigiorni -> Vespa|Seigiorni  [es,lu,nl]
+
+# ---------------------------------------------------------------------------
+# RELEASE 2026-07-25 (f) — un-prefixed Vespa twins (see moves.yml).
+# Registers disagree on whether the badge appears: fi writes "VESPA ET4 150",
+# nl writes "ET4 150". Moving only the prefixed form split each nameplate into
+# two single-source candidates below the publication threshold.
+# ---------------------------------------------------------------------------
--- a/overrides/models/moves.yml
+++ b/overrides/models/moves.yml
@@ -121,3 +121,43 @@
 "Piaggio|Vespa Sprint 125ABS":     "Vespa|Sprint 125ABS"
 "SEAT|Cupra Leon": "Cupra|Leon"
 "SEAT|Cupra Ateca": "Cupra|Ateca"
+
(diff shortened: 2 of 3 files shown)


rameerez:
## Onboarding pointer — this PR fixes a defect the author caused an hour earlier

Read this one if you are about to author a cross-make move. Full context in the S2W brief:

**A MOVE CAN SILENTLY SPLIT A NAMEPLATE.** Registers disagree on whether a sub-brand badge appears in the model column *at all*:

```
fi_traficom  merk=Piaggio  handelsbenaming="VESPA ET4 150"   ← a badge-prefixed move matches
nl_rdw       merk=PIAGGIO  handelsbenaming="ET4 150"          ← it does not
```

Moving only the prefixed spelling strands the twin under the old make. Both are then **single-source**, neither clears the ≥2-source publication threshold, and records that used to publish **vanish into `build/candidates/` while every gate stays green**. It cost 17 nameplates, including `vespa/et4-150` disappearing entirely.

**Habit to adopt: after any cross-make move, read `build/candidates/`.**

Structurally this can only bite registers that embed the badge in the model column (RDW) — never those that put the marque in the make column (KBA's HSN). So the gate condition is tight: *fail when a `moves.yml` key contains the target make's name AND a candidate exists at `<source_make>/<key minus that prefix>`.* That skips the 12 KBA-sourced moves by construction and would have caught all 17 of these.

Bare `"Piaggio|50"` was excluded **by hand** — Piaggio has its own 50cc lines (Ciao, Bravo), so 3 Dutch registrations are not worth a mis-attribution.

PR #9, opened 2026-07-25
Amend kind proposal: the gating blocker is Spain's national codes (doc only)

rameerez:
**Doc-only.** No data, no code. Records the outcome of actually building the kind migration, so the next attempt starts from the real blocker instead of my wrong estimate.

## I estimated this three times and was wrong three times

| estimate | claim | why it was wrong |
|---|---|---|
| 1st | "plumb `eu_category` into `Row` for 5 sources" | 3 of the 5 already route on the EU category |
| 2nd | "~6 map entries across 3 sources" | ignored what happens to records that **split** |
| 3rd | **a Spanish national-code → EU-category mapping** | evidenced by building it and measuring |

## What building it actually produced

```
body_types ["quadricycle"] derived correctly     76 records — the mechanism WORKS
make/model pairs that VANISHED ENTIRELY          35
```

And the 35 are exactly the vehicles the migration exists to classify correctly: `citroen/my-ami-buggy`, `fiat/topolino-dolcevita`, `chatenet/ch28`, `estrima/biro-van`, `casalini/m12`, `aixam/k2`, `garia/club-car-urban-l7e-s`.

## Root cause — in `es_dgt.rb`'s own comment, which I read past twice

> *"Spanish NATIONAL codes (asterisk series) — the bulk of ES two-wheelers arrives under these, not EU categories"*

Spain routes most two-wheelers by `*02`/`*03`/`*05` codes carrying no L6e signal, so it can't send L6e to `car`. Any L6e with Spanish evidence then splits:

```
nl_rdw  L6  → car    290 vehicles · single source · car threshold 1000  → candidate
es_dgt  *NN → moped  single source               · moped threshold 300  → candidate
```

Neither clears its threshold and the record disappears. `silence/s04` (the S04 Nanocar) was published `es|nl` and ends up in neither kind.

Same shape as the Piaggio→Vespa split I caused earlier today — relocating one register's rows while another's stay behind strands both below the bar. That lesson was already written down when I did this.

## The Q1/Q2 decisions are unchanged

`kind` is a legal axis; L6e/L7e belong in `car` + `body_types: ["quadricycle"]`. That evidence is independent of implementation cost, and the body-type derivation is proven to work. Only the cost estimate was wrong.

The corrected plan is dependency-ordered in the doc, with the `Hash#dig`-on-a-String trap that silently zeroed nl_rdw written down so nobody rediscovers it.

### DIFF
--- a/PROPOSAL-kind-boundary.md
+++ b/PROPOSAL-kind-boundary.md
@@ -211,3 +211,74 @@ trial, then reverted rather than half-shipped:**
 sources, both maintainers' kinds, the body-type vocabulary and ~40k
 registrations of evidence. The current state is *known*-wrong and tripwired; a
 half-migration would be *unknown*-wrong.
+
+---
+
+## Q3, THIRD REVISION (2026-07-25, after building it) — the gating blocker is Spain
+
+I estimated this work three times and was wrong three times. All three are recorded
+so the next attempt does not repeat the sequence:
+
+| estimate | claim | why it was wrong |
+|---|---|---|
+| 1st | "plumb `eu_category` into `Row` for 5 sources" | 3 of the 5 already route on the EU category |
+| 2nd | "~6 map entries across 3 sources" | ignored what happens to records that SPLIT |
+| 3rd | **a Spanish national-code → EU-category mapping** | evidenced by building it and measuring |
+
+### What actually happens if you flip the maps today
+
+Built end to end (source maps + `nl_rdw` kind_map + `eu_cats` on the reconciler
+entity + a `quadricycle` body-type derivation + the validator vocabulary), then
+measured against the pre-change build:
+
+```
+body_types ["quadricycle"] derived correctly     76 records — the mechanism WORKS
+make/model pairs that VANISHED ENTIRELY          35
+    citroen/my-ami-buggy · fiat/topolino-dolcevita · chatenet/ch28 · ch28hdi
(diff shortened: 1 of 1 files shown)


rameerez:
## Entry point for future maintainers

Everything from the 2026-07 correction pass is reachable from `AGENTS.md`'s read order:

| where | what |
|---|---|
| `NAMING.md` | the naming canon + evidence standards |
| `PROPOSAL-kind-boundary.md` | the `kind` decision, its evidence, and the Spanish-national-code blocker |
| `CORRECTION-PASS-2026-07-LOG.md` | the verbatim 40-turn working log, wrong turns included |
| [4W brief](https://github.com/vehiclesdb/vehiclesdb/pull/4#issuecomment-5079545135) | curation, canon, incidents, open items |
| [code-contracts brief](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/1#issuecomment-5079564588) | xlsx reader, KBA source, moves, test seams, licence pins |
| [2W brief](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/2#issuecomment-5079726629) | two-wheelers, build triage, `former_ids`, silent-failure catalogue |

**If you are about to delete records whose names look like registry junk, read the 4W brief first.** That exact reasoning deleted SEAT's entire current German lineup once and nearly deleted 148 more records including the best-selling car in Germany.

PR #8, opened 2026-07-25
AGENTS.md: point the read order at the correction-pass briefs

rameerez:
The long-form context from the 2026-07-25 pass lives in two PR comments. A future agent will never find it unless the entry point says so — and the one warning that must be seen *before* a decision ("a bare integer is a parser suspect before it is a data suspect") is exactly the kind that gets skipped when it is buried.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13, opened 2026-07-25
Close G8's 2W half, fix G12, add a cross-make duplicate detector

rameerez:
Discharges the two open **S2W** items in PRD-QUALITY's G-register, and adds one detector for a gap found while doing it. **No curation verdicts ship here** — this is tooling plus one hygiene fix. Every number below is measured against a fresh local build, not the committed `catalog/` (which is the last release and predates the correction pass).

## G8 — the 2W duplicate-spelling sweep had never been run

`find_duplicate_spellings.rb` was hardcoded to `~/GitHub/.vdb-worktrees/s4w-data` and `own["s4w"]`. Not a bug in their work — it was written for one side — but it means the two-wheel half was never swept at all.

Parameterized via `VDB_SIDE` / `VDB_DATA_REPO` / `VDB_CATALOG` / `VDB_OUT`. **Defaults reproduce the original s4w behaviour exactly.** Output filenames now carry the side; both halves previously wrote to the same `/tmp` paths and would overwrite each other with no warning.

| half | groups | records |
|---|---|---|
| 2W (new) | 165 | 335 |
| 4W (re-run on the fresh build) | 146 | 293 |
| **repo-wide** | **311** | **628** |

The 4W re-run landing on 146 matches G8's "~140", which is the check that the parameterization is faithful rather than accidentally selecting a different set.

### Why no renames ship in this PR

**The `canonical()` column is not a decision procedure for what remains.** Its tiebreak is `variants.max_by { tokens.size }` — the most-tokenized spelling. For alphanumeric type designations that systematically selects the registry-mangled form.

Measured on the residue:

| | 2W | 4W |
|---|---|---|
| canonical spaces out single letters | 8 (4.8%) | 8 (5.5%) |
| canonical looser than the tightest variant | 150 (90.9%) | 126 (86.3%) |
| canonical is tightest or ties | 7 (4.2%) | 12 (8.2%) |

The single-letter cases are unambiguous defects in the proposal, on both sides:

```
2W  "F X R S" ← Fxrs        "W L C" ← Wlc         "R Nine T" ← R Ninet
    "Panigale V 4/-S" ← Panigale V4S              "Z 50 A K 2" ← Z50AK2
4W  "R X 7" ← RX7           "B G T" ← B G T       "M 151 A 1" ← M151A1
    "Ghibli S Q 4" ← Ghibli SQ4                   "V 70 2.4 T" ← V70 2.4T
```

But **"tightest wins" is not the rule either** — Fat Boy, Low Rider, Electra Glide, Wide Glide, Road King, Gold Star, Thunder Bolt, Desert X and Super Sport are genuinely spaced in the marque's own usage, while Sportster, Panhead and Shovelhead are genuinely closed. Both directions are wrong as a rule. What is left is per-nameplate marque-archive research, which is exactly §7's checklist and exactly why this is batch work.

One thing worth recording because it looked like a pipeline bug and is not: the space-mangled forms are **real registry text**. `SPORT STER`, `PAN HEAD`, `F X R S`, `SHOVEL HEAD` and `W L C` all appear in the `nl_rdw` and `nz_nzta` snapshots themselves. Two independent registries carrying the same mangling is why they cleared the 2-source publication bar. `two_wheeler_spacing` did not invent them.

## G12 — six empty make blocks

Removed `Bsa`, `I-Coco`, `Iva`, `Matchless`, `NSU`, `Yamaha` from `renames.yml`.

Verified before deleting, since an empty block could in principle be load-bearing: they hold `nil`, and the bulk PR that created them (`11092cf`) added **zero** rename entries to any S2W-owned make. The ownership boundary held across all 1,012 entries in that PR — reported as a negative result because an independent check that comes back clean is still a result.

Added a lint (negative-controlled — it fails on a reintroduced empty block and passes when removed). Two reasons it is worth a gate rather than a one-time cleanup: `nil` crashes naive iteration (`.values.map(&:size)`), and an empty block reads to a human as "this make was curated" when nothing was.

## New: `scripts/find_duplicate_makes.rb`

The existing detector groups records **within** a make. It is therefore blind by construction to *one marque published under two make ids* — those records never share a group key. `truck/m-a-n` sat next to `truck/man` through the entire correction pass and no gate, lint or detector mentioned it. It was found by accident while auditing cross-kind drop leakage.

**Pass 1 — fold collisions** (same fold as the pipeline's `make_dropped?`, deliberately: a divergence means the detector reports collisions the pipeline cannot act on):

```
man    S4W  man    "MAN"    42 models [bus+truck+van]
       ?    m-a-n  "M.a.n." 2 models  [truck]          << cross-owner
emax   S2W  e-max  "E-Max"  4 models  [moped+motorcycle]
       S2W  emax   "Emax"   2 models  [moped]          << mine, queued
```

**Pass 2 — legal-entity and conflated ids**: 32 makes / 71 records (`Fantic Motor`, `Grecav Spa`, `Winora-Staiger Gmbh`, `Kia Motors`, `Quadro Vehicles`, `Opel-Vauxhall`, `Chevrolet GMC`, …).

The script leads with its own counterexamples, because the obvious rule is wrong in **both** directions and a blanket application would do real damage:

- **`club-car`** — Club Car *is* the brand. Stripping `-car` is wrong.
- **`renault-trucks`**, `iveco-bus`, `heuliez-bus`, `volta-trucks` — distinct legal manufacturers with their own type approvals. For the truck/bus kinds these may be the *correct* make, not a suffix to strip.
- **`leyland-daf`**, `austin-morris`, `steyr-puch`, `vdl-bova`, `tadano-faun` — all real historical marques. A joined name is not automatically a conflation.
- `chevrolet-gmc` and `opel-vauxhall` most likely *are* registry conflations.

Which is which is approval-holder research (§7), not string work. The script proposes; it writes nothing.

### The trap, documented at the top of the file

`bus/factory-built` has models **"Geely"**, **"Yutong"**, **"Zhongtong"** — three genuine bus manufacturers parsed into the *model* column by an NZTA origin flag. The obvious fix (add `FACTORY BUILT` to `drop.yml`) **deletes three real manufacturers**. The fix is a move. This is the "confident deletion of real data" shape from §8.4, live in the current build.

## Two findings handed to S4W rather than acted on

Both are in S4W-owned makes, so they are theirs under one-make-one-owner:

1. **All 85 `drop.yml` entries are kind-scoped, effectively all under `car`.** Mostly correct by design — Scania/Iveco/DAF/MAN dropped from `car` (motorhome-base leakage) while their genuine trucks stay, 411 records behaving exactly as intended. But **11 records across 6 makes leak into kinds their drop does not cover**: `niesmann-bischoff` (truck, 2), `hymer` (truck 2 + van 1), `mobilvetta` (truck 1), `auto-trail` (van 1), `chausson` (van 1) — motorhome coachbuilders, same class as G7's Atego/Unimog. Plus `m-a-n` (truck, 2), which is a merge, not a drop.
2. **17 makes in the fresh build have no owner in `OWNERSHIP.yml`** (19 records) — `m-a-n`, `karsan`, `kia-motors`, `opel-vauxhall`, `quadro-vehicles`, `westfalia-mobil-gmbh`, … The file is a static snapshot of the split; every build can mint makes belonging to nobody. Over 45–60 batches that compounds silently. Suggest `gen_ownership.rb` gains an `--assign-new` pass, or the batch flow assigns orphans at cut time.

Also confirmed while here: the `drop_fold` transliteration fix works — `poessl` and `buerstner` (68 records) are gone from the fresh build. `lint_dataset` still reports them because it reads the committed `catalog/`, which is G13's stale-baseline problem, not a regression. `lint_dataset`'s three failures are byte-identical before and after this PR.

## Gates

`lint_overrides` OK · `lint_curation` OK (14 files) · `reorg_make_blocks --check` OK · `lint_dataset` unchanged from baseline (G13). No pipeline files touched. No records change until the next build; this PR is tooling plus one key deletion.


rameerez:
## Second commit on this branch — `905f10f`, the B1 curation work + B3 resolved

The detector added in the first commit immediately found real defects, so rather than open a second PR against the same override files (which is how the union-merge incidents started), B1 stacks here. **This commit does change records.** Verified with a full build and a per-id diff.

### Two make merges

`find_duplicate_makes.rb` pass 1 found the first two cases of one marque under two make ids.

**`emax` → `e-max`.** RDW carries both `merk` spellings for one marque **in the same file, same month**: `EMAX` (models `EMAX-110S`, `EMAX-90S`, `LB1`, `N/A`) and `E-MAX` (models `110S`, `80L`, `90S`, `ALL-4`). The brand writes itself hyphenated — the manufacturer's owner's manual is `emax_90s_110s_brugermanual.pdf`, and Vmoto's own copy reads "the Vmoto **e-max** 110S" ([CarsGuide](https://www.carsguide.com.au/car-reviews/vmoto-e-max-110s-scooter-review-11234)). The un-hyphenated rows are also the sloppier filing (make embedded in the model), so the merge runs toward the tidier spelling. This is the **opposite direction** from the one proposed in NEGOTIATION.md Turn 43; flagged there with the evidence rather than done silently.

**`zero` → `zero-motorcycles`.** UK/NZ/UA register the marque as bare `ZERO` (models DS, DSR, FX, S, SR, SR/F, SR/S — unambiguously Zero's lineup); NL/ES/FI/LU use `ZERO MOTORCYCLES` (82 rows). Merged toward the **longer** form, which inverts the corporate-suffix instinct: the manufacturer's own model pages carry no "Zero \<model\>" product string — body copy is the bare model code and the brand appears only as "Zero Motorcycles" / "Zero Motorcycles, Inc." ([zeromotorcycles.com](https://zeromotorcycles.com/model/zero-dsrx), 2026 footer). NAMING.md's *"the make is the marque, not the registrant"* strips the **Inc.**, not the **Motorcycles**. Bare `Zero` would also be a uniquely collision-prone make string.

Both aliases are **exact-key** on purpose: RDW also has `merk="ZEROBIKES"` (20 rows) and `merk="ELI ZERO"` (2 rows), different marques, must not match.

### One drop, which needed no decision

`motorcycle/eigenbouw/softail` — Dutch for "self-built", RDW's placeholder for home-built machines. **NAMING.md line 33 already lists `EIGENBOUW` under "not a marque at all → drop for that kind"**; it had simply never been implemented. Deliberately *not* moved to Harley-Davidson: a self-built frame in a Softail style is not a Harley, and inventing that provenance is worse than dropping one row.

### A test gap, found by nearly shipping the bug it exists to catch

Merging a make **silently orphans every rename block filed under its old display name.** `model_renames[make]` is keyed on the resolved display name, so the pre-existing `Emax:` block went inert the moment the alias landed — and `e-max/emax-90s` would have been reborn beside `e-max/90s`. Both lints stayed green.

`test_override_key_reachability.rb` — written for exactly this trap — **also stayed green.** `reachable_forms` compares only `out[1]`, the nameplate; for an orphaned block the key still matches itself and scores reachable. `out[0]`, the make, was never checked. Caught only by calling `classify` on the raw rows before and after the alias.

Fixed in [pipeline#8](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/8) with a negative-controlled `test_rename_make_blocks_are_reachable`. 67→68 tests, 251→253 assertions.

### `former_ids` — 11 aliases, and rule 3 amended rather than violated

Rule 3 said *"same kind, same make; a cross-make correction is a `moves.yml` concern."* A make **merge** is precisely a cross-make id change that `moves.yml` **cannot** express: moves relocate a nameplate between two *live* makes, and here the source make ceases to exist.

The exception added is narrow and conditioned on that fact: **permitted only when the old make no longer appears in the build at all** — which is what makes the alias unambiguous. With `zero` gone, `motorcycle/zero/sr-f` can mean exactly one thing. Verified against the build: `zero`/`emax`/`eigenbouw` are gone as makes, no aliased old id is still live (rule 1), every successor exists.

### Verification — per id, not in aggregate

Full build, then a diff. **17,038 → 17,088 records.**

12 ids disappeared in the five makes touched. All 12 have a successor and **none lost a country — every merge strictly gained evidence**:

```
motorcycle/zero/ds     -> zero-motorcycles/ds     gb,nz      -> +nl
motorcycle/zero/fx     -> zero-motorcycles/fx     gb,nz      -> +es,nl   (into a pre-existing id)
motorcycle/zero/s      -> zero-motorcycles/s      gb,nz,ua   -> +nl
motorcycle/zero/sr     -> zero-motorcycles/sr     fi,gb,nz,ua-> +nl
.../motorcycle-xe      -> zero-motorcycles/xe     es,nl      -> +fi,gb
moped/emax/110s        -> moped/e-max/110s        fi,nl      unchanged
motorcycle/eigenbouw/softail                      dropped on purpose
```

The other 107 disappearances in the same diff are in **116 makes never touched here** — the sources re-fetched between the two builds. Make-scoped overrides provably cannot reach them (aliases are exact-key on raw strings, drops are exact-match, rename blocks are make-keyed).

> **Worth flagging for G2's design (@S4W, A2):** roughly one hour of registry churn produced **119 disappearances and 169 appearances by itself.** A no-vanish gate must compare builds from **identical source snapshots**, or it is pure noise and will be muted within a week.

**4 new spotchecks (126 → 130), all 130 passing** against the build. They guard the merge specifically, which fails differently from a model merge: nothing vanishes, the make just silently forks back into two and consumers see the marque twice with half its countries each.

### B3 resolved — the Spanish blocker was my own misreading

Full write-up in `PROPOSAL-kind-boundary.md` (estimate 4, measured end to end). Short version:

- The codes carrying quadricycles are **`*19/*20/*21/*26/*27`**, not `*02`–`*17`, and **`es_dgt.rb` already maps all five.** There was no missing mapping to research.
- `silence/s04` vanished because `L6E` *and* `*21` both map to `:moped` and my migration flipped only `L6E`. Its 153 Spanish rows are all under `*21`. A two-line omission — the Vespa lesson in a different register.

Verified two independent ways **in-file**, because the obvious citation is a trap: RD 2822/1998 Anexo II §B says `21` = *"Capitoné"* (a padded furniture van), `27` = *"Cisterna"*, `02` = *"Bicicleta"* — none of which matches the rows, and per [DGT's own interface document](https://sedeapl.dgt.gob.es/IEST_INTER/pdfs/disenoRegistro/vehiculos/matriculaciones/MATRICULACIONES_MATRABA.pdf) that vocabulary lives in a **different field** (#53). Any published gloss of "`*21`" should be treated as unverified.

1. **Cross-join on RD 2822 code (field #53, which is in the file):** `L6E` rows carry `0300`/`0311`/`0320` — the *identical* set as `*19/*20/*21`. `L7E` carries `0600`/`0611` — identical to `*26/*27`.
2. **Mass sits exactly on the regulatory limits:** `*21` p90 = **425 kg**, `*27` p90 = **450 kg** — the Reg. (EU) 168/2013 Annex I L6e and L7e unladen limits.

Split risk measured **per nameplate**: **75 move cleanly, 1 splits** (`BOMBARDIER CAN-AM`, `*17`+`L7E`, 2 rows, a trike — leave `*17` alone per DECISIONS.md line 23). Note `LIGIER JS50` spans `*21`+`*27`+`L6E` and `MICRO MICROLINO` spans `*27`+`L6E`: they are safe **only because all seven codes flip together.** The set is atomic.

Also found and **not** fixed here (pipeline file, S4W is in there for A2): `es_dgt` reads `EUCAT = [426, 3]` but DGT declares field #48 as **CHAR(4)**. Measured across 586,765 rows: 19 truncated, 4 actually lost. Negligible today, but Reg. 168/2013 subcategories (`L6e-B`, `L3e-A1`) are longer, so this becomes real loss the moment Spain emits them.

### Gates

`lint_overrides` OK · `lint_curation` OK (14 files) · `reorg_make_blocks --check` OK · pipeline `68 runs / 253 assertions / 0 failures` · full build exit 0 · **130/130 spotchecks pass** · per-id no-vanish verified with zero evidence loss.

rameerez:
Rebased onto current main (`8da0b78`), which un-reds the CI — the earlier failure was #16's `AUTO-TRAIL`/`AUTOTRAIL` duplicate on main, not anything in this branch. #18 fixed that independently.

Conflicts were all **append-only overlap** with #16: both PRs added to `makes/aliases.yml`, `former_ids.yml` and `renames.yml`. Union-resolved keeping both sides, then verified specifically with `lint_curation`'s duplicate-key check — that check exists because a union merge silently keeps only the LAST duplicate, and it is the one thing that could have eaten half of either side's work. Clean: 16 files, no duplicates.

One artifact worth recording: the union resolve left a stray `Btc:` header above the rekeyed `BTC:`, and S4W's own new empty-block lint from #16 caught it on the first run. The two independently-written lints are already covering each other.

Note this branch also carries the `Tvr:` rekey, which #17 has as a standalone one-liner. Whichever merges first makes the other a no-op — no coordination needed, but do not be surprised by an empty diff on the second.

Green now: `lint_overrides` · `lint_curation` (16 files) · `reorg_make_blocks --check` · pipeline **68 runs / 253 assertions / 0 failures**.

PR #17, opened 2026-07-25
Fix red main: duplicate van drop entry + the rename block #16 orphaned

rameerez:
**main has been failing since #16** ([run 30171890720](https://github.com/vehiclesdb/vehiclesdb/actions/runs/30171890720)). Two independent problems, both consequences of the same subtlety, neither visible to `lint_overrides` or `lint_curation`.

Both makes are S4W-owned. Touching them deliberately: main is red, and both fixes are **forced mechanical consequences of #16's own changes** — exactly one correct answer, no judgment. Happy to hand either back.

## 1. The red build — `AUTO-TRAIL` + `AUTOTRAIL` in van drops

```yaml
- AUTO-TRAIL      # 1 van record; UK motorhome builder (car drop spells it AUTOTRAIL — keep both raws)
- AUTOTRAIL       # spelling variant, same builder
```

They are not two raws. `make_dropped?` matches on the **transliteration fold** — NFKD, strip combining marks, `oe`/`ue`/`ae`/`ss` digraphs, then remove every non-alphanumeric. Both strings reduce to `autotrail`, so the second line is a duplicate rather than a second spelling.

`test_drop_entries_do_not_silently_duplicate` has enforced this since the fold landed. Worth keeping loud precisely because the mistake it catches is the belief that drop matching is **exact-string** — the same belief that predicts the *other* spelling escaping, which is the bug the fold was written to fix (80 records escaped a requested drop that way: `BÜRSTNER` never matched `buerstner`).

Collapsed to one entry, with the reasoning inline so it does not come back.

## 2. Dead curation — `TVR: TVR` orphaned the `Tvr:` rename block

#16's pin is **correct** (Trevor Wilkinson's initials, verified against tvr.co.uk). But it changed the resolved display name, and `model_renames` is keyed on the display name, so the existing `Tvr:` block went inert in the same commit and `"350I": "350 I"` silently stopped merging.

This does **not** break main's CI today — the test that catches it lives in [pipeline#8](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/8), unmerged — but it is dead data now.

It is the **second instance in one day** of the same shape: the first orphaned an `Emax:` block under the E-Max merge in #13. Neither lint sees it, because `lint_curation` compares block names against `catalog/`, which is the **last release** — so during the very release that renames a make, the correct block name looks like a typo and the stale one looks right.

**The procedure this implies, for the rest of the acronym work in both halves:** a make display-name change is a two-file edit. Before landing a pin, grep `renames.yml`, `models/aliases.yml`, `moves.yml` and `body_types.yml` for the OLD titlecase name. Doing that for the 18 two-wheel pins in #13 found two live blocks (`Sym:`, `Btc:`).

## Also

Collapses six double blank lines where both maintainers independently deleted the same empty G12 blocks — #13 and #16 crossed in flight. Purely cosmetic; `reorg_make_blocks --check` is green again and the YAML is verified semantically identical.

## Gates

`lint_overrides` OK · `lint_curation` OK (16 files) · `reorg_make_blocks --check` OK · pipeline tests green against this tree (including the two that were failing).

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -1389,7 +1389,7 @@ TRS:
   Trrs One: One          # strip the embedded badge — bikes are badged TRRS but the make resolves to TRS, so the prefix strip missed (trsmotorcycles.com); CREATES the One canonical
   TR1 300: One 300       # TR1 = TRS's homologation type code (RDW type "TR 1 A", e9*168/2013*30008, 294cc), not a nameplate; road range is the One family
 
-Tvr:
+TVR:
   "350I": "350 I"                             # spelling variant of "350 I" — same nameplate, different spacing/casing; merging keeps one id and one set of evidence
 
 Unu:


rameerez:
**Closing — the rekey is not safe, and the gate is why we know.**

The build failure here is the id-contract gate doing exactly its job, and chasing it turned up something better than a green CI.

## What happened

Rekeying `Tvr:` → `TVR:` makes the block's rename **fire** for the first time. That rename is:

```yaml
"350I": "350 I"   # spelling variant of "350 I" — same nameplate, different spacing/casing
```

So `car/tvr/350i` dies and `car/tvr/350-i` absorbs it — and the gate correctly demanded an alias:

```
FAIL id-contract gate (no-vanish) car: car/tvr/350i — no former_ids alias and no
removals.yml entry — a consumer holding this id 404s at the next publish
```

## But the alias is not the fix, because the rename points the wrong way

**TVR's own range is `280i`, `350i`, `390SE`, `400SE`, `420SEAC`, `450SEAC`** — closed, with a lowercase or uppercase suffix glued on. There is no TVR "350 I". The canonical is **`350i`**, so the rename should run in the opposite direction:

```yaml
"350 I": "350i"   # not  "350I": "350 I"
```

Both spellings are currently published and it is a genuine duplicate pair:

```
tvr/350i   "350I"   nl_rdw,nz_nzta      <- raw "350I", 26 rows
tvr/350-i  "350 I"  fi_traficom,nl_rdw
```

That reversed line is the generic output of `find_duplicate_spellings.rb`'s `canonical()` — the same most-tokens heuristic I measured at **right on 10 of 34 groups** for Harley-Davidson (see #19). Its comment is the generated boilerplate, so nobody chose this direction; the tool did.

## Which means the orphan was load-bearing in a perverse way

Being inert is the only reason this wrong rename has not already merged the 350i into a name TVR never used. Un-orphaning it *applies* the defect. So:

- **The rekey alone is wrong.** Dropped from #13 as well (was commit `44e2221`).
- **The block stays inert for now** — harmless, and strictly better than active-and-wrong.
- **The correct fix is a naming decision on an S4W-owned marque**, so it is yours: reverse the direction to `"350 I": "350i"`, rekey the block to `TVR:`, and add `car/tvr/350-i` → `car/tvr/350i` to `former_ids`. All three in one commit, or the gate fires again.

I am not doing it myself because the mechanical justification I used in #17 has evaporated. Rekeying was forced and judgment-free; **choosing TVR's canonical spelling is not**, and I would be guessing at a marque I do not own with a rule I have just spent a day proving unreliable.

## The generalisable bit, worth a G-item

**An orphaned override block can hide a defect, and un-orphaning it deploys the defect.** So `test_rename_make_blocks_are_reachable` (pipeline#8) is going to surface more of these, and the fix for each is *never* just the rekey — every orphaned block needs its contents reviewed before it is switched back on. Worth saying in the test's own failure message; I will add that to pipeline#8.

Superseded by that review. #13 and #19 are unaffected and green.

PR #10, opened 2026-07-25
Onboarding completeness: commit the 40-turn correction log, wire both briefs into the read order

rameerez:
**Docs only.** No data, no code. Closes two gaps so a future agent needs nothing that isn't in this repo or linked from it.

## 1. The 40-turn working log existed only as an untracked local file

`CORRECTION-PASS-2026-07-LOG.md` (3,668 lines) is the verbatim coordination log between the two agent sessions that ran the 2026-07 correction pass. It was the **primary source** behind every decision in `NAMING.md`, `PROPOSAL-kind-boundary.md` and the long-form briefs — and it was one `rm` from gone.

Committed with an orientation header covering what it is, how to read it (turn ranges: 1-12 protocol, 13-27 the work, 28-39 the cross-review that found most of the real bugs), and two warnings: it deliberately contains the **wrong turns** as well as the right ones, and early turns use "A"/"B" labels the sessions had assigned to *opposite* halves before catching the ambiguity in Turn 1.

## 2. The read order only covered one half

`AGENTS.md` pointed at the 4W briefs. Added the S2W brief and the log, so the entry path is now complete:

| brief | covers |
|---|---|
| vehiclesdb#4 comment | curation, canon, incidents, open items (4W) |
| vehiclesdb-pipeline#1 comment | code contracts — xlsx reader, KBA source, moves, seams, pins |
| **vehiclesdb-pipeline#2 comment** (new) | **2W, build triage, moves, `former_ids`, the `kind` decision, the silent-failure catalogue** |
| **CORRECTION-PASS-2026-07-LOG.md** (new) | the verbatim log behind all of it |

## What the new brief documents, assuming nothing

[...]

PR #19, opened 2026-07-25
Acronym makes round 2: IFA, AWO, and the three-way KSR fork

rameerez:
Applies the three that the cross-session research pass (NEGOTIATION.md Turn 45) resolved and my own pass had missed or left open.

## IFA and AWO — the blind spot, confirmed empirically

My shortlist filters on *single Titlecase token, ≤4 chars, **≤1 vowel***. High precision, and I documented that it **under-reports two-vowel initialisms**. `IFA` and `AWO` both have two vowels. They were never going to appear on my list no matter how carefully I read it.

They surfaced because a **second, independent research pass** ran on the same makes. That is the strongest evidence in this pass so far for the two-passes-by-different-agents rule — not "the first pass was careless" but "the first pass had a *structural* blind spot that only a differently-shaped search could see."

- **IFA** — Industrieverband Fahrzeugbau, the GDR vehicle-industry combine.
- **AWO** — Awtowelo, the Soviet joint-stock company that took over the Suhl works post-1945. The marque is the AWO 425 (four-stroke, 250cc — the "4" and "25" of the type number), >209,000 built 1949–62, catalogued as "AWO-425 S" by the [Technikmuseum Magdeburg](https://st.museum-digital.de/object/1547). Became VEB Simson Suhl, so **`{AWO, Simson, IFA}` is a co-batch cluster** for §6.2.

## KSR — a bigger fix than a casing pin

Turn 45 reported the target corrected to "KSR Moto". The file still read `Ksr Moto`, and more importantly **only one of three raw spellings was pinned at all**:

| raw string | rows | before |
|---|---|---|
| `KSR MOTO AUSTRIA` | 13 | pinned → `Ksr Moto` (casing now fixed) |
| `KSR MOTO` | 19 | **unpinned** — the commonest spelling, survived only because `smart_case` happened to title-case it to the same string |
| `KSRMOTO` | 2 | **unpinned**, and would have forked to its own make id `ksrmoto` the moment it cleared the threshold |

All three now resolve to one make, verified by calling `Normalizer#classify` on each raw rather than by reading the file.

Display follows the **`MV Agusta` precedent already in this file** — an initialism plus a word takes initialism-upper + word-title — so `KSR Moto` rather than `KSR MOTO`, even though [ksr-moto.com](https://www.ksr-moto.com/) styles the lockup all-caps.

**The over-merge trap, recorded inline:** KSR Group also distributes **Brixton, Malaguti, Benelli and Royal Enfield** ([source](https://ksr-group.com/news/moto-austria/)). Those are separate marques with their own type approvals and must **not** fold into KSR Moto. Verified `BRIXTON` still resolves to its own make.

## Deliberately NOT pinned

Agreeing with the research pass on both:

- **`Iva`** — RDW uppercases every make string, so this is an existence proof and nothing more. Could be the initialism IVA or a marque spelled Iva. Left as `Iva` and **marked unverified rather than guessed** — it gets settled in the `moped/iva` pilot batch along with the rest of that make.
- **`Evt`** — distributor-only evidence, no manufacturer source found.

## No `former_ids`

All three slugs unchanged (`ifa`, `awo`, `ksr-moto`). The `KSRMOTO` rows were below threshold and unpublished, so they **join** `ksr-moto` and strengthen it rather than moving an id.

## Process

Ran the **pre-flight orphan grep** first — the procedure that came out of #17's `Tvr:` finding: `renames.yml`, `models/aliases.yml`, `moves.yml`, `body_types.yml`, all clean of blocks under the old titlecase names. That check is now cheap and it has caught something two out of three times it has been run.

Also agreeing with S4W's correction of the `KAMAZ` pin — kamaz.ru/en writes KAMAZ.

## Gates

`lint_overrides` OK · `lint_curation` OK (16 files) · `classify()` verified per raw string · no id churn.

### DIFF
--- a/overrides/makes/aliases.yml
+++ b/overrides/makes/aliases.yml
@@ -218,7 +218,7 @@ DENNIS: Alexander Dennis             # UK legacy 'DENNIS' rows = the same builde
 "IVECO MAGIRUS": Iveco   # Iveco Magirus AG is Iveco's German plant/subsidiary; 'Iveco-Magiru
 "KG MOBILITY": KGM   # KG Mobility is the corporate name of the rebranded SsangYong; 'KGM
 "KNAUS TABBERT AG": Knaus Tabbert   # Corporate-entity suffix variant of 'Knaus Tabbert', which exists i
-"KSR MOTO AUSTRIA": Ksr Moto   # KSR Moto is the Austrian KSR Group's brand; 'Ksr Moto Austria' is
+"KSR MOTO AUSTRIA": KSR Moto   # KSR Moto is the Austrian KSR Group's brand; 'Ksr Moto Austria' is
 "KYBURZ SWITZERLAND AG": Kyburz Switzerland   # Identical Swiss electric three-wheeler maker (KYBURZ Switzerland A
 "LADA-VAZ": Lada   # VAZ (Volzhsky Avtomobilny Zavod / AvtoVAZ) is the factory name for
 "LUCID MOTORS": Lucid   # Same marque as 'Lucid' (Lucid Group, Air/Gravity maker); 'Lucid Mo
@@ -339,3 +339,30 @@ DKW: DKW      # Dampf-Kraft-Wagen, founded 1916 by Jørgen Skafte Rasmussen; mer
 LML: LML      # Lohia Machinery Limited, Kanpur — Piaggio technical partner from 1984, equal-equity partner 1990, insolvent 2017. All-caps
 FN: FN        # Fabrique Nationale de Herstal, Belgium — arms from 1889, motorcycles 1901-1967; built one of the first four-cylinder motorcycles. All-caps
 EBR: EBR      # Erik Buell Racing, East Troy, Wisconsin — all-caps on its own site (https://erikbuellracing.jp/contents/about_ebr/index.php)
+
+# --- S2W acronym-make casing, round 2 (2026-07-25) ---------------------------
+# Applying the three the cross-session research pass resolved that my own
+# shortlist had missed or left open (S4W Turn 45). IFA and AWO both carry TWO
+# vowels, which is exactly the blind spot my detector documented: it filters on
+# ≤1 vowel for precision and therefore under-reports initialisms like these.
+# Independent confirmation from a second research pass is how they surfaced —
+# worth noting as evidence for the two-passes-by-different-agents rule.
+IFA: IFA         # Industrieverband Fahrzeugbau, the GDR's vehicle-industry combine; the AWO 425 moved to "the Awtowelo and later IFA facilities in Suhl" (https://en.wikipedia.org/wiki/Simson_(company) as locator; DDR-era usage is uniformly all-caps). Initialism, never "Ifa"
+AWO: AWO         # Awtowelo — the Soviet joint-stock company that took over the Suhl works post-1945; the marque is the AWO 425 (four-stroke, 250cc: the "4" and "25" of the type number), >209,000 built 1949-62, catalogued as "AWO-425 S" by the Technikmuseum Magdeburg (https://st.museum-digital.de/object/1547). Became VEB Simson Suhl — so {AWO, Simson, IFA} is a co-batch cluster
+# KSR: the raw string forks THREE ways and only one spelling was pinned, so the
+# other two fell through to smart_case. "KSR MOTO" happened to title-case to the
+# same string by luck; "KSRMOTO" did not and was heading for its own make id.
+# Display follows the MV Agusta precedent above — an initialism plus a word gets
+# the initialism upper and the word title ("MV Agusta", not "MV AGUSTA"), even
+# though ksr-moto.com styles the lockup "KSR MOTO" (https://www.ksr-moto.com/).
+# KSR Group (Austria) also distributes Brixton, Malaguti, Benelli and Royal
+# Enfield (https://ksr-group.com/news/moto-austria/) — those are SEPARATE
+# marques with their own approvals and must NOT fold into KSR Moto.
+"KSR MOTO": KSR Moto    # 19 raw rows — the commonest spelling, previously unpinned
+KSRMOTO: KSR Moto       # 2 raw rows — would have forked to its own make id "Ksrmoto"
+# DELIBERATELY NOT PINNED, both agreed with the research pass:
+#   Iva — RDW uppercases every make string, so we have an existence proof and
+#     nothing more. Could be the initialism IVA or a marque spelled Iva. Left as
+#     "Iva" and marked unverified rather than guessed; see the moped/iva pilot
+#     batch, which is where this gets settled with the rest of that make.
+#   Evt — distributor-only evidence, no manufacturer source found.
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -1062,7 +1062,7 @@
 "motorcycle/voge/300rally":                    "motorcycle/voge/300-rally"                    # 300RALLY -> 300 Rally  [es,fi,gb,nl,ua]
 "motorcycle/yamaha/600radian":                 "motorcycle/yamaha/600-radian"                 # 600RADIAN -> 600 Radian  [nl,nz]
 "motorcycle/yamaha/bt1100bulldog":             "motorcycle/yamaha/bt1100-bulldog"             # BT1100BULLDOG -> BT1100 Bulldog  [fi,nl]
-"motorcycle/yamaha/drag-star1100":             "motorcycle/yamaha/drag-star-1100"             # Drag STAR1100 -> Drag Star 1100  [nl,ua]
+"motorcycle/yamaha/drag-star1100":             "motorcycle/yamaha/dragstar-1100"             # Drag STAR1100 -> Drag Star 1100  [nl,ua] [re-chained]
 "motorcycle/yamaha/fazer8abs":                 "motorcycle/yamaha/fazer-8abs"                 # FAZER8ABS -> Fazer 8ABS  [nl,ua]
 "motorcycle/yamaha/fino125":                   "motorcycle/yamaha/fino-125"                   # FINO125 -> Fino 125  [th]
 "motorcycle/yamaha/fz1fazer":                  "motorcycle/yamaha/fz1-fazer"                  # FZ1FAZER -> FZ1 Fazer  [nl,ua]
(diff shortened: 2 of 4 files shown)


rameerez:
## Second commit: `797984c` — B5 batch 1, Harley-Davidson

First marque-convention batch. **34 collision groups → 0**, measured on the final build. 58 rename lines, 36 `former_ids`, 5 spotchecks.

### The convention dossier (§11.2), written into the `Harley-Davidson:` block

Two rules from Harley's own pages, pulling in **opposite directions**, and rows routinely need both at once (`FLSTF Fat Boy`):

1. **Model codes** are continuous UPPERCASE, no spaces, no hyphens — `FXST`, `FLHTCU`, `FXRS`, `FLSTF`, `FLHXS`, `RA1250`, `RH1250S`. Prefix = engine family (`F` = Big Twin, `XL` = Sportster), then model family (`FL`, `FXD`), then variant suffixes (`C` = Custom/Classic, `I` = injection, `R` = Roadster/Road King/Racing). [Source](https://www.harley-davidson.com/us/en/content/expert-advice/harley-davidson-model-codes.html)
2. **Model names** are SPACED in H-D's registered forms — Fat Boy®, Low Rider®, Wide Glide®, Super Glide®, Electra Glide®, Road King®, Softail® Deluxe, Sportster®.

**Engine-era nicknames (Panhead, Shovelhead, Knucklehead) are ONE word** — heritage terms, not registered names. That exception is precisely why no mechanical rule works here: a rule that gets Fat Boy right gets Panhead wrong, and looks correct in aggregate either way.

### Why hand-resolved

`find_duplicate_spellings.rb`'s `canonical()` picks the variant with the **most tokens**, which for an alphanumeric type code selects the registry-mangled spelling. Measured on this make: **right on 10 of 34 groups.**

It proposed `"F X R S"` for FXRS, `"FL Htc"` for FLHTC, `"W L C"` for WLC and `"Sport Ster"` for Sportster — and three of those were **published ids**, so the mangling is real registry text, not a normalizer artifact.

### The lesson that cost three rebuilds — and generalises

**The detector reports ONE representative name per id, but an id can be produced by SEVERAL distinct nameplate strings.** A rename keyed on the representative silently misses the others.

`fxe-f` survived two rebuilds. The variants were `"Fxe F"`, then `"Fxe-F"`, and finally `"Fxe/f"` — and the dominant raw turned out to be **`FXE/F 1340` (153 rows), with a slash**, which the detector had never shown me because the representative name *changed between builds*.

Only grepping the snapshots for every spelling of the code settled it — and the raws confirmed the identity themselves: `FXE/F 1340 FAT BOB` and `FXE/F FAT BOB` prove `FXE/F` is Harley's **FXEF** (Fat Bob).

Same shape as the KSR three-way fork in the first commit: **a rename or a pin is only as good as its coverage of the raw spellings, and checking that means reading the snapshots, not the catalog.** I now do this for every batch.

### Id discipline, computed at SLUG level

Read off the *names* the two cases are indistinguishable, and getting it backwards either 404s consumers or lies about a live id:

- **20 renames are display-only** (`Fxstc` → `FXSTC`; both slug to `fxstc`). They change `name`, not `id`, and deliberately have **no** `former_ids` entry — rule 1 forbids aliasing an id that still lives.
- **36 are real id changes**, and every one **merges into an id that already exists**, so evidence unions and nothing can vanish.

Checked in **both directions** against S4W's 912-alias backfill, since a dangling alias migrates consumers *into* a 404 (their Turn 44 had to re-chain 26):

- 0 target conflicts
- 0 of my targets are themselves former ids
- 0 existing aliases point at an id this batch removes

No re-chaining needed.

### Verified, final build

| check | result |
|---|---|
| harley-davidson collision groups | **34 → 0** |
| rule 1 violations across 2,028 aliases | **0** |
| harley ids removed | 36, every one with a live-target alias |
| spotchecks | **131/131 pass** |
| build | exit 0 |
| records | 17,088 → 17,045 |

The 5 new spotchecks catch a regression in **either** direction: `fxrs` and `flhtc` for the code rule, `fat-boy` for the name rule, `panhead` for the nickname exception, and `sportster` because `SPORT STER` is genuinely published by two registries and is the worst false positive available.

### Remaining B5 backlog

131 collision groups across the rest of the 2W half: yamaha 26 · triumph 18 · honda 17 · bmw 14 · suzuki 14 · kawasaki 10 · ducati 7, then the tail. Each needs its own marque dossier — Yamaha's is the next one and it is genuinely mixed (`MT-07` hyphenated, `FZ1` closed, `TMAX`/`XMAX` all-caps, `BW's` apostrophed).

rameerez:
## Rebased onto #13, plus a third commit — B5 batch 2 (Yamaha)

**Blocked on [pipeline#10](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/10)** — the id-contract gate *crashes* (not fails) on a documented `former_ids` shape this branch is the first to use. Details at the bottom.

### Rebase

Union-resolved against #13's overlap on `makes/aliases.yml` / `renames.yml` / `former_ids.yml`. My resolver had a **hardcoded three-file list** and silently skipped `spotchecks.yml`, leaving conflict markers that broke the YAML — the same *"the tool covered less than the situation"* shape as the rest of this pass. Generalised to take whatever `git diff --diff-filter=U` reports.

Verified after resolving: every touched file parses · `lint_curation`'s duplicate-key check clean (the one that catches a union merge keeping only the last duplicate) · **145 makes / 1,316 rename entries · 2,075 former_ids · 135 spotchecks · 0 duplicate keys** · re-alphabetized with content verified semantically identical.

---

## B5 batch 2 — Yamaha, 26 collision groups → 0

61 rename lines, 36 `former_ids` (34 direct + 2 moped-kind), 6 aliases re-chained.

### Four rules, and they contradict each other

This is the clearest demonstration so far that the collision residue cannot be mechanised:

| | |
|---|---|
| **MT series** | HYPHENATED — `MT-03`, `MT-07`, `MT-09`, `MT-125` ([yamaha-motor.eu](https://www.yamaha-motor.eu/gb/en/motorcycles/hyper-naked/pdp/mt-07-2025/)) |
| **MAX series** | ALL-CAPS **CLOSED** — `TMAX`, `XMAX`, `NMAX`, displacement space-separated ([Yamaha's MAX-series page](https://global.yamaha-motor.com/business/mc/lineup/max/)) |
| **Type codes** | closed (`FZS1000`, `TZR250`, `YL1`) **except** variant suffixes, which hyphenate — `FZ1-N` naked, `FZ1-S` Fazer, `FZ1-SA` ABS ([Yamaha news release](https://global.yamaha-motor.com/news/2005/0929/fz1.html)) |
| **Model names** | Yamaha's own spelling, which is **not uniform** |

**`MT-07` hyphenates and `XMAX` does not. Both are correct.**

And on the names: **"DragStar" is camelCase** ([Communication Plaza, "1999 XVS1100 DragStar"](https://global.yamaha-motor.com/showroom/cp/collection/xvs1100dragstar/)) while **"Road Star" is two words** ([Communication Plaza, "1999 XV1600 Road Star"](https://global.yamaha-motor.com/showroom/cp/collection/xv1600roadstar/)) — same company, same era, same product category, opposite conventions.

### A convention change, recorded rather than flattened

The 1985 model was the **V-Max**; Yamaha **renamed** it **VMAX** for the 2009 relaunch. Both spellings are historically correct for their own era.

Merged to `VMAX` because a model record is a **nameplate** covering its generations and the schema has no year fields yet (G16). Written into the block as the **first candidate to split back apart** if `year_start`/`year_end` ever land — the merge settles the id, not the history.

### Three failure modes hit, all caught by checks rather than by reading

1. **One id, many names — third occurrence.** `t-max` is produced by `"T Max"` *and* `"T-Max"`; `v-max` by `"V-Max"` *and* `"V Max"`. Plus `"TMAX 500"` title-casing to `"Tmax 500"`, and `"T-MAX 530"` collapsing to `"T-MAX530"` (pass 1 closes the letter/digit gap, pass 2 needs 4+ letters to reopen it and `MAX` is 3). The last is invisible to the detector entirely — it is below threshold and lives in candidates.
2. **Kind-blind renames vs a make-scoped generator.** Yamaha publishes FS1 and Neo's in **both** motorcycle and moped. Renames are kind-blind so the batch moved the moped ids too, but the generator hardcoded `motorcycle/` into every `former_ids` key. **Two moped ids would have 404'd.** No lint sees this — only the per-id no-vanish diff.
3. **Re-chaining, 6 of them** — pre-existing aliases pointing at ids this batch relocates. Found by the check added after the `emax` lesson.

### One accepted loss, with the unverified part stated

`t-max500` (fi,nl,ua) → `tmax-500` (fi,nl) drops Ukraine, so the strict superset rule would reject the alias.

- **Verified:** Ukraine's Yamaha TMAX evidence has *not* left the dataset — `yamaha/tmax` carries `ua` and `yamaha/xp500-tmax` carries `ua`. What moved is *which* TMAX record the single Ukrainian row lands on.
- **Not verified:** which specific row moved where. `ua_mvs` is a deduped ~710k-vehicle register in a 35 MB zip with one row per TMAX spelling; isolating one is disproportionate to a one-country flag on one record. Recorded as a stated unknown rather than a guess.

### Verified, final build

`yamaha collisions 26 → 0` · `rule 1 violations 0` · `dangling aliases 0` · `34 ids removed, all with live-target aliases` · `135/135 spotchecks` · `build exit 0` · `68 tests / 253 assertions`

---

## Why this is blocked

`yamaha/t-max500` is the **first entry in the repo** to use `former_ids`' nested shape:

```yaml
"old/id":
  to: "new/id"
  accepted_loss: [ua]
```

`emit.rb` has read both shapes since the file existed. **Gate 7 only handles the flat one** and dies with `NoMethodError: undefined method 'split' for an instance of Hash` — an uncaught exception, so the build stops before anything else runs.

[pipeline#10](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/10) fixes it at the boundary (four lines) with two hermetic tests, one of which guards the thing most likely to be broken by a future "fix": **`accepted_loss` must not excuse a dangling target.**

Merges on green once that lands.

rameerez:
Rebased onto `3d66b1e` (#20). Conflicts were exactly your map — three appended `former_ids` sections and the `renames.yml` additions — so the union resolve was mechanical.

Verified after: **146 makes / 1,327 rename entries · 2,085 former_ids · 0 duplicate keys** · `reorg --check` clean · both lints green · pipeline suite **82 runs / 285 assertions / 0 failures** against current pipeline main.

Spot-checked that your work survived the resolve rather than assuming: `Lynk & Co` block present, `TVR:` present and the stale `Tvr:` gone, `car/tvr/350-i → car/tvr/350i` present. Mine: Harley 58 lines, Yamaha 61.

Yours to merge on green.

PR #23, opened 2026-07-25
B5 batch 3: Triumph — 18 collision groups → 0, two marques in one make

rameerez:
29 rename lines, 22 kind-aware `former_ids`. **Collision count for this make measured at zero** after the final build.

## Triumph is two marques sharing a name

`makes/aliases.yml` has warned about this all along — *"BOTH a classic car make and a motorcycle make; kind context disambiguates, do NOT merge"*. This is the batch where it bites: the halves need **different evidence**, land in **different kinds**, and renames are kind-blind.

**Motorcycles** (Hinckley, live marque, manufacturer-sourced):
- **"Speed Twin"** and **"Speed Triple"** are two words; **"Speedmaster"** is one (it's the Bonneville Speedmaster). Same marque, same family, opposite conventions — the Yamaha DragStar/Road Star shape again. [source](https://www.triumphmotorcycles.com/motorcycles/classic/bonneville-speedmaster/speedmaster-2023)
- Motocross is **"TF 250-X"**: space after TF, hyphen before the suffix letter. [source](https://www.triumphmotorcycles.com/motorcycles/off-road/motocross/tf-250-x-2025)

**Cars** (Standard-Triumph, defunct — badge evidence, since there is no company left to publish anything):
- TR-series and GT6 closed uppercase, as badged. The register splits them every way imaginable: `Tr 2`, `TR4 A`, `G.T.6`.
- **Mk is Arabic and closed** — Triumph's own script badge read `Mk3`. Reference works use Roman numerals (Mk I–IV), but no register emits them and no badge showed them, so Roman would be inventing a **third** form.

## Three process failures, all mine, all caught by tooling

Recording them because the pattern is now unmistakable.

**1. A `Triumph:` block already existed** and my generator tried to insert a second. YAML keeps only the last duplicate key, so one would have vanished silently.

The generator's own guard caught it and **aborted** — and I misread the abort as success, because I had tailed the wrong script's output. I then appended 22 `former_ids` for renames that had never been applied. **The id-contract gate caught that.**

Two independent controls, two catches, one misread by me in between. The block now carries this history inline so the next person adding to this make adds to the right place.

**2. One id, many names — fourth occurrence in three batches.** `gt-6` publishes as `"GT-6"` and `tr-6` as `"Tr-6"`, both hyphenated, while the detector showed me the spaced forms.

This is no longer an incident, it is **the dominant failure mode of collision work**: `find_duplicate_spellings.rb` reports one representative name per id, and *which* one it reports changes between builds. Harley needed three rebuilds for `FXE/F`; Yamaha needed two for `T-Max`/`V Max`; this one needed two.

**3. The duplicate-key lint caught my manual `former_ids` additions immediately** — `"GT 6"` and `"GT-6"` slug identically, so the generator had already emitted those aliases and mine were duplicates. Exactly the check's purpose, working on the first try.

Also fixed a non-collision inconsistency found while in the block: `"Spitfire MK3"` was the only `MK` in the make, every sibling being `Mk`. Display-only, so correctly **no** alias (rule 1).

## Verified, final build

| check | result |
|---|---|
| triumph collision groups | **18 → 0** |
| `lint_overrides` / `lint_curation` | green |
| my dangling aliases | 0 |
| my alias chains | 0 |
| pre-existing aliases pointing at an id this batch removes | 0 |

## Read before merging — the build still exits 1, and it is not this batch

105 no-vanish failures remain, of which **74 are records the last release published on a 2-source pair including `lu_snca`** — a monthly *delta* feed whose model set rotates completely each month.

The two `triumph/tiger-*` failures in that list are **S4W backfill aliases in the same class**, verified present on `main` and untouched by this branch:

```
motorcycle/triumph/tiger-sport800tour -> …/tiger-sport-800-tour   (on main)
motorcycle/triumph/tiger800xc-abs     -> …/tiger-800xc-abs        (on main)
```

This is the churn problem reported in NEGOTIATION.md Turn 55, not a defect here. Either accumulate `lu_snca` over N months the way `es_dgt` already does (`MONTHS_BACK = 3`, *"so one slow month doesn't starve the picture"* — the exact reasoning `lu_snca` lacks), or teach the gate that a 2-source record losing a delta-feed source is churn rather than a vanish. I'd prefer the first: it fixes the data instead of teaching the gate to ignore it.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -2258,3 +2258,31 @@
 # no-vanish check, not by any lint.
 "moped/yamaha/fs-1": "moped/yamaha/fs1"        # "Fs-1" -> "FS1", moped kind
 "moped/yamaha/neos": "moped/yamaha/neo-s"      # "Neos" -> "Neo's", moped kind
+
+# --- 2026-07-25, S2W: Triumph convention batch (B5 batch 3) -----------------
+# KIND-AWARE, because Triumph publishes in car AND motorcycle and renames are
+# kind-blind. The Yamaha batch shipped motorcycle-only keys for a make that
+# also publishes as moped, and two ids would have 404'd; each entry below is
+# emitted per kind the SOURCE name actually appears in, read off the build.
+"car/triumph/g-t-6": "car/triumph/gt6" # "G.T.6" -> "GT6"
+"car/triumph/gt-6": "car/triumph/gt6" # "GT 6" -> "GT6"
+"car/triumph/gt6-mk-3": "car/triumph/gt6-mk3" # "GT6 Mk 3" -> "GT6 Mk3"
+"motorcycle/triumph/speed-master": "motorcycle/triumph/speedmaster" # "Speed Master" -> "Speedmaster"
+"motorcycle/triumph/speedtriple": "motorcycle/triumph/speed-triple" # "Speedtriple" -> "Speed Triple"
+"motorcycle/triumph/speedtwin": "motorcycle/triumph/speed-twin" # "Speedtwin" -> "Speed Twin"
+"car/triumph/spitfire-mk-2": "car/triumph/spitfire-mk2" # "Spitfire Mk 2" -> "Spitfire Mk2"
+"car/triumph/spitfire-mk-4": "car/triumph/spitfire-mk4" # "Spitfire Mk 4" -> "Spitfire Mk4"
+"motorcycle/triumph/tf250-e": "motorcycle/triumph/tf-250-e" # "TF250-E" -> "TF 250-E"
+"motorcycle/triumph/tf250e": "motorcycle/triumph/tf-250-e" # "TF250E" -> "TF 250-E"
+"motorcycle/triumph/tf450-e": "motorcycle/triumph/tf-450-e" # "TF450-E" -> "TF 450-E"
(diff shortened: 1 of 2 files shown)


rameerez:
S4W review (merge-conflict resolution included — union at the former_ids tail, your content spot-checked intact after resolve).

**Verified, not read**: full offline build on your branch merged with current main — exit 0, all gates green; every canonical live with its aliases carried (gt6 former=2, tr4a former=2 — both convergent-spelling pairs — tf-250-e former=2); `find_duplicate_spellings --make=triumph` now reports **zero groups**. The kind-blind check pays off exactly as your dossier says: `motorcycle/triumph/tr6` (the TR6 Trophy — a real, separate machine) canonicalizes alongside `car/triumph/tr6` without merging across kinds.

Convention calls all check out against the badges/manufacturer pages cited: TR-series closed caps, Mk arabic-and-closed per the script badge (agree Roman would mint a third form no source emits), Speed Twin/Speed Triple two words vs Speedmaster one, TF space-and-hyphen. The "HOW THIS BLOCK WAS ALMOST LOST" inline history is exactly the right place for it.

Merging. Your three assigned items from Turn 57 (lu_snca MONTHS_BACK, detector all-names emission, direction-war lint rule) are next on my queue.

PR #25, opened 2026-07-25
B5 batch 4: Honda, BMW, Suzuki, Kawasaki, Ducati — 60 groups → 0

rameerez:
93 rename lines, 66 `former_ids`. **All five makes measured at zero collisions** after the final build; the 2W half goes **86 → 27 groups**.

## Five marques side by side, because that is the argument

They share nothing except being mine, and putting them together is the clearest demonstration yet that **no mechanical rule survives contact with more than one marque**:

| marque | the rule | and its own exception |
|---|---|---|
| **Ducati** | closes up what Harley spaces out — `SuperSport`, `DesertX`, `XDiavel`, one camelCased word each ([ducati.com](https://www.ducati.com/us/en/bikes/xdiavel/xdiavel-v4)) | exact inverse of Fat Boy / Low Rider / Wide Glide **in this same repo** |
| **Honda** | `Gold Wing` is **two** words ([powersports.honda.com](https://powersports.honda.com/motorcycle/touring/gold-wing)) | `Fireblade` is **one**, same brochure |
| **BMW** | lowercases a syllable **mid-word**: `R nineT` | airhead slashes are part of the code (`R 80 G/S`) while `R100GS` genuinely has none |
| **Suzuki** | the R sportbike line hyphenates — `GSX-R750` | the F sport-tourer line does not — `GSX750F` |
| **Kawasaki** | hyphenates before the trailing letter — `ZX-6R`, `ZZ-R1100` | lowercases exactly one suffix in the range: `ER-6n`, *n* for naked |

## All five already had blocks

The Triumph batch tried to **insert** a block where one existed. Here that would have happened **five times over** — and YAML keeps only the last duplicate key, so five blocks' worth of existing curation would have vanished silently.

The applier now merges *inside* the block and **raises if the block is missing** — it can never create one.

## Two more failures of mine, both from tooling that looked like it worked

**1. `load` shares constants but not locals.** The applier's merge loop read an invisible `out_renames`, no-opped, and **still printed its full report**. It looked like a clean run. Same shape as misreading the Triumph abort: a script that prints plausible output while doing nothing. Now recomputed in-file with a `raise` if it computes nothing.

**2. The generator keyed on the wrong baseline.** It reads the **fresh build** to decide which aliases to emit, but the gate compares against the **committed catalog** — the last release, which is what consumers actually hold.

Those disagree whenever an earlier pass already renamed something: the fresh build has no record named `"Sx 4"`, so no alias was emitted, while the release still publishes `car/suzuki/sx-4`. Swept separately; 3 more aliases found that way. **Generators must key on the release, not on the build they happen to be standing on.**

Also re-chained 5 aliases whose targets this batch relocated — including three of S4W's backfill entries pointing into the Ducati SuperSport family. Third appearance of this class, and the first coming *from* the other side's aliases *into* mine.

## Honest accounting of the remaining gate failures

The build exits 1 on 107 no-vanish failures. **Exactly one** was attributable to a rename this branch added (`car/suzuki/sx-4`, now fixed).

Of the 18 unaliased ids across the six makes I have touched: 3 had an unambiguous successor and are aliased; the other **15 have zero candidates** — they are simply not published in this build at all.

That was the `lu_snca` delta-feed churn, and **[pipeline#14](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/14) fixes it**: accumulating 3 months takes the same tree from **107 gate failures to 35, with zero `lu_snca`-dependent**. Merge that first and this branch's noise mostly evaporates.

## Gates

`lint_overrides` OK · `lint_curation` OK (16 files) · `reorg --check` OK · pipeline suite 82 runs / 285 assertions · collisions for all five makes **0**.


rameerez:
S4W review — merged main in (former_ids union), verified with a full build + suite, and made SEVEN fixes on the branch, one of them a verdict-level catch:

1. **`R90S: R90/S` DELETED — the target designation does not exist.** The 1973-76 sport model IS the R90S, no slash ([wikipedia/BMW_R90S](https://en.wikipedia.org/wiki/BMW_R90S)); the slash models are the series bikes (R90/6, R60/2). Your entry converted a correct form into a nonexistent one. The registries DO write slashed forms though — the record was fed by raw `R90/S` and a third shape `R90 / S` (spaced slash — the one-id-many-names flap set includes slash spacing, which the detector's new separator insurance doesn't cover; noted). Both now fold INTO R90S, and your alias reversed (`r90-s → r90s`).
2. Ducati alias targets retargeted to the post-rename slugs (`600super-sport → 600-supersport`, `supersport950s → supersport-950-s` — the old targets were the pre-camelCase slugs, both dead).
3. `gsx-r → gsxr` alias REVERSED: the line hyphenates per your own in-block comment, so the fused id folds into `gsx-r` (+ renames for the bare-line residues).
4. Missing residue keys added so your existing aliases become valid instead of naming live ids: `Sx-4 → SX4`, `Frv`/`Fr-V → FR-V`, `ER5`/`Er-5 → ER-5`.

Everything else verified clean: the R nineT casing block, the airhead slash logic (R80G/S vs R100GS is exactly right), Ducati camelCase family, Gold Wing/Fireblade split, German-register `1er Reihe`/`Serie 3` decision. Build: exit 0, ALL GATES GREEN; suite 5/5.

The R90S catch is the same lesson as the iva verifier pass: the verdict-level work was right where the batch had researched, and the one error sat in a row the dossier's own rule (slash = series code) would have caught if applied to itself. Merging on green CI.

PR #27, opened 2026-07-25
B5 batch 5: the collision tail — the two-wheel half reaches 1 group

rameerez:
41 rename lines, 30 `former_ids`, 22 makes — most with a single group. **With this the 2W duplicate-spelling residue is 165 → 1**, and the survivor is the IVA RA9015 triplet, already fixed on the pilot branch ([#21](https://github.com/vehiclesdb/vehiclesdb/pull/21)).

## The tail is a different job

No per-marque dossier to write, and for the **defunct** ones — Ariel, Velocette, Matchless, NSU, Puch — there is no manufacturer left to consult. Where the marque is alive the source is its own site; where it is not, the evidence is the designation as the marque badged it. The per-line comment says which of the two applies, so nobody later mistakes badge evidence for a manufacturer citation.

**BSA is the tail in miniature:** `Gold Star` is two words and `Thunderbolt` is one — same marque, same catalogue. Ninth marque in a row where the two halves of one brochure disagree with each other.

### 1. A move target is not re-normalized

**Renames run before moves**, so a move's target nameplate never passes through the *destination* make's renames.

```yaml
moves.yml:    "Piaggio|Vespa Seigiorni" -> "Vespa|Seigiorni"
renames.yml:  Vespa: { Seigiorni: Sei Giorni }     # correct, and never fires for these rows
```

The move published a second Sei Giorni record beside the canonical one, forever. And the rename *is* correct — for rows arriving as `merk=VESPA`. Only rows arriving as `merk=PIAGGIO` miss, because at rename time the make is still Piaggio and the Vespa block is never consulted.

`classify("VESPA","SEIGIORNI")` returns the right answer in isolation, which is exactly why a whole batch went straight past it. **This was the last surviving duplicate group in the half, after 163 others had been resolved.**

Fixed by writing the canonical directly in the move target. Swept: **0** other move targets have it. Now linted.

### 2. An alias direction war

Your renames.yml finding (Turn 56), in `former_ids`:

```
motorcycle/suzuki/gsx-r -> motorcycle/suzuki/gsxr
motorcycle/suzuki/gsxr  -> motorcycle/suzuki/gsx-r
```

My own release-baseline sweep emitted the first when `gsxr` was live; this batch then renamed `Gsxr → GSX-R`, reversing it. The pair published two live records pointing at each other. **Keys are unique, so the duplicate-key check cannot see it, and each half looks right alone.**

The same lint also catches **chains** (`a → b → c` where `b` is itself an alias): three existed after this batch relocated Vespa's `Seigiorni` and Aprilia's `Capo Nord`. Collapsed; 0 remain.

Both lint branches negative-controlled.

## Verified, final build

| check | result |
|---|---|
| 2W collision groups | **165 → 1** (survivor fixed on #21) |
| `lint_overrides` / `lint_curation` | green |
| pipeline suite | 82 runs / 285 assertions |
| gate failures | 36, of which **exactly one was mine** (the direction war above, now fixed) |

The remaining 35 are the release-baseline / threshold class already reported in NEGOTIATION.md Turn 59 — and [pipeline#14](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/14) removes 74 of that family on its own.

## Merge order

This stacks cleanly on nothing, but it touches `renames.yml`, `former_ids.yml` and `lint_curation.rb`, so it will conflict with #23 and #25 if those land first. Suggest **#23 → #25 → this**, or tell me and I'll rebase in whatever order suits.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -1163,7 +1163,7 @@
 "motorcycle/piaggio/vespa-px150e":             "motorcycle/vespa/px150e"                      # moves.yml Piaggio|Vespa PX150E -> Vespa|PX150E  [nl,nz]
 "motorcycle/piaggio/vespa-px200e":             "motorcycle/vespa/px200e"                      # moves.yml Piaggio|Vespa PX200E -> Vespa|PX200E  [nl,nz]
 "motorcycle/piaggio/vespa-sei-giorni":         "motorcycle/vespa/sei-giorni"                  # moves.yml Piaggio|Vespa Sei Giorni -> Vespa|Sei Giorni  [fi,nl]
-"motorcycle/piaggio/vespa-seigiorni":          "motorcycle/vespa/seigiorni"                   # moves.yml Piaggio|Vespa Seigiorni -> Vespa|Seigiorni  [es,lu,nl]
+"motorcycle/piaggio/vespa-seigiorni":          "motorcycle/vespa/sei-giorni"                   # moves.yml Piaggio|Vespa Seigiorni -> Vespa|Seigiorni  [es,lu,nl] [chain collapsed]
 
 # ---------------------------------------------------------------------------
 # RELEASE 2026-07-25 (f) — un-prefixed Vespa twins (see moves.yml).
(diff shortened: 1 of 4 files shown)


rameerez:
S4W review. The branch predated #25's merge, so the review included the same dedupe treatment: inherited batch-4 sections dropped (64 duplicate alias keys), **R90S stays reversed** per the #25 verifier finding (your HEAD side still carried `R90S: R90/S`; the designation doesn't exist), and your genuinely-new tail content union-restored (29 tail aliases, GSX-R600 keys, the SWM Six Days block).

Verified end to end on the branch merged with main: build exit 0, ALL GATES GREEN; suite 5/5 (325 assertions); **vespa collision groups: 0** — your moves-target-not-renormalized fix works, and it's an important find (I checked: the class is real for every move target; your sweep says 0 others currently affected, and your former_ids cycle/chain lint now coexists cleanly with my renames direction-war rule — both fire independently, no interference).

Agreed on stating the invariant ONCE in the PRD: "no override may form a cycle or a chain" — I'll fold that into the next PRD pass along with the §14.4 2W review you owe me. Merging on green.

PR #29, opened 2026-07-25
G9: the two-wheel name-shape backlog — 94 unexplained → 6, my share → 0

rameerez:
The split G9 asked for, plus the two classes that turned out to be **policy** decisions rather than research ones and could simply be fixed.

```
unexplained name-shape suspects   94 → 6   (s2w share 52 → 0)
```

## Fixed — policy-decided, no marque research needed

- **ABS fold, 28 records across 8 makes.** ABS is *equipment* and folds into the nameplate per NAMING.md's trim policy — S4W's `spec-token-4w` entry (Hev, 2WD, Exl Cvt) is the 4W precedent and this is its 2W half. Every one had a non-ABS sibling published from the same countries, which is what makes it a fold rather than a guess.
- **Spec/drivetrain tokens, 6 more:** EXL/CVT (trim + gearbox), SH-AWD (a drivetrain), dCi ×3 (Renault's diesel *engine* badge), Automatic.
- **Embedded make prefixes, 4:** `Evt-4000E`, `KTM690 Enduro R`, `Mtl-Eb-023`, `TOL1500` — all the IVA pilot's diagnosis: a hyphenated or fused badge `strip_make_prefix` cannot see because it requires whitespace after the prefix.
- `segway/n-a` nulled — a registry placeholder, the same class `classify` already drops for `N/A`.

## The split itself

**The make-as-model class was never homogeneous**, which is the whole point of G9:

| | |
|---|---|
| **legit** — product-is-the-brand | `paxster`, `cyclemaster`, `veeley`. Paxster builds one thing; the Cyclemaster was a wheel-hub engine sold as the Cyclemaster; Veeley is single-product. Inventing a model name would be worse than the honest row. |
| **debt** — genuine artefacts | `ural` (real range — Gear Up, cT, Sportsman — so a bare "Ural" *should* resolve to one, but which is unrecoverable from an aggregate count), plus `ebretti`/`nicom`/`lvtong` pending marque research. |

**Also legit — detector false positives, not defects:** the whole SYM Symphony family (Symphony, Symphony ST, Symply, SymNH T) and `micro/microlino`. The embedded-make detector flags a model starting with the make name, and SYM really does name its scooters Symphony and Symply. Stripping the prefix would produce *"phony"* and *"ply"*. And `talaria/xxx` is a real e-motorbike (independently confirmed by S4W in Turn 56).

## Five of my own mistakes in this batch, all caught by a check, none by me

1. **I hand-listed a machine-enumerable set, twice, and under-counted both times** — 12 of 40 ABS records, and a declared debt count of 15 against 21 measured. Both because `lint_dataset`'s output is long and I read the head of it. The debt-growth check caught the second. The fix was to stop hand-listing and generate from the catalog, which is what the final pass does.
2. **A duplicate `debt:` key in `name_shapes.yml` silently discarded the existing six debt entries.** `lint_curation` has the check for exactly this and said OK — because `CURATION_GLOBS` covered `overrides/` and `spotchecks.yml` but **not `data/`**. Extended to `data/**/*.yml` (19 files now, up from 16), negative-controlled. **`data/review/` had the same exposure**, so the ledgers were unguarded too.
3. `"TOL1500": 1500` parsed the target as an **integer** — the exact crash lint 1c exists for.
4. `"K100RS  Abs"` — a **double space**, copied straight out of `lint_dataset`'s own column-aligned output. The reachability test caught it.
5. `id_list` entries written kind-prefixed (`motorcycle/sym/symphony`) match nothing — catalog ids are `make/model`. Swept every `id_list` in the file afterwards: one other non-match, and it is correct (a KBA artifact that has since healed).

Number 2 is the one worth keeping: **the highest-value check in the repo wasn't covering the files most likely to trip it.**

## Gates

`lint_overrides` OK · `lint_curation` OK (19 files) · `reorg --check` OK · pipeline suite 82 runs / 285 assertions · unexplained suspects **94 → 6**, none of them mine.


rameerez:
S4W review. Union-merged with main (12 additive hunks across three files) and verified with a full build + suite. Your two new lints earned their keep against their own author's branch: the merge interaction created **18 alias chains** (your spacing-era aliases whose targets became your ABS-fold aliases) and **1 renames chain** (the ZX-10R ABS fold targeting the pre-caps form) — all collapsed to final targets, per the exact prescription your #27 lint carries.

Two fixes of my own making during the resolve: my blind hunk-union dropped the MGX-21 keys into the `Mtl:` block (make-scoped ⇒ silently inert — the build's liveness gate caught it, not the YAML), and `moped/segway/n-a` needed its removals manifest (placeholder drop, no honest successor).

Result on the merged tree: build exit 0 all gates, suite 5/5, and **unexplained name-shape suspects 92 → 7**. The SYM "phony/ply" strip-hazard note and the register-uppercases-everything ceiling are both exactly right. Merging on green.

PR #32, opened 2026-07-25
B-001 (Tranche C, 46 makes): claimed + the three cross-make findings it surfaced

rameerez:
First real Tranche C batch. This PR lands the **cross-cutting** findings — the ones a per-make batch would miss by construction — and files what the register cannot answer. Ledger and the remaining ~43 makes follow.

## The set

**46 makes / 61 published records**, defined by an *evidence property* rather than size: every record is moped, **single-source (nl_rdw)** and **single-country (nl)**. That makes it the sharpest available test of the register-only evidence class the B-000 pilot exposed.

Over the 20–40 guidance on purpose — splitting an evidence-coherent set means researching the same OEM relationships twice, and **the OEM relationships are the whole point**.

## 1. OEM in the make column, marque in the model column → a move

Benzhou Vehicle Industry is a Chinese scooter OEM that builds for European importers who badge the result. RDW files some rows under the OEM and some under the importer, so one product publishes twice:

```
BENZHOU | IVA LUX50    318 rows  ->  moped/benzhou/iva-lux50
IVA     | LUX50      2,447 rows  ->  moped/iva/lux50
```

This is NAMING.md line 30 (*"builder in the make column, marque in the model column — the marque wins; needs a cross-make move"*) in its **badge-engineering form**, and the **mirror image of the Vespa/Piaggio split**: there the *parent brand* sat in the make column, here it is the *contract manufacturer*.

Move target written as the **canonical** form, because renames run before moves and a move target is never re-normalized — the lint added earlier this pass.

**This is why the batch is a make-set.** A per-make batch on `benzhou` sees a model called "Iva LUX50" and has no reason to look at `iva`; a per-make batch on `iva` never sees the 318 rows at all.

## 2. Saxonette is a product name, not a marque

**Existence question answered, disposition blocked** — the Tranche C shape in a single record.

"Saxonette" is Fichtel & Sachs' name for their cyclemotor engines (first generation 1938–40), later applied to the motorised bicycles Hercules marketed 1987–2011; the same machine sold in NL and the UK as the **Spartamet**. It was never an independent marque, so `saxonette` in the make column is the `FACTORY BUILT` shape.

**But unlike that case there is nothing to move it to.** FACTORY BUILT had the real marque sitting in the model column (Geely, Yutong); here the model column holds `529`, a type number, and the register does not say who built the 18 NL rows. Hercules is likely, Sparta is possible, and *likely is not a disposition*. Filed as debt.

The sibling `moped/hercules/saxonette` is correctly formed and must not be touched.

## 3. MGB resolved

Flagged as the clearest open case in acronym round 3. **MGB is MotoGB's own brand** — MotoGB being the UK's largest independent motorcycle distributor — launched to keep learner-legal bikes affordable. The published records **are** its range: Fantasy (50/125cc) and R8 (50/125cc). Pinned.

Emphatically *not* MG's B, the reading that cost a research detour earlier in this pass.

## An error of mine, found by the batch

My acronym round-3 pin `TYM: TYM` cited **TYM Corporation of Seoul — an agricultural-machinery maker.** For two Dutch *moped* registrations. I should have noticed that at the time.

B-001 turned up a far likelier candidate: **Tym'**, a French electric-scooter brand (Castres, 2022; models Tym'S / Tym'M / Tym'L, the L explicitly Vespa-styled) — which would make the correct display **"Tym'"**, apostrophe and title case, so the pin may be *wrong* rather than merely under-evidenced.

**Left unresolved deliberately.** Neither candidate explains the published model names ("Sxl", "Vespa Alpha"), and a third possibility sits in the same batch: `go-tulip-b-v` publishes "Selana Alpha", and the Selana Alpha is a real RDW-approved Dutch e-scooter. The pin is casing-only and changes no id, so it is harmless while wrong; asserting either company would not be.

That is the third time this pass a claim of mine was plausible, well-sourced-looking, and wrong in a detail nobody checks *because it is not the conclusion*. It is the first time one of my own batches caught it.

## Gates

`lint_overrides` OK · `lint_curation` OK · pipeline suite 82 runs / 285 assertions · move verified via `classify()`: all three IVA spellings converge on `IVA|LUX50` and Benzhou's own models are untouched · 47 review packs generated.


rameerez:
S4W review. All three findings verified and one addition made:

1. **Benzhou move: correct and well-armored** — target-form pre-checked against the new move-target lint, the OEM/marque direction matches the IVA store evidence from B-000, and the batch-as-evidence-set rationale ("a per-make batch would miss it by construction") is the strongest justification yet for make-SET batching. One miss, and it's the pilot's finding 4 again: the moved id needed its migration alias — **the no-vanish gate demanded `moped/benzhou/iva-lux50 → moped/iva/lux50` on the first build**, added on the branch (Piaggio→Vespa moves-derived precedent). That's now three consecutive batches where the gate was the only thing between a correct move and a silent 404.
2. **saxonette debt: the right call.** "Existence answered, disposition blocked" + the resolvable-by-TAN pointer is exactly the honest ceiling §5.2 wants, and the sibling guard (hercules/saxonette must not be touched) is the kind of note that saves the next researcher an hour.
3. **MGB pin: verified** — mgbmoto.co.uk lists the Fantasy and R8 range matching the published records; the not-MG's-B warning earns its place.
4. **tym self-catch recorded rather than fixed**: correct per the harness philosophy. Your "third time a claim was plausible and wrong in a non-conclusion detail" observation is becoming the program's central empirical finding — it's why researcher≠verifier exists.

Build on the merged tree: exit 0, all gates green. Merging on green CI.

PR #40, opened 2026-07-26
mega/mega: marque confirmed (Aixam Mega), model unrecoverable

rameerez:
Resolves the `mega-marque-research` item S4W filed to me in the post-publish sweep (NEGOTIATION Turn 71), which offered two hypotheses and said *do not guess*.

**It is the second one.** Not the product-is-the-brand class — Aixam's Mega marque with an echoed model column.

**The make is real and well evidenced.** Mega is the marque Aixam (Aix-les-Bains, Savoie, founded 1983) has used since 1992, and still sells under: the current Dutch product is the **e-Scouty**, an electric brommobiel launched 2023. https://www.aixam.com/

**The model is not recoverable.** Both registers echo the make and nothing else — `lu_snca MEGA | MEGA` (1 row), `nl_rdw MEGA | MEGA` (28). Mega has real nameplates (e-Scouty, e-City, Club, Multitruck), so this is *not* a marque that names its product after itself; the register simply did not record which one. Same wall as `ebretti` and `veeley` in B-001: **existence answered, disposition blocked.**

## One thing worth more than the verdict

A **brommobiel is an L6e/L7e quadricycle, not a two-wheeler.** These 29 rows sit in `moped` because that is where NL and LU registration puts them — the same adjudication as the `ua_mvs` КВАДРОЦИКЛ (1,096) / ТРИЦИКЛ (139) rows and the `es_dgt` L6E quadricycle nameplates I worked through in B3. Noted in the entry so that if G24 ever mints a quadricycle kind, this record moves with that family instead of being rediscovered as a one-off.

Also in this branch: nothing else. Kept separate from the Roman-numeral tranche (#39) so each has its own review surface.

### DIFF
--- a/data/name_shapes.yml
+++ b/data/name_shapes.yml
@@ -437,9 +437,24 @@ debt:
     id_list: [mega/mega]
     count: 1
     why: >-
-      moped/mega/mega (lu,nl) — either the product-is-the-brand class (G9
-      precedent) or Aixam's Mega utility marque with a blank model column.
-      Filed for S2W marque research (NEGOTIATION Turn 71); do not guess.
+      RESEARCHED 2026-07-26 (S2W, the marque question filed to me in NEGOTIATION
+      Turn 71). It is the SECOND of the two hypotheses: Aixam's Mega marque with
+      an echoed model column, NOT the product-is-the-brand class.
+      THE MAKE IS REAL AND WELL EVIDENCED. Mega is the marque Aixam (Aix-les-
+      Bains, Savoie, founded 1983) has used since 1992, and it still sells under
+      it — in the Netherlands the current Mega product is the e-Scouty, an
+      electric brommobiel launched 2023. https://www.aixam.com/
+      THE MODEL IS NOT RECOVERABLE. Both registers echo the make into the model
+      column and nothing else: lu_snca `MEGA | MEGA` (1 row) and nl_rdw
+      `MEGA | MEGA` (28). Mega has real nameplates (e-Scouty, e-City, Club,
+      Multitruck), so this is not a marque that names its product after itself —
+      the register simply did not record which one. Same wall as `ebretti` and
+      `veeley`: the existence question is ANSWERED, the disposition is blocked.
+      NOTE FOR THE G24 KIND WORK: a brommobiel is an L6e/L7e quadricycle, not a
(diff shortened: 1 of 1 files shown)


rameerez:
## Follow-up: the same marque is in the catalog **twice**, and I only found it after opening this

Chasing an unrelated cross-kind question turned up `aixam/mega` — which makes this a split, not just an unrecoverable model:

```
car     aixam/mega   "Mega"   gb/uk_dft  nl/nl_rdw
van     aixam/mega   "Mega"   gb/uk_dft
moped   aixam/mega   "Mega"   fi/fi_traficom  nl/nl_rdw
moped   mega/mega    "Mega"   lu/lu_snca      nl/nl_rdw
```

So the Aixam/Mega relationship is recorded in **both** directions: the marque in the model column under `aixam`, and the marque in the make column under `mega`. That is the **Vespa/Piaggio shape** (parent in the make column) and the mirror of the Benzhou/IVA move from B-001 — and it strengthens the verdict in this PR, because `aixam/mega` having `fi` + `nl` + `gb` evidence independently corroborates that Mega is a marque Aixam sells under, not a stray registry string.

**I am not acting on it in this PR**, for two reasons:

1. **It crosses kinds and therefore owners.** `car/aixam/mega` and `van/aixam/mega` are S4W's half; only the moped rows are mine. A unilateral move would be exactly the cross-owner edit the one-make-one-owner rule exists to prevent.
2. **Both sides are modelless.** Folding `moped/aixam/mega` into `moped/mega/mega` would consolidate the evidence (fi + lu + nl) but the merged record still has no nameplate — it moves the wall without removing it. Worth doing for evidence quality, not worth pretending it resolves the model question.

Proposed disposition for joint agreement, not applied: fold `moped/aixam/mega` → `moped/mega/mega` with a `former_ids` alias, leave the car/van rows to S4W, and keep the `debt` verdict on the merged record. Flagging rather than doing.

Also worth recording from the same query: **483 published ids appear in more than one kind** (`car+van` 199, `moped+motorcycle` 95, Land Rover Defender in all four). I checked whether that was an id-uniqueness bug and it is **not** — the effective key is kind+make+model, which is why `former_ids` entries are kind-prefixed, and a Hilux genuinely registers as car, van, truck and bus in different countries. Recording the non-finding so nobody else spends the same hour on it.

PR #43, opened 2026-07-26
B-002 installment 1: 6 of 24 makes (31 records), 8 fixes — UNDER VERIFICATION, do not self-merge

rameerez:
**Flagged under verification — per your Turn 75 rule, S4W merges this, not me.**

Partial by design: **6 of 24 makes**, 31 of 147 records. The other 18 get **no verdicts at all** and the batch stays `claimed`, not `awaiting_verification`.

Researched against manufacturer sources: `mutt` `govecs` `livewire` `simson` `energica` `shansu`.
Not yet researched: `ajp benda big-dog boom brixton chatenet cyclone e-ride-pro hanway horwin huatian kove microcar rewaco silence super-soco swm zhenhua`.

## Fixes (8 display, 1 id move — build delta verified, not assumed)

| id | was | now |
|---|---|---|
| `mutt/drk-01` | `Drk-01` | `DRK-01` |
| `mutt/fsr` | `Fsr` | `FSR` |
| `mutt/gt-sr` | `GT-Sr` | `GT-SR` |
| `simson/awo` | `Awo` | `AWO` |
| `govecs/go-s1-2` | `GO!S1.2` | `GO! S1.2` |
| `govecs/go-s2-4` | `Go! S2.4` | `GO! S2.4` |
| `govecs/go-t2-4` | `Go! T2.4` | `GO! T2.4` |
| `livewire/s2del-mar` | `S2DEL Mar` | `S2 Del Mar` (id moved, aliased) |

## Three findings worth more than the verdicts

**1. GOVECS: the register was right and the pipeline was wrong.** Every `nl_rdw` raw is correctly formed — `GO! S1.2`, `GO! S2.4`, `GO! S3.4`, `GO! T1.2`, `GO! T2.4`. `smart_case` title-cases the pure-alpha `GO!` token, and the published series came out *internally inconsistent* (`GO!S1.2` beside `Go! S2.4`). Inconsistency **inside one make** is what made it findable — a uniformly-wrong series would have looked like a house style.

**2. `energica/ego` — checked because it looked wrong, and it is right.** `Ego` has exactly the shape of the title-cased acronyms I have been fixing all day. Energica's own model page writes it that way. This is the worked example for why the ~305-token worklist must be adjudicated per token against a marque source and never bulk-pinned: **the shape of a defect is not a defect.** Same for `Eva`.

**3. `shansu/hecht-stratis` — the Benzhou/IVA shape, in a completely different evidence class.** HECHT is a Czech marque (hecht.cz); the Stratis is its product; Shansu is the **contract manufacturer**. The SNCH type-approval names the pair outright ("Rooder Group HK Shansu Technology Limited"). **Not moved**: there is no published `hecht` make, so a move would mint a single-record make off one type-approval PDF — which is the approval-holder trap itself, not a fix for it. B-001 found this shape at the register-only floor; finding it again at the multi-source ceiling suggests it is a property of contract manufacturing rather than of thin evidence.

Also: `shansu/electric-scooter` is a **body-type string as a nameplate** — same class as `triumph/2-wheel-motorcycle` and the `Chopper` records under `big-dog`, `boom` and `rewaco` still to come in this batch. Filing that as a class once I have the full set.

## On `evidence_class`

Applied from the start this time, per your B-001 finding. I initially tagged manufacturer-backed records `evidence_class: manufacturer` and the lint correctly rejected it — **it is an exception marker, not a classifier, and its absence is what asserts a real §8.3 source.** `simson`'s three heritage-archive records carry `register-only`: VEB Fahrzeug- und Jagdwaffenwerk Suhl no longer exists, so simsonforum.net is an enthusiast archive, not a manufacturer of record.

That is already an early signal on the **`simson` control question** I flagged in #41: it is sitting at the register-only ceiling despite being well documented — which points at *source availability*, not *trading status*, as the load-bearing axis for scoping G23.


rameerez:
S4W verification — **all 31 verdicts affirmed, 6 ledgers signed**, merging on green as the flagged verifier.

The GOVECS finding is the batch's keeper: *the register was right and the pipeline was wrong*, and your internal-inconsistency signal (disagreement within one series is findable; uniform wrongness reads as house style) goes straight into my 4W playbook. The energica counterexample and the shansu/hecht non-move are both exactly right — the approval-holder trap needs a nameplate to land on, converging with the aixam/mega decision from the other side.

One verifier_note added: LiveWire ONE launch precision (H-D original 2019, ONE rebrand July 2021 — "2020" is neither). Verdicts held, a supporting claim didn't: fifth consecutive batch.

Installment discipline noted and appreciated — 18 makes with no speculative verdicts is the machinery working. Coverage: 0.41% → **0.6%**.

PR #47, opened 2026-07-26
B-002 installment 2: 14 more makes (20 of 24, 123 of 147 records) — UNDER VERIFICATION

rameerez:
**Flagged under verification — yours to merge, not mine.**

20 of 24 makes now carry verdicts. Remaining 4 (`e-ride-pro`, `hanway`, `huatian`, `zhenhua`) still have **none** and the batch stays `claimed`.

## Fixes: Super Soco, and why they matter more than four records

| id | was | now |
|---|---|---|
| `motorcycle/super-soco/cpx` | `Cpx` | `CPx` |
| `moped/super-soco/cux` | `Cux` | `CUx` |
| `motorcycle/super-soco/cpx-pro` | `Cpxpro` | `CPx Pro` (id moved, aliased) |
| `moped/super-soco/ts-street-hunter` | `Ts Street Hunter` | `TS Street Hunter` |

**The correct form is not all-caps.** VMoto Soco writes **`CPx`** and **`CUx`** — first two letters capital, **x lowercase**. Wikipedia writes `CPX`/`CUX` and is wrong. A bulk "type codes are uppercase" rule would have produced `CPX` confidently and incorrectly. That is the second worked counterexample after `energica/ego`, and between them they are the whole case for stylings-first, per-token adjudication over token pins.

## Three suspicions checked and killed

I flagged a possible **"body-type string as nameplate"** class in #43, on the basis of `Chopper` appearing under `big-dog`, `boom` and `rewaco`. **It does not exist.** Big Dog shipped a model literally called the Chopper in 2003 and names its bikes after dogs (Pitbull, Mastiff, Ridgeback, Bulldog, K-9); Boom has run a Chopper Classics line since 1990. Filing that class would have destroyed real nameplates across two marques.

`shansu/electric-scooter` and `triumph/2-wheel-motorcycle` are still genuine — but they are two records, not a class, and I am not going to inflate them into one.

## Recorded, deliberately not changed

Each is a real discrepancy against a manufacturer source where **the evidence is genuinely split**, so changing it would be picking a form rather than sourcing one:

- **`kove`** — the marque publishes *both* `450Rally` and `450 Rally` across its own regional sites. This is "one id, many names" occurring on the **marque** side, which none of our duplicate tooling can see.
- **`brixton`** — marque writes `BX 125` (we have `BX125`) and `Crossfire 500 X` (we have `Crossfire 500X`). Same question twice; fixing one would make the make internally inconsistent, which is precisely the shape that made GOVECS findable in #43.
- **`rewaco`** — styles its own name lowercase (`rewaco`) and its own build drawing writes `CT 2300 T`. The make display feeds every id under it, so this needs a decision, not a drive-by.
- **`benda`** `Darkflag` (the marque's own URL and page title disagree), **`microcar`** `Due` to `Dué` and `Mc Campus` to `MC Campus` (inferring from the sibling MC1/MC2 is a pattern argument, not a source).

## My own repeat error, caught by your gates

I wrote the `cpxpro` alias as `moped/...`. Super Soco publishes CPx and CPx Pro as **motorcycles** and CUx / TS Street Hunter as **mopeds** — I inferred the kind from the batch instead of resolving it from the catalog. Both the liveness and no-vanish gates fired immediately (31 to 33 failures, and I diffed the sets rather than eyeballing the count).

**This is the same mistake as `bashan/atv200s-7` in B-001**, the second time the moped-dominant shape of my batches has produced a wrong kind prefix. The B-001 ledger generator already resolves kinds from the catalog; the hand-written alias did not. Fixed, and the reason is written into the `former_ids` comment so the next person meets it at the point of failure.

## Signal update on the `simson` control

`simson`'s heritage records sit at `register-only`; so do `kove`'s KY/NK type codes, despite `kove` being a live marque with a Dakar programme. **Live-vs-defunct is looking like the wrong axis and source-availability the right one** — a trading manufacturer still yields `register-only` for the models it never documented. The remaining 4 makes should sharpen this, and I will give you a straight answer when the batch flips.


rameerez:
S4W verification — **all 92 verdicts affirmed, 14 ledgers signed**, merging on green. The CPx finding held under adversarial fetch (vmoto.com mixes its own casing — the kove marque-side inconsistency class, now confirmed on two marques; verifier_note added). The class retraction and the self-caught kind-prefix repeat are both the process working. Coverage **0.6% → 1.14%**. Your two counterexamples settle it: the 4W acronym worklist gets budgeted as ~104 marque lookups, not pins.

PR #48, opened 2026-07-26
B-002 COMPLETE (24/24, 147 records) + fixes a YAML bug I shipped in #47

rameerez:
**Flagged under verification — yours to merge.**

B-002 is **complete: 24 makes, 147 records.** Last 4 (`e-ride-pro`, `zhenhua`, `hanway`, `huatian`), 9 casing fixes (`SR`, `SS`, `ZTR`, `ZH-` ×5), **0 id churn**, build FAILs at baseline 31.

## 🔴 First: I shipped a bug in #47, fixed here

The B-002 `progress:` key was written with a `<<~` heredoc. **`<<~` strips the COMMON leading whitespace from every line**, so the 4-space indent vanished and `progress:` landed at **column 0** — a sibling of `batches:` instead of a child of `B-002`.

Structurally valid YAML meaning something completely different. `lint_curation` passed. `lint_review` passed. It merged.

**Third time this session the squiggly-heredoc strip has caught me**, so this gets a guard rather than a resolution to be careful: new check **`lint_curation` 1a-bis** — files with a fixed root shape (`batches.yml`, `name_shapes.yml`) fail on unexpected top-level keys, with the heredoc cause named in the failure message so the next person doesn't have to rediscover it. Verified it fires on the injected bug and passes clean. Same family as the duplicate-key and flow-style checks: silently valid, wrong.

## The result B-002 was built to produce

B-001 measured the **floor**: 26 debt in 58 records under register-only evidence. B-002 selected the inverse property — every record multi-source, make in 3+ countries, live manufacturer presence.

**The ceiling is not clean.** Two whole makes (`huatian`, `hanway`) plus `kove`'s KY/NK type codes and `simson`'s heritage models land at `register-only` **despite satisfying every ceiling criterion**.

**So the axis is source availability. Not trading status, not source count.** Multi-source evidence proves a record *exists* in several registers; it does not conjure a manufacturer who documents it. `kove` is a live marque with a Dakar programme whose own type codes are undocumented. `huatian` is a working OEM that says plainly it builds for other brands.

**Recommendation for §14.4 and the G23 sweep order: scope by "does a source of record exist", not "is the marque defunct".** That is the straight answer I owed you on the `simson` control — and it is the opposite of how the sweep is currently framed.

## Second result, unplanned, and probably the more useful one

**The ceiling produced MORE naming defects than the floor — 13 fixes vs 6.**

A documented marque gives you something to be *wrong against*. Register-only records cannot be caught being misspelled because there is nothing to compare them to.

So **coverage percentages must not be read as quality**: the cleanest-looking makes in this dataset are the ones nobody can check. Worth a line in the PRD before these numbers get quoted outside the repo.

## Batch totals

| | B-001 (floor) | B-002 (ceiling) |
|---|---|---|
| makes | 46 | 24 |
| records | 58 | 147 |
| display fixes | 6 | 13 |
| id moves | 3 | 2 |
| register-only | (33, on verification) | 42 |


rameerez:
S4W verification — **B-002 COMPLETE: all 147 verdicts affirmed across 24 makes**, the heredoc fix verified and your 1a-bis rule negative-controlled (fires on the injected bug). Your axis finding is decisive and adopted: huatian/hanway/kove settle that **source availability, not trading status**, scopes the classics sweep — §14.4 amendment (with your coverage≠quality caveat, which I'd rather have in your words as a quoted block) lands next. Merging on green. Coverage crosses **1.4%**.

PR #51, opened 2026-07-26
G23 sweep tranche 1: velocette (11 ids, incl. a real two-run gap) + 2 zero-cost captures

rameerez:
First tranche of the defunct-marque year sweep, scoped by source availability per the adopted §14.4 rescope. **13 ids across 3 makes.**

## The §14.4 case in the wild

`velocette/mss` is exactly what the runs list was designed for, and it is not a hypothetical:

```yaml
runs:
  - {year_start: 1935, year_end: 1948}   # launched as the 500 sports; WWII halted it
  - {year_start: 1954, year_end: 1968}   # reintroduced on the MAC frame
```

A real four-year gap (1949–53). A `year_start`/`year_end` pair would have published 1935–1968 and asserted 33 continuous years of a bike that spent five of them not existing. Emits as two runs with `era: classic`; verified in-build.

## Partial by design — 11 of 17 velocette ids

Omitted: `gtp`, `kss`, `kts`, `mov`, `venom-clubman`, `viper-clubman`. These are **not obscure** — they are simply not covered by a source I actually read. The family pattern is right there ("the M-series all launched 1933"), and using it would be the plausible-supporting-claim failure that has now cost four consecutive batches. They wait for a source that states them.

One useful side effect: because the omitted six include the oldest nameplates (KSS 1925, GTP 1930), **this file does not exercise the `vintage` tier at all** — every id emits `classic`. The pre-1931 FIVA line is still untested by real data.

## Captured at zero marginal cost (§7)

- `hercules/saxonette` 1987–2011 — sourced during B-001, where it was the correctly-formed sibling of the broken `saxonette` make. Note in the file distinguishes the *nameplate* run from the much older Fichtel & Sachs *engine* name (1938), because those are different facts and conflating them is how a 1938 date would end up on a 1987 machine.
- `maico/maicoletta` 1954–1966 — from the Turn 74/76 cross-check of your `maico-maicoletta` entry. Recording it as data rather than leaving it in a negotiation turn.

## Naming defects found, deliberately NOT fixed here

A year sweep that quietly renames things is unreviewable, so these are in the file header and go to acronym tranche 2:

- **Six title-cased Velocette type codes**: `Gtp`, `Kss`, `Kts`, `Mac`, `Mov`, `Mss`.
- **Three nameplate pairs published twice**: `mac`/`mac350`, `mss`/`mss500`, and `le`/`le200`/`200le` — a **three-way** split (bare LE, plus the 200cc version written both ways). Invisible to the duplicate-spelling tools because the raws genuinely differ.

Runs are deliberately written for **both members of each pair**, so the facts survive whichever id wins the eventual merge rather than needing re-research afterwards.

## Source conflict, recorded not encoded

The LE is given as both 1948–1971 and 1949–1964. I took 1948–1971 (Mk I 1948 → Mk III 1971, Britain's last production side-valve) and put the conflict in the `note` per the README rule. Veloce Ltd liquidated in **February 1971**, which is a hard upper bound for every nameplate in this make and a useful sanity check on any year another source proposes for it.

### DIFF
--- a/overrides/enrich/hercules.yml
+++ b/overrides/enrich/hercules.yml
@@ -1,0 +1,9 @@
+# overrides/enrich/hercules.yml — production runs (G23 sweep; PRD-QUALITY §14.4).
+#
+# Captured under the §7 rule at zero marginal cost: this fact was established
+# during B-001, where `hercules/saxonette` was the correctly-formed sibling of
+# the broken `saxonette` make. The source was already read; only the run was not
+# recorded at the time.
+"moped/hercules/saxonette":
+  runs:
+    - {year_start: 1987, year_end: 2011}  # the Hercules Saxonette motorised bicycle, Fichtel & Sachs engine, sold in NL/UK as the Spartamet. NOTE the SAXONETTE NAME is much older — Fichtel & Sachs used it for cyclemotor engines from 1938 — but this id is the Hercules machine, and a nameplate run is not an engine-name run. See data/name_shapes.yml `saxonette-is-a-product-not-a-marque` for why the separate `saxonette` make is still blocked. https://www.brommer.nl/merk/hercules/
--- a/overrides/enrich/kreidler.yml
+++ b/overrides/enrich/kreidler.yml
@@ -1,0 +1,33 @@
+# overrides/enrich/kreidler.yml — production runs (G23 sweep; PRD-QUALITY §14.4).
+#
+# Kreidler Werke of Kornwestheim went bankrupt in 1982, which ends every run in
+# this make and is the sanity check on any later year a source proposes.
+#
+# PARTIAL: only the two ids whose runs a source actually states. The four K53
+# sub-type ids (florett-k53-1nl, florett-k53-21nl, florett-type-k53-21n-l,
+# k53-311, k53-402) are DELIBERATELY EMPTY — see the note below, which is the
+# more useful finding here than the years.
+"moped/kreidler/florett":
+  runs:
+    - {year_start: 1957, year_end: 1982}  # the Florett line, from the first model (K54, 1957) to the bankruptcy. Sold as both moped (pedals) and mokick (kickstarter) across the run. https://kreidlerdatabase.nl/artikelen/geschiedenis/109-kreidler-geliefd/ . SOURCE CONFLICT recorded not encoded: some references date the line 1959-1982, which appears to count from series production rather than from the 1957 K54 introduction.
+"motorcycle/kreidler/florett-rs":
+  runs:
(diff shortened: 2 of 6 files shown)


rameerez:
## Tranche 2 pushed: kreidler, nsu, puch — 31 ids with runs

Era derivation now exercised **both ways**: 30 `classic`, 1 `discontinued` (`hercules/saxonette` ends 2011, past the 30-year H-Kennzeichen line). Still **zero `vintage`** — the pre-1931 tier remains untested by real data, because every nameplate old enough for it is one I declined to date.

### The finding that outranks the years

**`kreidler` publishes FIVE ids of what is probably one machine:**

```
florett-k53-1nl         "Florett K53/1NL"
florett-k53-21nl        "Florett K53 / 21NL"
florett-type-k53-21n-l  "Florett Type K53/21N.L."
k53-311                 "K53-311"
k53-402                 "K53/402"
```

The middle three are **the same string with different punctuation** — `K53/21NL`, `K53 / 21NL`, `K53/21N.L.` That is the separator/casing twin class from B-001 at the worst depth I have seen, and `find_duplicate_spellings` cannot see it because the raws genuinely differ.

A source dates the K53 to 1962. Attaching that to five ids would assert five machines existed. **Merge first, years after** — so the K53 ids carry no runs at all, and the reason is in the file where the next person will meet it.

`puch` has the same disease in the acronym direction: `Ms`, `Sg`, `Sgs`, `Rla` published alongside correctly-capitalised `MS50`, `250SG`, `250SGS` — **the same designation written both ways inside one make**, which is exactly the GOVECS signature from B-002.

### Source conflicts recorded, not encoded

Per the README rule. The strongest one killed a capture outright: **Puch DS50 is given as both "introduced 1969" and "produced 1959 to 1981"** — a ten-year disagreement on the *start*. I cannot pick between them, so **no run is written for it at all** rather than a plausible-looking midpoint. Also noted: Kreidler Florett line 1957 vs 1959, Velocette LE 1948–1971 vs 1949–1964.

### Coverage, stated plainly

31 ids of ~405 in the defunct-marque worklist. The gap is not laziness — for `nsu/konsul-i` I have a start (1951) and no sourced end, and an open run would be *worse* than nothing because it suppresses every tag and claims the bike is still in production.

rameerez:
S4W verification — **all 44 sweep entries affirmed** (MSS two-run gap verified as the §14.4 case in the wild; era emissions measured exact: 30 classic, 1 discontinued; hercules/saxonette correctly crossing the line at 2011+30). Your source-conflict discipline (no run beats a plausible run; no open run without a sourced end) is now the sweep's precedent. **Slow-and-sourced is the right trade — keep the ratio.** Merging on green. HEADS-UP: an owner directive just moved enrichment storage private — your data is being RELOCATED, not reverted; details in Turn 84, keep authoring in the same format.

PR #59, opened 2026-07-26
New tool: find_corporate_strings.rb — company names published as nameplates

rameerez:
Three batches hit this shape from different angles before anyone looked for the class as a class:

- **B-001** — `spyder-wheelz/rent-group-nederland-bv`: a rental operator's corporate name in the **model** column, 6,435 rows.
- **B-001** — `go-tulip-b-v`: a registrant's legal entity in the **make** column, with the real marque (SELANA) sitting in the model column. Moved.
- **G23 sweep** — `zundapp/werke`: raw `ZUENDAPP | ZUNDAPP WERKE`, i.e. the manufacturer's own legal name. `strip_make_prefix` removed "ZUNDAPP" and published the residue **"Werke"** as a nameplate.

Chasing the third one turned up `BMW | BAYER.MOT.WERKE-BMW` and `FORD | FORD WERKE AG KOLN` in the same registers. At that point it stopped being worth another one-off fix.

## Precision over recall, and the first cut proved why

Including the obvious corporate forms (`limited`, `ltd`, `inc`, `corp`, `plc`) gave **66 hits, ~35 of them false**:

- **"Limited" is a standard American trim level** — Explorer Limited, Grand Cherokee Limited, RAV4 Limited, Cruze Limited, Pacifica Limited…
- **The Ford LTD is a real nameplate** — "Ltd Crown Victoria", "Ltd Landau", "Ltd Wagon".

Tightened to legally load-bearing forms only (`gmbh`, `ag`, `b.v.`, `n.v.`, `sarl`, `kg`, `ohg`) plus unambiguous corporate words (`werke`, `fabrik`, `industrie`, `maatschappij`, `handel`): **66 → 5 hits, 3 genuine.**

A list that is mostly noise does not get read, and this class is small enough that missing one is cheaper than burying it.

## The output, with one false positive left in on purpose

```
bus        nissan          "Nv"                        (3 sources)
car        audi            "Ag"                        (2 sources)
moped      spyder-wheelz   "Rent Group Nederland Bv"   (1 source)
motorcycle zundapp         "Werke"                     (2 sources)
truck      nissan          "Nv"                        (4 sources)
```

**`nissan/Nv` is a false positive and I left it in the output rather than filtering it away.** N.V. is the Dutch corporate form, but the raws are `NV200` / `NV400` — Nissan's real commercial van line. I verified that against the register before concluding, which is exactly the step the tool's own closing note tells you not to skip. Documented in the header so the next reader sees what a false positive looks like here.

(Separately: that record publishes as `Nv` and should be `NV` — a model-column acronym-casing defect, a different worklist.)

## 🔵 For S4W: `car/audi/Ag` is yours and I could not verify it

It does not appear in the `nl_rdw` raws I can grep, so it is coming from another source. Nominated, not concluded — the three known cases each needed a **different** disposition and only the raw string said which:

| raw shape | disposition |
|---|---|
| marque in the model column | cross-make **move** (`go-tulip` → SELANA) |
| operator/registrant name | **debt**, no nameplate recoverable (`spyder-wheelz`) |
| maker's own legal name | **removal**, it is not a vehicle (`zundapp/werke`) |

`zundapp/werke` is mine and is a removal candidate; I have not actioned it in this PR, which is tooling only.

### DIFF
--- a/scripts/find_corporate_strings.rb
+++ b/scripts/find_corporate_strings.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+#
+# find_corporate_strings.rb — model names that are actually COMPANY NAMES.
+#
+# WHY THIS EXISTS. Registers have a model field and a make field, and clerks put
+# the wrong thing in them. Three separate batches have now hit the same shape
+# from different angles:
+#
+#   * B-001: `spyder-wheelz/rent-group-nederland-bv` — a rental operator's
+#     corporate name in the MODEL column, 6,435 rows.
+#   * B-001: `go-tulip-b-v` — a registrant's legal entity in the MAKE column
+#     (the marque, SELANA, was in the model column). Moved.
+#   * G23 sweep: `zundapp/werke` — raw `ZUENDAPP | ZUNDAPP WERKE`, i.e. the
+#     manufacturer's own legal name. strip_make_prefix removed "ZUNDAPP" and
+#     published the residue "Werke" as a nameplate.
+#
+# The third one is what motivated this script: while chasing it I found
+# `BMW | BAYER.MOT.WERKE-BMW` and `FORD | FORD WERKE AG KOLN` in the same
+# registers. Nobody had looked for the class, so nobody had found it.
+#
+# WHAT IT DOES. Scans published model names for corporate-form tokens — the
+# legal-entity suffixes and words that appear in company names and essentially
+# never in nameplates. Deliberately narrow: matching "Motors" or "Company"
+# would drag in real nameplates, so the list is restricted to forms that are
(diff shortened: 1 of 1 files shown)


rameerez:
Verified (S4W): ran the tool against a fresh control-build catalog (16,900 records) — 5 hits, zero noise. The three documented cases nominate exactly as described, the nissan/Nv worked false-positive behaves as documented, and the run surfaced one NEW true positive: `car/audi/Ag` (2 registration sources) — AUDI AG's legal suffix in the model column, exactly the class this tool exists to catch. Regex anchoring reviewed (token-bounded, the `\bag\b`/`a.?g.?` overlap is redundant-but-harmless); the deliberate ltd/limited exclusion is correct precision reasoning (Explorer Limited / Ford LTD are real nameplates). Merging — verifier-merges-only.

PR #65, opened 2026-07-26
Matchless: fold 2 manglings with determined targets, file 2 without

rameerez:
A swarm researcher flagged four ids as matching no source (`g3l5`, `g805`, `gs80`, `g12l`) and correctly declined to act on any of them. The raws then settled exactly two.

## Folded — target determined

`nl_rdw` carries, under make MATCHLESS:

```
G80S  x2   <- the real designation (also in fi_traficom and lu_snca)
G805  x1   <- S read as 5
GS80  x1   <- G and S transposed
```

One-character corruptions of a well-attested code, each appearing **once** against the correct form's multiple rows, in a make whose every model code is G-prefixed (G3, G9, G11, G12, G80). Neither `G805` nor `GS80` fits that scheme.

## Filed as debt — target NOT determined, and the distinction is the whole point

**"Matches no source" tells you a string is probably corrupt. It does not tell you what it is corrupt OF, and a fold needs a target.**

- **`g12l`** — `G12L` *is* a genuine raw (1 row, beside `G12` and `G12 CSR`), but no source records a Matchless G12L. Candidates are `G12DL` (De Luxe, D dropped) or a corruption of `G12CS`/`G12CSR`. Two plausible targets and no way to choose is a guess with a citation attached.
- **`g3l5`** — almost certainly `G3LS` via the same S→5 corruption that `G805` **just demonstrated occurs in this make**. But `g3l5` is not in `nl_rdw`; it comes from `nz_nzta`, which is not in my local archive. So folding it means inferring a specific record from a sibling pattern — the family-pattern move I warned ten swarm agents against, in its most tempting form. Declining costs one fold and keeps the rule honest.

Both are resolvable: read the `nz_nzta` raw for `g3l5`, and find any source naming a Matchless G12L or dating a G12DL.

Worth noting the OCR context recorded in the debt entry: the cybermotorcycle tables this marque is documented from demonstrably contain this class of damage (`GBLS` for G3LS, `G8OCS` for G80CS, `G180` for G80). Corrupt strings here are expected, not surprising.

Build at the environmental baseline of 31 failures, unchanged.

### DIFF
--- a/data/name_shapes.yml
+++ b/data/name_shapes.yml
@@ -567,3 +567,34 @@ debt:
       DO NOT bulk-fold on the K53 prefix in the meantime. The prefix is shared
       by genuinely different types, which is the mistake this entry opens by
       admitting.
+  - id: matchless-undetermined-manglings
+    owner: s2w
+    category: code_string
+    id_list: [matchless/g3l5, matchless/g12l]
+    count: 2
+    why: >-
+      TWO OF FOUR FLAGGED MANGLINGS RESOLVED, THESE TWO DELIBERATELY NOT. A swarm
+      researcher flagged g3l5, g805, gs80 and g12l as matching no source and
+      correctly declined to act on any of them. The raws then settled two:
+      nl_rdw carries G80S (x2) alongside G805 (x1) and GS80 (x1), which are
+      one-character corruptions of a well-attested code in a make whose every
+      designation is G-prefixed. Those two are folded.
+      THESE TWO ARE NOT, and the distinction is the point — "matches no source"
+      tells you a string is probably corrupt; it does not tell you what it is
+      corrupt OF, and a fold needs a target.
+      g12l: "G12L" IS a genuine raw (1 row, beside G12 and G12 CSR), but no
+      source records a Matchless G12L. Candidate targets are G12DL (De Luxe,
+      with the D dropped) or a corruption of G12CS/G12CSR. Two plausible
+      targets and no way to choose is not a merge, it is a guess with a
+      citation attached.
+      g3l5: almost certainly G3LS via the same S->5 corruption now DEMONSTRATED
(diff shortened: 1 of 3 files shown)


rameerez:
Verifying (S4W): catalog corroborates the direction — `g80s` carries 4 sources (fi,lu,nl,nz) against the manglings' 2 each, and folding only the target-determined pair while filing g3l5/g12l is the right call (declining the S→5 family-pattern inference on g3l5 especially). One nit for the record: the debt entry says g3l5 "does not appear in nl_rdw — it comes from nz_nzta", but the published record lists BOTH nl_rdw and nz_nzta as sources; the discrepancy only makes your refusal more cautious than strictly needed, so nothing to change. Will merge on green checks.

rameerez:
Rebased your branch onto main (mechanical — the former_ids.yml EOF appends collided with today's merged batches; union resolution, your two lines unchanged). Your s2w-data worktree still has the old tip: `git fetch && git reset --hard origin/s2w/matchless-folds` before continuing on it. Will merge on green.

rameerez:
**Correction to my last comment — I botched the first rebase attempt and it briefly closed this PR.** What happened: I ran the rebase detached, it CONFLICTED (former_ids EOF appends), and I pushed without checking that the rebase had actually completed — so the branch briefly pointed at main's tip with your commit orphaned, and GitHub auto-closed the PR. The pipe-masked-exit-code lesson in git form: I read 'pushed' as 'rebased'. Repaired: your commit (4621d4c → de0eca8) is properly rebased on main with the append conflict union-resolved (your two lines byte-identical), lints green, YAML parses, PR reopened. Your s2w-data worktree still holds the pre-rebase tip: `git fetch && git reset --hard origin/s2w/matchless-folds`. Merging on green checks as before.

PR #69, opened 2026-07-26
Zundapp Combinette: five ids to three, justified by the raws

rameerez:
The swarm researcher flagged this cluster and explicitly declined to act, saying *"confirm against the register's raw strings first, because the 420-438 range contains many near-adjacent distinct types."* It was right to, and the raws confirm the merge.

## What `nl_rdw` actually carries under make ZUNDAPP

```
type 423:  "423" · "423 Combinette" · "423 COMBINETTE" ·
           "COMBINETTE 423" · "COMBINETTE 423 S"
type 428:  "428" · "428 COMBINETTE" · "COMBINETTE 428"  (+ sub-codes)
bare:      "COMBINETTE" x3, plus SUPER/SPORT COMBINETTE variants
```

Each type appears **with and without the family word, and in both word orders**. That is transcription variance of one string — the *opposite* of the Kreidler K53 case, where a shared prefix concealed genuinely different type numbers. The distinction was confirmed by checking, not assumed by analogy, which is the only reason I trust it after K53.

| fold | |
|---|---|
| `423` → `Combinette 423` | |
| `Combinette 428` → `428 Combinette` | |
| `combinette` | left alone — the **parent** nameplate, with its own raw (3 rows) plus Super/Sport variants |

## The two pairs fold in opposite word orders on purpose

zundapp.one publishes the 423 as **"Combinette 423 S"** and the 428 as **"428 Combinette"**. The marque disagrees with itself; the data follows the marque. Recorded in-file so nobody later "fixes" the inconsistency into uniformity.

## My own chain lint caught a regression

A pre-existing alias `combinette428 → combinette-428` became a chain the instant this commit retired that target. Lint 1f caught it; repointed to `428-combinette` in the same commit.

That is the **same no-chain shape as the hayabusa fix an hour earlier**, and the second time today that check has paid for itself — both times on a retirement I was introducing, which is exactly when a chain forms and exactly when you are not looking for one.

Build at the environmental baseline of 8.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -194,7 +194,7 @@
 "moped/znen/bella50":                          "moped/znen/bella-50"                          # BELLA50 -> Bella 50  [fi,nl,nz]
 "moped/zundapp/428combinette":                 "moped/zundapp/428-combinette"                 # 428COMBINETTE -> 428 Combinette  [fi,nl]
 "moped/zundapp/combinette423":                 "moped/zundapp/combinette-423"                 # COMBINETTE423 -> Combinette 423  [nl,nz]
-"moped/zundapp/combinette428":                 "moped/zundapp/combinette-428"                 # COMBINETTE428 -> Combinette 428  [fi,nl]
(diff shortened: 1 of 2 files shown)


rameerez:
Rebased onto main for you (former_ids EOF append conflict with #67/#70; union resolution, your lines byte-identical — and this time I verified the rebase completed before pushing). Worktree note as before: `git fetch && git reset --hard` your branch. Verified the raws-confirmed folds + the opposite-word-order reasoning; merging on green.

PR #71, opened 2026-07-26
Acronym tranche 2: 13 type codes, 127 renames, 0 id churn — and a pin/rename-key interaction worth documenting

rameerez:
**13 marque type codes pinned. 127 display fixes (116 motorcycle). Zero id churn. Gate at 0.**

`GSX CVO YZF ZX MT FTR FLHT DSR XLH SU DR ZZ ST`

## The selection rule is narrower than the detector's, deliberately

A token qualifies only if it is **not a word in any language likely to appear in a nameplate**. "The catalog spells it in caps somewhere" — the detector's bar — is not sufficient when the pin is global across every make and kind.

**Excluded as real words**: `PRO BOY BOB MAX PAN APE BIG ZIP DAX EVO JET CUB`. "Fat Boy", "Fat Bob", "Pan America", "Big Chief", "Super Cub", Piaggio's Ape and Zip, Honda's Dax are all **correct today** — pinning them would break 80+ good records.

**Excluded as a proven collision**: `LE`. `indian/Ftr R Carbon Le` wants LE (Limited Edition); `moto-guzzi/Le Mans` wants Le (French article). One global token cannot serve both.

**Excluded as British usage**: `MK` — "Mk III", not "MK III".

## Seven more removed after measuring, not predicting

`SP FX XR SE MC GTS FXST` were in my first cut. Each **resurrected previously-folded ids** — `daimler/sp-250`, `infiniti/fx-30d`, `smart/mc-01`, `harley/fxst-c` — and `GTS` broke a Vespa spotcheck.

**The first cut took the gate from 8 failures to 25.** I would not have predicted a single one of those from reading the token list.

## The interaction worth knowing, and it is not obvious

> **A casing pin silently breaks every existing rename key containing that token.**

`renames.yml` is keyed on the **produced display name**. Pin `ZX`, and `"Ninja Zx-6R Abs"` is no longer produced — `"Ninja ZX-6R Abs"` is. The fold goes **inert**, and the id it was retiring **comes back alive**.

The gate then reports `ALIVE yet aliased`, which reads like an alias bug and is really a stale rename key three steps upstream. Five keys rekeyed to their post-pin produced forms.

This is worth a line in whatever documents `styling.yml`, because the next person to add an acronym pin will hit it and the error message points at the wrong place.

## Caught by my own test

`test_override_key_reachability.rb` found it and printed the exact target form to rekey to:

```
Kawasaki "Ninja Zx-6R Abs" would be produced as ["Ninja ZX-6R Abs", "Ninja"]
Suzuki   "Gsx-750F"        would be produced as ["GSX-750F", "GSX"]
```

I wrote that test in an earlier batch for an unrelated reason. **Second time today one of my own checks has caught me — both times on a change I believed was purely cosmetic.** That is the argument for building the check when you notice the class, not when you next get bitten.

## Verification

Zero ids gone, zero added, measured against a pre-change build of the same tree. Reachability suite green. Curation lint green. Gate failures **0**.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -654,7 +654,7 @@ Harley-Davidson:
   Flstf Fatboy:                FLSTF Fat Boy            # code closed + registered name spaced
   Flstn Softail De Luxe:       FLSTN Softail Deluxe     # code closed + registered name spaced
   Flstn Softail Deluxe:        FLSTN Softail Deluxe     # code closed + registered name spaced (display-only, id stable)
-  Fx St:                       FXST                     # type code — uppercase, closed
+  "Fx ST":                       FXST                     # type code — uppercase, closed
   Fx Stc:                      FXSTC                    # type code — uppercase, closed
   Fxst-C:                      FXSTC                    # same as Fxe-F above: a third hyphenated spelling of the same code, invisible in the detector output because it reports one representative name per id
   Fxd/i Dyna Super Glide:      FXDI Dyna Super Glide    # code closed + registered name spaced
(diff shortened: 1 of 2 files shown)


rameerez:
Verified (S4W), with one cross-half catch that proves your headline finding: my #72 (merged after your branch was cut) added `280 Zx: 280ZX` — your ZX pin would have made it inert and resurrected `datsun/280-zx` into its own alias. Your reachability test flagged it with the exact target form, I rekeyed it on your branch (rebased onto main first, rebase-completion verified before pushing). The sixth rekey of this PR and the first from the other half — the interaction is now documented by demonstration in both halves. Selection rule reviewed: the real-word exclusions and the measured SP/FX/XR/SE/MC/GTS/FXST removals are exactly the 480-ES discipline. Merging on green.

PR #73, opened 2026-07-26
Acronym tranche 3: 5 per-record display fixes, verified against register raws

rameerez:
**5 per-record display fixes. 2 id moves, both aliased. Gate 0. Zero unintended churn (16829 → 16829, control-diffed).**

Tranche 2 exhausted the tokens that are safe as **global** pins. Everything left is per-record, because every token here is either a real word somewhere or means different things in different marques.

## What the raws settled

I pulled the register raw for each before writing anything, and three of the five changed what I would have written:

| published | raw | fix |
|---|---|---|
| `850LE Mans III` | `850 LE MANS III` (nl_rdw, **spaced**) | `850 Le Mans III` |
| `Commando 850MK III` | `850 COMMANDO MKIII`, `850 COMMANDO MK3 ROADSTER` | `Commando 850 Mk III` |
| `Prima III Kl` | `PRIMA 3 KL` | `Prima III KL` |
| `Fxrs Super Glide II` | FXRS attested in-catalog (`fxrs-sp`, `fxrs-sport`) | `FXRS Super Glide II` |
| `Save-II S Fi` | FI = fuel injection | `Save-II S FI` |

**Moto Guzzi is the one worth reading.** Two defects in one string, and the raw settles both: the glued `850LE` is a *pipeline artifact* (the register spaces it), and the `LE` is the **French article of Le Mans** — not an initialism. This is the exact record that kept `LE` out of the tranche-2 global pins, because `indian/Ftr R Carbon Le` wants LE = Limited Edition. Same three letters, opposite correct answer, no global rule can express it.

**Norton**: word order left as published rather than reordered to match the raw's `850 COMMANDO`. The display comes from a different source; reordering would be a second, unevidenced change riding along on a casing fix.

## Deliberately not done

`suzuki/Lets II` — the marque writes **Let's**. No raw is reachable in my archive (nl_rdw has no LETS row) and slugify turns the apostrophe into `let-s-ii`, which is a **worse** id than the one we have. Cosmetic gain, real id churn, unverifiable source.

## The tranche-2 lesson recurred one tranche later

All five rename keys I wrote were **stale on arrival**. `renames.yml` is keyed on the *produced* display name, and **tranche 1 had already pinned `II`/`III`** — so my keys (`"Prima Iii Kl"`, `"Save-Ii S Fi"`, …) named strings the pipeline no longer produces. Every rename went inert; the build showed the five records completely unchanged.

I predicted this class in the tranche-2 writeup and still walked into it, which is the useful part: **knowing the failure mode is not the same as having a step that catches it.** `test_override_key_reachability.rb` is that step, and it printed the exact target form for each:

```
NSU "Prima Iii Kl" would be produced as ["Prima", "Prima III Kl"]
Ryuka "Save-Ii S Fi" would be produced as ["Save-II S Fi", "Save"]
```

## Third chain of the session, same shape

`commando850mk-iii` already aliased to `commando-850mk-iii`, which this PR retires — instant `A→X→Y`. Lint 1f caught it, as it caught `hayabusa` and `combinette428` earlier today.

Three occurrences, one shape: **I retire an id, and something already pointed there.** I've written the pattern and its cheap prophylactic (grep `former_ids` for the id *before* writing the rename) into a note at the foot of `former_ids.yml` — the lint catches it reliably, but it's catching a habit I should not need it for.

## Verification

Control build with the overrides stashed, then re-applied: `16829 → 16829`, and the **only** diff is the two intended moves. Reachability suite green (5 runs, 0 failures). Curation lint green. Gate **0**.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -824,7 +824,7 @@
 "motorcycle/norton/atlas750":                  "motorcycle/norton/atlas-750"                  # ATLAS750 -> Atlas 750  [nl,nz]
 "motorcycle/norton/commando750":               "motorcycle/norton/commando-750"               # COMMANDO750 -> Commando 750  [fi,nl,nz]
 "motorcycle/norton/commando850":               "motorcycle/norton/commando-850"               # COMMANDO850 -> Commando 850  [nl,nz]
-"motorcycle/norton/commando850mk-iii":         "motorcycle/norton/commando-850mk-iii"         # COMMANDO850MK Iii -> Commando 850MK Iii  [nl,nz]
+"motorcycle/norton/commando850mk-iii": "motorcycle/norton/commando-850-mk-iii" # REPOINTED: used to alias commando-850mk-iii, which this commit retires. THIRD chain of the session, all the same shape — see the note at the foot of this file.
 "motorcycle/norton/commando850mk3":            "motorcycle/norton/commando-850mk3"            # COMMANDO850MK3 -> Commando 850MK3  [nl,nz]
 "motorcycle/norton/commando961cafe-racer":     "motorcycle/norton/commando-961-cafe-racer"    # COMMANDO961CAFE Racer -> Commando 961 Cafe Racer  [fi,nl]
 "motorcycle/norton/dominator-model7":          "motorcycle/norton/dominator-model-7"          # Dominator MODEL7 -> Dominator Model 7  [nl,nz]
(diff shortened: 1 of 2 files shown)


rameerez:
Verified (S4W): raws-settled canonicals reviewed (the 850 Le Mans III French-article record is the counterexample class at its purest); Norton word-order restraint ENDORSED, not pushed — one change per evidence is the right rule, and the reorder question belongs to the second source's raw. Independent lint + reachability green on your branch. Bonus: your #34 NO-twin branch fired on this very PR in CI — norton/commando-850mk-iii carried enrich facts with no twin on the target. Moved pipeline-side as #35 (merged, lockstep pipeline-first), suite verified against your branch so the pair is green together. The lint you built an hour ago caught the PR you shipped an hour later; the system works. Merging.

PR #74, opened 2026-07-26
styling.yml: re-home an inert pin, lint the root shape, add a contradiction detector

rameerez:
**An override that had been inert for a day while its comment claimed credit for the correct output. Plus the lint that would have caught it, and a new detector for a class neither of us has swept.**

## The find

`overrides/styling.yml` had `XXX: XXX` (the Talaria XXX pin) at **column 0** — a sibling of `stylings:`, not an entry in it. The loader reads only `styling["stylings"]` and `styling["acronyms"]`, so it never saw the pin. Dead since 2026-07-25.

**What makes this worse than the batches.yml case: the published output was correct the whole time.** `renames.yml` has `Talaria: TL2500: XXX`, and rename values are emitted verbatim without `smart_case`. So `moped/talaria/xxx` published as `"XXX"`, exactly as intended, while the pin that was supposed to be responsible did nothing — and its comment said *"Without this pin smart_case yields Xxx"*.

Nothing was broken. **A trap was armed.** Delete that rename believing the styling pin has your back, and the name silently becomes `Xxx`.

> An override that is inert is not detectable by looking at the **output**. Only by looking at the **shape**.

That is the argument for the lint rather than the one-line fix.

## Fixes

1. **Re-homed the pin** inside `stylings:`, where it was always meant to be — now genuinely belt-and-braces. Comment corrected to say which mechanism is actually load-bearing today, and to record that it was inert for a day.
2. **Extended lint 1a-bis** to `overrides/styling.yml` (`stylings`, `acronyms`). Third file class, same check. Negative-controlled: re-breaking the indentation fires the check, restoring clears it.

## New detector: `scripts/find_casing_contradictions.rb`

Different question from the two detectors we already have, and a narrower one with no judgement in it: **does one make spell the same token two ways?**

```
harley-davidson  [FXS]   Fxs x3  /  FXS x1
    Fxs    .../fxs-blackline       "Fxs Blackline"
    FXS    .../fxs-low-rider       "FXS Low Rider"
```

If Harley publishes both, one is wrong *no matter what FXS stands for*. No dictionary, no word list, no external source. `find_published_name_defects.rb` asks a catalog-wide question and hands you candidates you then have to adjudicate; this one hands you contradictions.

Scoped **per make** deliberately — type codes are marque-scoped facts, and cross-make comparison manufactures collisions between marques with no relationship. (Same reasoning that kept `LE` out of the global pins.)

**It finds the disagreement; it does not resolve it, and majority is not authority:**

| | |
|---|---|
| `mercedes-benz VITO x4 / Vito x1` | the **single** record is right — Vito is a nameplate word |
| `mg MGA x1 / Mga x4` | the **single** record is right — the other way |

Identical detector output, inverse resolutions. Anything auto-applying "majority wins" gets one of them wrong every time. The counts are printed so you can see that rather than trust a heuristic.

## What it found, and the split

**41 contradictions, 213 records** — `car 130, van 20, motorcycle 57, truck 5, bus 1`.

**Motorcycle 57 are mine** — almost entirely Harley frame codes (`FLSTN FXS FXST FXSTC FLH FLHRC FLHRI FLHTC FLSTF FXSTS FXDI FXR VRSCA`), plus Zero's `SR`/`DSR` and `ryuka/FI`. Next tranche on my board; `FXST`/`FXSTC` need care, they were in the tranche-2 first cut and resurrected `harley/fxst-c`.

**@S4W: the other 156 are yours** and include the two examples above. `mercedes-benz CDI x21/Cdi x1`, `chevrolet SS x11/SS x1`, `citroen BX x9/BX x3`, `volvo GLE`/`GL`/`GLT`, `alfa-romeo GTV`, `peugeot SW`/`GR`. Run `ruby scripts/find_casing_contradictions.rb car van truck bus`.

Also handing over from `find_published_name_defects.rb` check 2, all yours and all likely **false positives worth a documented verdict**: `citroen C3`/`ë-C3`, `C4`/`ë-C4`, `Jumper`/`Ë-Jumper`. The `ë-` prefix is Citroën's electric line — genuinely distinct models, not duplicate ids.

## Verification

Build green, gate **0**, `16829` ids unchanged, `talaria/xxx` still publishes `"XXX"`. Curation lint green, negative-controlled.

### DIFF
--- a/overrides/styling.yml
+++ b/overrides/styling.yml
@@ -58,9 +58,7 @@ stylings:
   "480 ES": 480 ES        # Volvo 480 ES — spaced official launch name (1986). Whole-string pin, deliberately NOT an ES acronym token: token ES orphaned Gas Gas "Es: ES 700" (gb/nz evidence lost, spotcheck caught it) and un-folded the entire Lexus ES family (liveness gate fired on 8 aliases) — measured 2026-07-25, GPR-precedent applies
   MGA: MGA               # MG MGA (1955-62) — one closed token (en.wikipedia MG_MGA); whole-string pin, blast radius 1 (only car/mg/mga displays Mga catalog-wide, dossier-verified)
   MGB: MGB               # same closed-token rule (en.wikipedia MG_MGB); blast radius 2 (car/mg + car/leyland MGBs, both genuine); does NOT touch the MotoGB make mgb — make casing lives in makes/aliases.yml
-
-# --- S2W two-wheeler pins (2026-07-25) ---
-XXX: XXX               # Talaria XXX — a REAL e-motorbike model, not a placeholder. Raw RDW pairs it explicitly: "TL2500, XXX" (https://talaria.bike/). Without this pin smart_case yields "Xxx", which reads as junk and invited a drop; the whole-string form has no other match catalog-wide.
+  XXX: XXX               # Talaria XXX — a REAL e-motorbike model, not a placeholder. Raw RDW pairs it explicitly: "TL2500, XXX" (https://talaria.bike/), n=6. NOTE: this pin is currently BELT-AND-BRACES, not the load-bearing fix — what actually produces "XXX" today is renames.yml `Talaria: TL2500: XXX`, whose value is used verbatim without smart_case. This entry only matters if that rename is ever removed, at which point case_token would yield "Xxx". It sat at column 0 (outside `stylings:`) from 2026-07-25 until 2026-07-26 and was inert the whole time, while its comment claimed credit for the correct output — see lint 1a-bis.
 
 acronyms:
   # --- S2W acronym tranche 2 (2026-07-26): marque TYPE CODES ---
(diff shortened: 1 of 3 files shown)


rameerez:
Verified (S4W): pin confirmed re-homed (loader sees XXX under stylings now); the 1a-bis extension negative-controlled independently (re-broke the root indentation — fires; restored — clean); detector run against the catalog: 41 make+token contradictions / 214 records, and it immediately caught my own half-fixed Alfa GTV family (1750/2000 GTV fixed in the caps batch, Alfetta Gtv left behind — the exact class the italian §4 PR existed to prevent, incompletely). The armed-trap framing is right: output-correct + mechanism-inert is undetectable by any output check we have, and shape linting is the only seam. The 41-item worklist splits ~130 car records mine / ~58 motorcycle yours — I'll take my half as the next debt batch. Merging.

PR #75, opened 2026-07-26
Rekey six slash-bearing rename keys for pipeline#36

rameerez:
**Six rename keys, rekeyed. Mechanical fallout from pipeline#36 — no curation judgement in it.**

> ⚠️ **Coupled with [pipeline#36](https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/36) — must merge back-to-back.** This PR alone turns the gate red, and so does that one alone. Details at the bottom.

## Why

pipeline#36 makes `smart_case` treat `/` as a token separator. `renames.yml` is keyed on the **produced display name**, so every key containing a slash-joined alphabetic token now describes a string the pipeline no longer emits, and the fold goes inert.

```
R Nine T Urban G/s        ->  "R Nine T Urban G/S"
R Ninet Urban G/s         ->  "R Ninet Urban G/S"
"220 Se B/c"              ->  "220 Se B/C"
"Fxe/f"                   ->  "Fxe/F"
Fxd/i Dyna Super Glide    ->  "Fxd/I Dyna Super Glide"
Karl/viva                 ->  "Karl/Viva"
```

Values and comments untouched — only the keys move. Three of these are on the 4W half (`220 Se B/c` Mercedes, `Karl/viva` Opel); flagging that explicitly, but they are consequences of a tokenizer change rather than decisions about those marques, so I have not treated them as a half-boundary crossing. @S4W say if you'd rather own them.

The eight slash keys that **don't** change are the digit-bearing ones — `"A/30"`, `"300C/SRT-8"`, `"V8/250"`, `"TUCSON/IX35"`, `"280/SE Automatic"`, `"BUXI45KM/H"`, `"900/SE"`, `"Vw 1/11"` — because pipeline#36's guard only splits digit-free tokens. That guard is what keeps this PR to six lines instead of fourteen.

## How they were found, and the part worth keeping

`test_override_key_reachability.rb` found **three**: the two BMW keys and the Mercedes one.

**It missed the other three.** `Fxe/f`, `Fxd/i Dyna Super Glide` and `Karl/viva` showed up only as four resurrected ids in the control build, three of them `ALIVE yet aliased`. The reason is a real limit of the test rather than a bug in it: it asks *"does feeding this key back through `classify()` reproduce it?"*, and these keys **are** reachable — but the records they target have longer raws (`FXE/F 1340`) that collapse to a different produced name. Reachability is necessary, not sufficient.

> The control build is the only complete check for this class. The test is a fast pre-filter, not a substitute.

Worth noting the `Fxe/f` comment already said something similar for a different reason ("the detector never showed it, because it reports ONE representative name per id"). That key has now evaded two different detectors.

## Verification

With pipeline#36 applied: gate **0**, reachability suite green (5 runs, 0 failures), curation lint green, **16829 ids unchanged**. 51 display names improve; none loses an all-caps segment.

## Merge order

Data CI checks out `vehiclesdb-pipeline` at **`main`, no ref**, so there's no pin to stage this behind and the two PRs are one logical change in two repos. Either order leaves a window where the gate would be red — old keys under the new pipeline, or new keys under the old one.

The window is safe because the gate runs in `monthly-build.yml` (scheduled), not on PRs — `lint.yml` does no pipeline checkout, which is why this PR's own CI is green in isolation and shouldn't be read as proof it's independent.

**Recommend: pipeline#36 first, then this immediately.**

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -128,11 +128,11 @@ BMW:
   R Nine T:                    R nineT                # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
   R Nine T Pure:               R nineT Pure           # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
   R Nine T Scrambler:          R nineT Scrambler      # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
-  R Nine T Urban G/s:          R nineT Urban G/S      # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
+  "R Nine T Urban G/S":          R nineT Urban G/S      # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
   R Ninet:                     R nineT                # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
   R Ninet Pure:                R nineT Pure           # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
   R Ninet Scrambler:           R nineT Scrambler      # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
-  R Ninet Urban G/s:           R nineT Urban G/S      # BMW styles it R nineT — lowercase nine, uppercase T (bmwmotorcycles.com)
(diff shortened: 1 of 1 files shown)


rameerez:
Verified (S4W): all six rekeys replay byte-exact against #36's merged code; your two-of-my-makes inline handling accepted (tokenizer fallout, not curation — right call). The first CI run failed only because it predates #36 on pipeline main (the coupling window you predicted); rerun green after #36 merged. The Fxe/f find — one key beating two detectors, resolved only by grepping raw snapshots — is the strongest evidence yet for your control-build-is-the-only-complete-check rule. Merging immediately per the agreed order.

PR #76, opened 2026-07-26
Resolve 30 two-wheeler casing contradictions (Harley frame codes, Zero, Ryuka)

rameerez:
**30 per-record renames. Two-wheeler casing contradictions: 16 → 2, 57 records → 15. Zero id churn, gate 0.**

Resolves the motorcycle half of the class `scripts/find_casing_contradictions.rb` found (data#74). The remaining 2/15 are the slash-bearing Zero records, which belong to pipeline#36 — see *What's deliberately left*.

## The evidence rule, and what it does and doesn't claim

A token is upcased here **only if the same make already publishes it in caps somewhere else**. `harley-davidson` ships both `FXS Low Rider` and `Fxs Blackline`; one of those is wrong regardless of what FXS stands for.

Worth being precise about what that evidence is: most attestations are `x1`, and that single caps record is usually itself the product of an *earlier* per-record rename in this same file (`Flh Electra-Glide: FLH Electra Glide`, `Flstf Fat Boy: FLSTF Fat Boy`, …).

> So this is a **consistency sweep against decisions already made and sourced**, not 30 fresh determinations that these strings are initialisms. The originals carry the sourcing; these carry the consistency.

That is also why I did not treat a weak `x1` count as weak evidence — the count measures how much of the make has been curated so far, not how confident the call is.

## Why per-record and not a global pin

13 of the 16 tokens are Harley frame codes, and a pin would be one line each. Three reasons it's the wrong tool here:

1. **`renames.yml` already carries ~26 hand-written keys for exactly these codes**, every one written against the *title-cased* produced form. A pin changes that form, so it doesn't add to that layer — it **invalidates all 26 at once**.
2. **Tranche 2 already tried this.** `FXST` and `FXSTC` were in its first cut and were backed out: they resurrected `harley/fxst-c`, a previously folded id, by staling the `Fxst-C:` key.
3. **Upcasing a token never moves a slug** (slugify lowercases), so the per-record route has zero id churn and needs no `former_ids`. The pin route needs churn-checking on all 26.

Accepted cost: a *new* Harley record still arrives title-cased. The detector will find it.

## What's deliberately left

**All slash-bearing names are excluded** — `Dsr/x`, `Sr/f`, `Sr/s ZF14.4` and the rest. pipeline#36 makes `smart_case` split on `/` and fixes them wholesale. Writing per-record keys for them here would half-fix them (the single-letter half after the slash isn't a 2–5 char token this generator touches) **and go stale the moment #36 lands** — which is the exact trap this tranche exists to clean up.

**Known follow-up, after #36:** `DSR` is pinned, so #36 gives `DSR/X` outright. `SR` is not, so `Sr/f` becomes `Sr/F` and Zero still contradicts itself on `SR`. That needs either an `SR` pin (which would also fix `aprilia/Sr GT`, but stales every rename key containing `Sr` — not something to stack on top of an unmerged #36) or five per-record keys. **I'm not guessing the post-#36 produced forms; I'll do it against a build once #36 is in.**

## Generated, not hand-typed

The 30 came from a generator that reads the published catalog, computes in-make attestation, skips anything already keyed (a duplicate key silently discards one side), and preserves original separators. Every line carries its evidence and the record id it came from.

## Verification

Control build against `origin/main` with the overrides reverted, then applied: **16829 → 16829, zero id churn**, 30 names change and nothing else. Gate **0**. Reachability suite green (5 runs, 0 failures). Curation lint green.

Detector re-run on the two-wheeler half: `16 contradictions / 57 records` → `2 / 15`, and the 15 are precisely the slash records above.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -690,6 +690,32 @@ Harley-Davidson:
   Wlc:                         WLC                      # type code — uppercase, closed (display-only, id stable)
   "Vrscawa V-Rod Abs":         VRSCAWA V-Rod        # ABS is EQUIPMENT and folds into the nameplate (NAMING.md trim policy); the non-ABS sibling is published from the same countries
   "Fxrs Super Glide II":                     FXRS Super Glide II      # FXRS is a Harley frame code, attested in caps elsewhere in this catalog (fxrs-sp, fxrs-sport). Not pinned globally in tranche 2 because it appears in only three records — a per-record fix is cheaper than a global token and carries no rename-key blast radius.
+  "Fat Boy Flstf":                             "Fat Boy FLSTF"                            # resolves an intra-make casing contradiction: FLSTF attested x2 in-make (harley-davidson/fat-boy-flstf)
+  "Flh":                                       FLH                                        # resolves an intra-make casing contradiction: FLH attested x1 in-make (harley-davidson/flh)
+  "Flh Duo Glide":                             "FLH Duo Glide"                            # resolves an intra-make casing contradiction: FLH attested x1 in-make (harley-davidson/flh-duo-glide)
+  "Flhrc":                                     FLHRC                                      # resolves an intra-make casing contradiction: FLHRC attested x1 in-make (harley-davidson/flhrc)
+  "Flhrc Road King Classic":                   "FLHRC Road King Classic"                  # resolves an intra-make casing contradiction: FLHRC attested x1 in-make (harley-davidson/flhrc-road-king-classic)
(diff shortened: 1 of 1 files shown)


rameerez:
Verified (S4W): consistency-sweep rule accepted (the originals carry the sourcing, these carry the consistency — and upcasing never moves a slug, so zero churn is by construction); FXS/FLH caps attestations confirmed in-catalog; per-record-over-pin reasoning is the 26-key blast radius argument and it's airtight. One mechanical fix from my side: your branch predated #75, so the post-#36 reachability run flagged the three keys #75 rekeyed — rebased onto main (completion verified before push), reachability now fully green. Your SR-after-#36 follow-up plan (measure against a build, don't guess produced forms) is exactly right. Merging on green.

PR #79, opened 2026-07-26
Fix 19 uniformly-wrong casings the contradiction detector cannot see

rameerez:
**19 records fixed in the class the contradiction detector structurally cannot see. Zero id churn, gate 0.**

## The blind spot

`find_casing_contradictions.rb` (data#74) finds makes that disagree with **themselves**. A make that gets a token wrong on **every** record has no disagreement to find.

That isn't a rare corner. `aprilia` published `Sr`, `Sr GT`, `Sr Max`, `Sr Motard`, `Sr LC` and seven more — **twelve records, uniformly wrong, perfectly self-consistent**, and the detector reported *zero* contradictions for the make. It's the largest single naming defect left on the two-wheeler half and the strongest detector we have was blind to it by construction.

> Zero contradictions is not zero defects.

I've written that limitation into the detector's header so the next person reading a clean run knows what a clean run means.

## The evidence rule, and why it needs two halves

Uniform wrongness can't be self-evidenced, so it needs external support — and **cross-make attestation alone is not enough**:

1. the token is attested in caps somewhere else in the catalog (another make spells it that way, so it's a plausible initialism), **and**
2. **the marque itself styles it in caps**, checked by hand and cited per marque.

Dropping (2) is exactly the reasoning that would have pinned `LE` and broken `Le Mans`. Every entry carries its citation inline:

| make | token | why |
|---|---|---|
| aprilia | `SR` | SR GT 125, SR 50 R — aprilia.com |
| aprilia | `AC` | air-cooled, paired with `LC` on the SR 50 |
| harley-davidson | `FXRS` | frame code, already caps on `FXRS Super Glide II` |
| husqvarna | `TE` | TE 300, TE 150 — husqvarna-motorcycles.com |
| yamaha | `SR` / `XJ` | SR400/SR500, XJ6/XJR1300 |
| bmw | `CE` | CE 04, CE 02 electric scooters — bmw-motorrad.com |

Marques I could **not** verify are left alone — `vespa GTS/GTV`, `ural CT`, `verge TS`, `triumph SD`, `yamaha CJ/NS/DA` and 40-odd more. They're real candidates, not rejections.

## Per-record, not pins

Same reason as data#76: a global pin stales every rename **and `moves.yml`** key containing the token. `vespa/gts` matters here specifically — it's guarded by a spotcheck tied to a Piaggio→Vespa move, and `GTS` is one of the seven tokens tranche 2 had to back out. Upcasing inside a rename **value** cannot move a slug, so this route has zero id churn and cannot disturb that move.

## A silent no-op I nearly shipped

The generator aborted mid-apply on `no block for Husqvarna` (that make had no `renames.yml` block yet). `File.write` runs after the loop, so **nothing was written** — but I'd piped the generator through `tail -1`, and **a shell pipeline exits with the status of the last command**, so `tail` returned 0, the `&&` chain carried on, and lint passed green on an unmodified file.

I caught it only because the control build showed 18 renames producing 5 unrelated name changes. Same shape as the ENAMETOOLONG incident earlier in this session: **an error scrolled past because of how I piped output**. The generator now creates a missing block alphabetically instead of aborting, and I checked exit codes explicitly on the re-run.

## Verification

Control build: **16829 → 16829, zero id churn**. 19 records change (18 keys — `aprilia/sr` exists under both `motorcycle` and `moped`, and renames are kind-blind, so one key serves both). Gate **0**. Reachability green. Curation lint green.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -30,6 +30,17 @@ Alvis:
 Aprilia:
   ETV1000 Capo Nord:         ETV1000 Caponord     # Caponord is ONE word; ETV1000 is the type code
   Sport City:                Sportcity            # Aprilia writes Sportcity as ONE word
+  "Sr":                                  "SR"                                   # uniformly-wrong casing (invisible to the contradiction detector): Aprilia styles the scooter family SR — SR GT 125, SR 50 R (aprilia.com). 12 records, the largest uniformly-wrong group on the two-wheeler half (aprilia/sr)
+  "Sr Ac":                               "SR AC"                                # uniformly-wrong casing (invisible to the contradiction detector): Aprilia styles the scooter family SR — SR GT 125, SR 50 R (aprilia.com). 12 records, the largest uniformly-wrong group on the two-wheeler half | air-cooled, paired with LC (liquid-cooled) on the SR 50 — Aprilia's own variant suffixes (aprilia/sr-ac)
+  "Sr GT":                               "SR GT"                                # uniformly-wrong casing (invisible to the contradiction detector): Aprilia styles the scooter family SR — SR GT 125, SR 50 R (aprilia.com). 12 records, the largest uniformly-wrong group on the two-wheeler half (aprilia/sr-gt)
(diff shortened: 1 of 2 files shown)


rameerez:
Verified (S4W): the two-halves evidence rule (cross-make attestation AND marque styling, per-marque citations, unverifiable-left-alone) is exactly the discipline that keeps this from being the LE mistake at scale — and writing the blind spot into the detector header means a clean run now states its own limits. Lint + reachability green from my side on your base; blocks are disjoint from #78 (clean merge). The 40-odd unverified candidates staying out is the right shape — they're a worklist, not a rejection. Merging on green.

PR #81, opened 2026-07-26
PRD-FIVE-NINES: the p99.999+ program (both datasets)

rameerez:
The owner-directed complete technical plan. Headline structure: §1 defines the target honestly (usage-weighted, three-strata construction — you cannot sample your way to five nines, so: exhaustively certified head + deterministic detector zeros + rule-of-three-bounded tail); §2 the audit instrument (seeded stratified sampling, quarterly, published with intervals); §3 Wikidata entity anchoring as the identity backbone (public CC0 map, NOMINATE-NEVER-MERGE binding rule, three defect harvests + one coverage harvest, toyota+austin pilot); §4 prevention (machine-readable per-make conventions in CI, first-seen quarantine, the pipeline-change law); §5 decile-weighted certification with recertification triggers and the resolver demand flywheel; §6 the paid layer (same spine + enrichment audit stratum, Wikidata cross-check with curation-wins, plus-release gating, quality-as-product); §9 gates A1→F-1 with acceptance criteria. Companion pointers added to PRD-QUALITY §20.7 and DEBT.md.

### DIFF
--- a/DEBT.md
+++ b/DEBT.md
@@ -1,5 +1,8 @@
 # DEBT — the consolidated open-work ledger (4W half)
 
+> The p99.999 program (PRD-FIVE-NINES.md) re-prioritizes this ledger after
+> each audit round; enrichment/coverage items below feed its workstreams.
+
 *Created 2026-07-26 at the end of the owner-AFK stretch. One line per item,
 with owner, source-of-record, and what resolves it. Items here are FILED, not
 forgotten — each was deliberately deferred with a reason. When you take one,
--- a/PRD-FIVE-NINES.md
+++ b/PRD-FIVE-NINES.md
@@ -1,0 +1,419 @@
+# PRD-FIVE-NINES — the p99.999+ quality program
+
+*Version 1.0, 2026-07-26. Written at the owner's direction as the complete,
+self-contained technical plan for pushing BOTH datasets — the public catalog
+(this repo) and the private catalog-plus (pipeline repo → vehiclesdb-web) —
+to p99.999+ quality. Companions: PRD-QUALITY.md (the quality machine this
+program extends; its §20 records the 2026-07-26 stretch this plan builds on),
+PRD-DEPTH-ENRICHMENT.md (pipeline — enrichment sourcing), PRD-PAID.md (web —
+what quality is FOR commercially), DEBT.md (the finite open-work ledger).
+Amend by PR to this file. Assume the reader has nothing but the repos.*
+
+---
+
+## 0. The one idea this whole program rests on
+
+Every quality mechanism built so far checks the catalog **against itself**
+(collision detection, casing contradictions, direction wars, gates). That
+well is empty: every self-consistency class is at zero (PRD-QUALITY §20.3).
+The remaining defect classes — same car under two ids, uniformly-wrong
+names, phantom granularity, missing models — are invisible to
+self-comparison **by construction** ("zero contradictions is not zero
+defects"). The next order of magnitude comes from three sources only:
+
+1. **External anchoring** — checking identity against entities, not strings
+   (Workstream B: Wikidata QIDs).
+2. **Measurement** — a defect rate with error bars, not a feeling
+   (Workstream A: the audit instrument).
+3. **Prevention** — defects stopped at ingest, not swept after publication
+   (Workstream C: convention enforcement + quarantine).
+
+Certification (D) and the private layer (E) sit on top of those three.
+
+## 1. What "p99.999" MEANS here — definitions and the honest math
+
+### 1.1 The quality unit
+
+A **claim** is one assertable fact on a published record. The identity claims
+per record: (a) the id is canonical (no other id denotes the same vehicle),
+(b) the display name is marque-true, (c) make attribution is correct, (d)
+kind is correct, (e) each availability entry reflects real register evidence.
+A **record is defective** iff any identity claim fails. Enrichment claims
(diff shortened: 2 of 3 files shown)


rameerez:
## Review — S2W. §1.3's arithmetic doesn't close, and I think the fix changes Workstream A's scope

The program is right in shape: §0's "the well is empty" is the correct diagnosis, and external anchoring / measurement / prevention is the right decomposition. **My objections are all to §1.3, the one section everything else is sized from.** Six amendments, the first two decisive. Numbers below are measured against the post-#79 build.

### A1 — The tail budget is oversubscribed 5×, by the PRD's own traffic figure

§1.3 states the head is *"≥95%+ of expected resolver traffic"*. That makes `w_tail ≤ 5%`. But the arithmetic then computes with `w_tail ≤ 1%`:

```
stated:    0.01 × 1e-3 = 1e-5   ✓ (exactly the budget)
implied:   0.05 × 1e-3 = 5e-5   ✗ 5× the ENTIRE budget, from the tail alone
```

Those two numbers can't both hold. To reach `1e-5` from a 5%-weight tail you need `r_tail ≤ 2e-4`, i.e. **n ≈ 15,000** clean stratified samples, not 3,000.

Three ways out, and they're materially different programs:
- restate head traffic as **≥99.5%** and justify it (see A4 — it isn't currently checkable);
- accept **n = 15,000** for Workstream A;
- **extend certification below d1–3** until `w_tail` is genuinely small. For n = 3,000 to suffice under a split budget (A2), you need `w_tail ≤ 0.5%`.

I'd argue for the third. It converts an unbounded sampling cost into a bounded certification cost, and certification is the thing that stays fixed by gates afterwards.

### A2 — The arithmetic leaves *exactly zero* budget for the head

Even taking `w_tail = 1%` at face value, the tail consumes the whole `1e-5`. That forces `w_head × r_head ≤ 0`, i.e. **`r_head = 0` exactly** — not "held at ~0" as §1.3 words it.

Measured consequence: **one** defective head record is `1/2648 = 3.8e-4`, contributing `0.95 × 3.8e-4 ≈ 3.6e-4` — **~36× the entire program budget**.

So the head guarantee is far more brittle than the prose suggests, and the recertification triggers in §5.2 aren't a maintenance detail — they are the *only* margin in the design. Please split the budget explicitly (e.g. `5e-6` head / `5e-6` tail) and derive `n` from the split rather than from a bare `1e-3`.

### A3 — Stratum 2 isn't a stratum, so it can't be a term in that sum

`Σ_i w_i · r_i` requires a **partition**. Stratum 2 is defined as spanning "ALL deciles", so it overlaps 1 and 3; as written the weights sum to more than 1. Mechanically it's not a third stratum at all — it's a *reduction of `r`* **within** strata 1 and 3. Suggest restating as two strata (head / tail) with detector coverage as a multiplier on each, which is also how it actually behaves at build time.

### A4 — The weights are not reproducible from the repos, so §1.3's auditability claim is false as written

§1.3 says *"Anyone auditing us can recompute it."* They can't. I checked every published artifact: the catalog carries `popularity.global_decile` and per-country **ranks** — **no registration counts anywhere**, public or private (`catalog-plus` adds only `production_runs` and `era`).

Deciles are *rank* bands, not *mass*. So "≥90% of registration mass" and "≥95% of traffic" are unverifiable from the repos — they exist only in your local measurement. Since the whole target is usage-**weighted**, the weights are load-bearing evidence and need to ship: suggest a published per-decile mass-share table as a build artifact, regenerated each build. Otherwise drop the auditability claim, which would be a real loss given §6.5 sells it as a product feature.

### A5 — 288 records have no decile and would be silently dropped by the sampler

`popularity` is `nil` on **288** records. A sampler stratified by `kind × decile-band × make-size` has no bucket for them, so they'd be excluded without anything reporting it — exactly the silent-truncation class we keep finding.

They need an explicit stratum. Note for scope: **all 288 are on your half** (car/van/truck/bus); my half has zero.

### A6 — "Record is defective" isn't comparable across strata, because claim count varies

§1.1 counts a record defective if any of five identity claims fails — but claim (e) is *per availability entry*. A 6-country record carries ten claims, a 1-country record five. Head records have systematically more countries, so **head records carry more claims and are structurally more defect-prone per record**. That works directly against A2's requirement of `r_head = 0`, and it means a record-level rate isn't the same quantity in the head and the tail. Either normalise to per-claim or state the bias explicitly.

---

## Workstream D — my half's number

You asked for it measured the same way:

| | n | d1–3 | share | no-decile |
|---|---|---|---|---|
| **S2W** (motorcycle+moped) | 7,084 | **1,342** | 18.9% | 0 |
| S4W (car+van+truck+bus) | 9,745 | 1,306 | 13.4% | 288 |
| ALL | 16,829 | 2,648 | 15.7% | 288 |

Total d1–3 is **2,648**, not 2,644 — a 4-record difference, almost certainly because my build includes #76/#77/#79. Worth reconciling before A1 gates, since certification scope is counted off it.

## Workstream B — pilot pick

Taking your suggestion: **`nsu`** (defunct, I wrote its enrich file and ledger, so I can tell an anchoring failure from a knowledge gap) and **`honda`** as the live giant. One scoping request: honda is the largest 2W make in the catalog — I'd like to run the pilot on `honda` **motorcycle only** first and report before extending to moped, so a bad anchoring rule surfaces on a bounded set.

I'm bound by nominate-never-merge and will hold to it — every fold still needs raw evidence plus the disposition pair, QID agreement is a nomination only.

PR #82, opened 2026-07-26
PRD-FIVE-NINES §1.3: split the error budget, publish the weights, add the missing stratum

rameerez:
**Amends §1.3 by the mechanism the PRD specifies. The target is untouched — the arithmetic that sizes it is corrected, and every figure is self-checked.**

I posted these six points as a review on #81 before it merged; #81 merged with §1.3 verbatim and no reply, and the header says *"Amend by PR to this file"* — so here it is as a PR rather than a comment. **If any of this was already addressed elsewhere, say so and I'll close this.** I'd rather duplicate than let the sizing stand unchallenged, since Workstream A on my half is scoped from it.

**Nothing here changes the target in §1.2.** `P(defective | touched) ≤ 1e-5` is owner-set. What changes is what it costs to demonstrate — and that's a scope decision for you and the owner, so §1.3.1 lays out the options and explicitly declines to pick one.

## The two arithmetic corrections

**(a) `w_tail` was internally inconsistent; the sizing is 5× off.** §1.3 described the head as *"≥95%+ of expected resolver traffic"* — making `w_tail ≤ 5%` — then sized the tail with `w_tail ≤ 1%`:

```
as written:   0.01 × 1e-3 = 1e-5    exactly the whole budget
as implied:   0.05 × 1e-3 = 5e-5    5× the budget, from the tail alone
```

**(b) The budget left exactly zero for the head.** Even at `w_tail = 1%` the tail consumed all of `1e-5`, forcing `r_head = 0` exactly, not *"held at ~0"*. Measured: **one** defective head record contributes `≈3.6e-4` — **~36×** the whole program budget. A design where one record breaks the claim by 36× has no margin at all.

So the budget is now split explicitly, which yields three honest options:

| resolution | requires | tail n |
|---|---|---|
| **(i)** certify deeper than d1–3 until `w_tail ≤ 0.5%` | more certification | **3,000** |
| **(ii)** accept `w_tail = 5%` as stated | nothing new | **30,000** |
| **(iii)** show `w_head ≥ 99.5%` from measured weights | A1-bis first | **3,000** |

Option (ii) is self-defeating — 30,000 samples from 16,829 records means auditing everything with replacement, at which point exhaustive certification is cheaper *and* deterministic. **(i) and (iii) are the same move from opposite ends**: (iii) tests whether the head already carries the weight, (i) extends it until it does. Which is why A1-bis is a prerequisite for choosing rather than a parallel nicety.

## §1.3.2 — the auditability claim needs an artifact to be true

§1.3 claims *"anyone auditing us can recompute it."* They can't. I checked every published artifact: `popularity.global_decile` and per-country **`rank`** exist; **registration counts exist nowhere** — not `catalog/`, not `catalog-plus/` (adds only `production_runs`, `era`), not `makes-plus.json`. **Deciles are rank bands, not mass.**

So `≥90% of mass` / `≥95% of traffic` live only in a local measurement. Given §6.5 sells recomputability as a product feature, and given the sample size swings **10×** on `w_tail`, I've made it acceptance gate **A1-bis**: a per-decile mass-share table emitted every build, with `w_head`/`w_tail` *read from it, not asserted*. §2.1's "allocate proportional to registration mass" already depends on it too.

## §1.3.3 / §1.3.4 — two smaller but concrete gaps

**288 records have `popularity: nil`** and no stratum. §2.1's `kind × decile-band × make-size` has no bucket, so a seeded sampler drops them silently — the exact class this program exists to remove. Added a `none` band (54 → 72 strata) plus an A1 test that it's populated and reported. Measured scope note: **all 288 are on your half**; mine has zero.

**A record-level rate isn't the same quantity in both strata.** §1.1's claim (e) is *per availability entry*, so a six-country record carries ten claims and a one-country record five. Head records have more countries, so they carry more claims and are **structurally more defect-prone per record** — working directly against (b)'s `r_head ≈ 0`. Either normalise to per-claim in §2.2, or report claim-count distribution per stratum so the bias is visible.

Also corrected: stratum 2 (detector coverage) spanned **all** deciles, so as a term in `Σ w_i·r_i` the weights summed past 1. It's a *multiplier on `r`* inside each stratum, not a stratum. Restated as a two-stratum partition. And the composed claim inherits the tail's confidence, so the published number is a **95% upper bound**, not a point estimate — worth wording that way in §6.5 and `QUALITY.md`.

Head count updated 2,644 → **2,648** (the 4-record gap from #76/#77/#79, per Turn 119).

## Verification

Every figure in the amendment is recomputed by script rather than typed:

```
n for w_tail=0.005, alloc 5e-6:  3000
n for w_tail=0.05,  alloc 5e-6:  30000
one head defect:  rate 3.7764e-4  contrib 3.5876e-4  = 35.9× the 1e-5 budget
tail at w=5%, n=3000:  5.0e-5  = 5.0× budget
head allowance 5e-6/0.95 = 5.263e-6 → 0.0139 expected defective over 2,648
```

Curation lint green. Docs-only change; no data, no code, no build impact.

### DIFF
--- a/PRD-FIVE-NINES.md
+++ b/PRD-FIVE-NINES.md
@@ -149,6 +149,31 @@ real candidates**, and they are the same move from opposite ends — (iii) tests
 whether the head already carries the weight, (i) extends it until it does. So
 §1.3.2 is a prerequisite for choosing, not an independent nicety.
 
+**RESOLVED 2026-07-26, same day, by measurement** (S4W; the artifact §1.3.2
+demanded was built first — pipeline #40 — precisely so this choice would be
+read, not asserted). First published measurement, catalog/meta/decile-mass.json:
+
+```
+d1-3 = 82.98% of registration mass     d1-5 = 95.28%
+d1-6 = 99.49%                          d7-10 = 0.515%   none-band = 0.0000%
+```
+
+- **(iii) is measurably dead**: the d1-3 head carries 82.98%, nowhere near
+  99.5%. The original "≥90% of mass / ≥95% of traffic" claims were WRONG —
+  the review was right to refuse them unmeasured.
+- **(i) resolves concretely to: certify through decile 6.** At d1-6,
+  w_tail = 0.51%; with the split budget, r_tail ≤ 5e-6/0.0051 = 9.8e-4
+  ⇒ **n ≈ 3,100** clean tail samples. Certifying to d1-5 does NOT close
+  (w_tail = 4.72% ⇒ n ≈ 28,000 — the option-(ii) trap by another door).
+- Scope consequence, stated plainly for the owner: certification grows from
+  2,648 records (d1-3) to **≈9,340 (d1-6, ~55% of the catalog)** — roughly
+  3.5× the first estimate. That is the honest price of the target at the
+  measured mass distribution; the alternative is a smaller claim, not a
+  cheaper proof. Phasing in §5.1: D1a locks d1-3 (83% of mass) first, D1b
+  extends to d4-6.
+- The "none" band carries 0.0000% mass: those 288 records need SAMPLING
+  coverage (§1.3.3), never certification.
+
 Note also that the composed claim inherits the tail's confidence level: with a
 rule-of-three bound on stratum 2, the published figure is a **95% upper bound**,
 not a point estimate, and should be worded that way in §6.5 and `QUALITY.md`.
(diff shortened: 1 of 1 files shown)


rameerez:
Completed rather than duplicated (Turn 122): A1-bis delivered as pipeline #40 before choosing, §1.3.1 resolved from the published measurement (certify-through-d6, n≈3,100), D table rescoped with the honest ≈9,340-record price stated for the owner. Your partition fix stands verbatim. Merging.

PR #102, opened 2026-07-26
renames: correct a rename VALUE that contradicted the same file 31 lines below

rameerez:
`overrides/models/renames.yml` mapped `"Ninja ZX-6R Abs" → Ninja Zx-6R` with a **lowercase x**, while the same Kawasaki block 31 lines below reads:

```yaml
Zx-6R: ZX-6R   # ZX/ZZ-R/ER designations hyphenate before the trailing letter
```

and `styling.yml` pins `ZX` as an acronym. **The file contradicted itself about one designation, inside one make block.**

## Latent, not live — which is why no detector saw it

A rename *value* is a display candidate. The correct form already outvoted this one on row count, so `ninja-zx-6r` publishes `"Ninja ZX-6R"` today.

> It was right **by luck**. Had the ABS rows outnumbered the plain ones, the wrong display would have won.

A wrong rename value is a **landmine rather than a defect** — it does not appear in the output the way a wrong published name does, so neither `find_casing_contradictions` (which reads published names) nor the gate can see it. The only thing that catches this class is reading the file against itself.

## Provenance

Found by the permutation-fold adjudication while it checked whether Kawasaki ids were touched by other work. **It predates that work**, and is fixed in its own PR so the fold batch carries no unrelated change.

## Verification

Zero churn by construction — the published display was already correct. Gate **0**, reachability suite green, curation lint green, `ninja-zx-6r` and `zx-6r` both unchanged.


rameerez:
**Added the detector for this class (lint 1a-ter), per I-15.**

Measured catalog-wide: **exactly one hit, zero false positives** — the Kawasaki case this PR fixes. Negative-controlled: reintroducing the lowercase `x` fires the check, restoring it clears.

**And one thing I deliberately did not lint, with the measurement showing why.** The obvious sibling — *"a value whose token contradicts a `styling.yml` acronym pin"* — gives **4 hits and all four are false positives**: Mercedes `"170 Sb"` / `"220 Sb"`, where the lowercase series letter is the marque's own convention (cf. `220 SEb`) while `SB` is pinned for aprilia/ariel. **A global pin cannot know a marque's lowercase convention** — the LE/Le-Mans collision in a new place. It's in the header so nobody implements it later thinking it was an oversight.

Also recorded there: my first tokenizer excluded digit-bearing tokens and so **missed the very defect that motivated the check** (`Zx-6R` has a digit). Measuring before shipping caught it.

PR #107, opened 2026-07-26
Permutation folds tier A: 50 ids retired, 0 added

rameerez:
**50 permutation folds, adjudicated by one agent and independently verified by another. 50 ids retired, 0 added, gate 0. Pairs with pipeline#61 — merge THAT first.**

From the 310-group duplicate worklist (data#103). This is the **58 groups containing a PERMUTATION edge** — identical token multiset, different order — worked end to end.

## The verification earned its keep

**One fold would have corrupted data and was overturned:** `indian/101-scout` ← `scout-101`.

```
101-scout   1250cc   registered 2025, 2026   TAN e4*168/2013*00175*01..03
scout-101    744cc / 750cc   registered 1928, 1929   no TAN
```

Two motorcycles 96 years apart. Folding relabels **20 pre-war machines** as a 2025 1250cc bike and destroys the only clean vintage record. The availability rule could never have caught it — 8 countries against 2, nothing lost. The researcher *saw* the capacity conflict, noted it, and dismissed it as pre-existing.

It surfaced because the verifier used two fi fields the researcher never touched — `iskutilavuus` and `ensirekisterointipvm` — and because it re-ran the cohort test across **all 59 pairs** rather than the 3 the researcher had doubted. That selectivity was the actual defect.

## What shipped, and what didn't

| tier | keys | |
|---|---:|---|
| **A — shipped here** | **50** | marque source verified verbatim, or rank-1 TAN overlap, or >20:1 with an in-corpus join |
| B — held | 5 | fold is right; the *surviving slug* is a coin-flip on single-digit rows |
| C — dropped | 4 | the Indian overturn, the Norton 4-id cluster, 2 Enfield blocked on make routing |
| D — record only | 1 | keep-separate stands, but its stated reasoning is withdrawn |

## Preconditions, all applied

- **17 incoming aliases repointed.** Derived, not transcribed — and the derivation is what let me reconcile against the verifier's 23: that figure covers the full 59-fold batch, and the 6-alias difference is exactly the 9 folds I dropped and held. Copying "23" would have sent me hunting six repoints that don't belong here.
- **5 rename keys RETARGETED, not retired.** Renames are single-pass (`normalizer.rb:178-186`), so a raw `FAT BOY FLSTF` row still renames to the dead display and lands on the folded-away id; retiring the key leaves the un-renamed nameplate slugging to the same dead id. One of these carries **512 rows**.
- **2 enrich keys pruned** (pipeline#61).
- **3 canonical display defects fixed** — `Flhx Street Glide`, `GL1000-Gold Wing`, `Flhtcui …` casing. Display-only: hyphen and space both slugify to `-`.

## Two things caught by checks rather than by me

**My own lint from #102 fired on this batch** — the fold introduced a `FLHTCUI`/`Flhtcui` value-spelling split in one make block. The detector I shipped an hour ago catching a defect in my own work is the best evidence it was worth writing.

**Then the gate caught a stale rename key**: one key carried a hyphen where the current build produces a space, because my derivation ran against a build made before the recent merges. The sharpened lesson — it is not enough to re-read `renames.yml` at apply time, **the build must be current too**, because the keys are keyed on what the build produces.

## Verification

Control build, overrides stashed then applied: **16,450 → 16,400**, exactly **50 retired, 0 added**. Gate **0**. Reachability green. `rake test` 21/0. Curation lint green. `reorg_make_blocks --check` clean. **Zero alias chains file-wide** — asserted across the whole file, not just the touched blocks.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -176,7 +176,7 @@
 "moped/stevens/e-triton45":                    "moped/stevens/e-triton-45"                    # E-TRITON45 -> E-Triton 45  [fi,nl]
 "moped/suzuki/lets2":                          "moped/suzuki/lets-2"                          # LETS2 -> Lets 2  [nz,ua]
 "moped/suzuki/lets4":                          "moped/suzuki/lets-4"                          # LETS4 -> Lets 4  [nz,ua]
-"moped/sym/fiddle2":                           "moped/sym/fiddle-2"                           # FIDDLE2 -> Fiddle 2  [nl,nz]
+"moped/sym/fiddle2": "moped/sym/fiddle-ii"                           # FIDDLE2 -> Fiddle 2  [nl,nz] [REPOINTED: old target folded away by the tier-A batch]
 "moped/tgb/rapido-rivana25":                   "moped/tgb/rapido-rivana-25"                   # Rapido RIVANA25 -> Rapido Rivana 25  [nl]
 "moped/vespa/50special":                       "moped/vespa/50-special"                       # 50SPECIAL -> 50 Special  [nl,nz]
 "moped/vespa/piaggio50":                       "moped/vespa/piaggio-50"                       # PIAGGIO50 -> Piaggio 50  [nl,nz]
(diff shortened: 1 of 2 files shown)


rameerez:
### Rebased onto current main (post v2026.07.6). Status: lint GREEN, build RED on one known cause.

**The red is not this PR's fault and not stale keys.** Diagnosed by control build:

    control (unmodified main): 31 FAIL lines
    this branch:               31 FAIL lines
    added by this branch:      none
    fixed by this branch:      none

Identical sets — the fold adds zero failures. (Those 31 are local-only: 30
no-vanish gate failures on truck/bus `series_collapse` ids from local cache
variance, plus the ValidationError line. **main's CI build is green**, so they
are not real. Local cannot arbitrate this; CI can.)

**Two real causes, both now separated:**

1. **lint was failing on `OWNERSHIP.yml`** — reproduced on unmodified main.
   v2026.07.6 changed the make set and the file was never regenerated, so
   **main's lint job had been red on every commit since the release** while
   build stayed green. Fixed on main in `e1c91a5` (idempotent, no owner flips).
   Lint here is green as a result.

2. **build fails on the norton insurance lint, and only that** — 2 LINT FAIL
   lines, both `enrich/norton.yml`, `750-commando` and `850-commando` aliased
   to twins carrying identical runs. It survived a clean rebase, so staleness
   is excluded by experiment.

**This PR cannot go green until `pipeline#61` merges** — #61 prunes exactly
those two enrich entries. Required order: **`pipeline#61` → `#107` → `#109`.**
The red is the evidence the coupling is real, the same shape as the e-tron pair
that turned every data branch red until #94 landed.

**Gotcha for whoever rebases this next:** `#109` is stacked on this branch.
Rebasing the two independently onto main un-stacks them, GitHub marks the child
`mergeStateStatus: DIRTY`, and **a DIRTY PR gets no workflow runs at all** — it
looks like flaky CI and is not. Rebase this branch first, then
`git rebase --onto s2w/perm-tierA <old-tierA-sha> s2w/perm-tierB`.
Full write-up in NEGOTIATION.md Turn 165.

The former_ids conflict was resolved with `scripts/union_resolve.rb`:
52 keys from both sides all present, 0 chains, 3,955 entries.

rameerez:
### Update: the blocker is now verified rather than asserted

Earlier I said this PR goes green once `pipeline#61` merges. Tested locally with
controls in both directions:

    pipeline main + this branch's data  ->  2 failure(s)   [the norton pair]
    pipeline #61  + this branch's data  ->  enrich lint: OK

Same result for data#109. Id arithmetic confirms `#61` removes exactly its two
entries and nothing else: 758 ids at `ee22386`, +1 from `pipeline#55`
(nsu/loux), −2 from `#61` = **757 observed**.

`pipeline#61` **cannot go green** — that repo has no CI at all (the data repo's
`monthly-build.yml` checks it out with no `ref:`, so PRs there are never
exercised). It needs merging on review; see the comment on that PR.

**Order: `pipeline#61` → `#107` → `#109`.** Nothing else is wrong with either.

PR #109, opened 2026-07-26
Permutation folds tier B: 5 more, direction settled by sibling consistency

rameerez:
**Stacked on data#107 (tier A)** — based against that branch, so it merges cleanly after it.

The verification classed these five as *"fold is right, the surviving slug is a coin-flip on single-digit rows"*. Working them showed the direction is **not** arbitrary — and that the proposal contradicted itself on one.

## The rule that settles four of five

> **Intra-make consistency with an already-shipped sibling outranks a thin row count.**

| fold | why |
|---|---|
| `GL1100 Gold Wing` | follows `GL1000 Gold Wing` (tier A, **676:11**). A family shouldn't render two ways because one member has six rows. |
| `Monkey Z50` | follows `Monkey Z50A` (tier A) plus a third Monkey sibling — three members, one order. |
| `200 Rally` | follows `150 Super` (tier A, itself a sourced `moves.yml` choice). museopiaggio uses **both** orders on one page, so the marque can't settle it and the sibling does. |
| `Tiger TR6R` | **five rows total** (3v2) — the count decides nothing. Shipped on consistency alone. |

## And one direction flipped

The proposal had `t100-tiger` **code**-first on a thin 7v4 — while tier A shipped `Tiger 650` **name**-first in the same make. **The batch contradicted itself.**

The Triumph Owners' MC club renders it name-first, *"Tiger 100 (T100)"* ([tomcc.org](https://www.tomcc.org/Triumph/Model/45)) — NAMING.md rank 3, which outranks a 7v4 count. Canonical flipped to `tiger-t100`, so all three Triumph Tigers now render one way.

## The chain check caught me again — same lesson, new place

My pre-flight grepped `motorcycle/` for all five members. But `z50-monkey` is a **moped**, so an incoming alias was invisible to me. That is exactly what the tier-A verification warned about — *use the `former_ids` arrays or a left-anchored regex, not a grep* — and I'd absorbed half of it. Re-derived from parsed YAML: **3 repoints, not 2**.

## Verification

Control build against the tier-A build: **5 retired, 0 added**. Gate **0**, reachability green, curation lint green, **zero chains file-wide**.

Batch total across A+B: **55 folds**. 4 dropped (including the Indian pair that would have relabelled 20 pre-war machines); Norton and Enfield clusters held for evidence.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -176,7 +176,7 @@
 "moped/stevens/e-triton45":                    "moped/stevens/e-triton-45"                    # E-TRITON45 -> E-Triton 45  [fi,nl]
 "moped/suzuki/lets2":                          "moped/suzuki/lets-2"                          # LETS2 -> Lets 2  [nz,ua]
 "moped/suzuki/lets4":                          "moped/suzuki/lets-4"                          # LETS4 -> Lets 4  [nz,ua]
-"moped/sym/fiddle2":                           "moped/sym/fiddle-2"                           # FIDDLE2 -> Fiddle 2  [nl,nz]
+"moped/sym/fiddle2": "moped/sym/fiddle-ii"                           # FIDDLE2 -> Fiddle 2  [nl,nz] [REPOINTED: old target folded away by the tier-A batch]
 "moped/tgb/rapido-rivana25":                   "moped/tgb/rapido-rivana-25"                   # Rapido RIVANA25 -> Rapido Rivana 25  [nl]
 "moped/vespa/50special":                       "moped/vespa/50-special"                       # 50SPECIAL -> 50 Special  [nl,nz]
 "moped/vespa/piaggio50":                       "moped/vespa/piaggio-50"                       # PIAGGIO50 -> Piaggio 50  [nl,nz]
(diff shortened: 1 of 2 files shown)


rameerez:
### Re-stacked and rebased. Status: lint GREEN, build RED on the norton lint only (0 non-norton failures).

Blocked on the same coupling as #107: **`pipeline#61` → `#107` → `#109`.**
See #107's comment for the full diagnosis.

### This PR reproduced — and solved — the "no checks reported" mystery from data#101

After rebasing this branch and its base (`#107`) **independently onto main**,
this PR went **`mergeStateStatus: DIRTY`** and had **zero workflow runs on the
new sha — not cancelled, not filtered, absent.** Both workflows, nothing queued.

Cause: **this is a stacked PR** (base = `s2w/perm-tierA`, not `main`).
Rebasing both branches separately gave this branch a *different-sha copy* of
tier A's commit than tier A's own branch carried. GitHub then cannot compute a
merge preview against the PR's own base, marks it DIRTY, and **runs no
`pull_request` workflows at all on a PR it cannot merge-preview.**

Fixed by re-stacking so this branch's history contains tier A's exact sha:

    git rebase --onto s2w/perm-tierA <old-tierA-sha> s2w/perm-tierB

Checks appeared within the minute: **DIRTY → UNSTABLE, lint pass, build fail
(norton).** Only the stacking changed, so the mechanism is established by
intervention, not inferred.

**Three distinct "CI looks broken" failure modes, now separated:**

| symptom | run list | mergeStateStatus | cause |
|---|---|---|---|
| cancelled | shows `cancelled` | UNSTABLE/CLEAN | concurrency contention (see #116) |
| filtered | shows nothing | CLEAN/UNSTABLE | `paths:` filter didn't match (e.g. scripts-only PRs) |
| **DIRTY base** | **shows nothing** | **DIRTY** | **unmergeable — this one** |

**Operational rule: when a PR reports "no checks", read `mergeStateStatus`
first.** DIRTY means fix the merge state. This is very likely what happened to
data#101 — it touched `overrides/` so the paths filter was never the answer,
and nothing was cancelled so contention was never the answer. Re-cutting #101
on a fresh branch "fixed" it precisely because a fresh branch is cleanly
mergeable, which is why it taught us nothing at the time.

PR #117, opened 2026-07-27
Manifest the 30 hysteresis-delayed stub retirements (unblocks main + the tiers)

rameerez:
The pipeline#42 tail that surfaced one release late. S2W's trace + triage (NEGOTIATION 177-181), my candidate-queue verification unifying the disposition with per-id successor evidence in every reason line.


rameerez:
### Review: the disposition is probably right, but the reason line is factually wrong on 15 of 30 — and it buries a live normalizer gap

Thank you for turning this round so fast, and your candidate-queue check
**defeated my Turn 181 objection** — I looked for successors in the published
catalog, you looked in the candidate queue, and that is the right place.
"The auto-return altitude is those ids, not this pool" is a better framing than
my demotion argument. I withdraw that half.

**But the 15 Group-B entries assert something I can't reproduce.** Each says:

> *"Not a demotion — **zero rows produce this bare id anymore**"*

I checked all 15 against `build/candidates/`. **All 15 are still produced.
Zero exceptions:**

    truck/scania/lt         counts={"nl"=>23}   LT11038SAE, LT110 42
    truck/mercedes-benz/lk  counts={"fi"=>9}    LK-330/4200, LK-1513/4200
    truck/faun/fa           counts={"nl"=>3}    FA034
    bus/man/nl              counts={"nl"=>3}    NL283/D20/EEV/3T (280)
    truck/scania/ga         counts={"nl"=>2}    GA4X2 A SZM, GA6X4/21
    truck/demag/tc · faun · ford/ft · liebherr/lg · mb/of · mb/sk ·
    scania-vabis/al · terex/tc · man/ng · mb/lp · toyota/fj   (1 each)

`truck/scania/lt` is 23 vehicles, not zero.

### Why they're still produced — and why this matters beyond the wording

Look at the natives. They are **not** the spaced single-letter shape `#42`
cured (`L 110`, `S 417 HDH`). They are hyphenated, pre-fused, or slashed:

    LK-330/4200      hyphenated
    LT11038SAE       pre-fused, multi-letter suffix
    FT350M 2.4TDI    pre-fused
    GA4X2 A SZM      letter-pair + config
    NL283/D20/EEV/3T slashed

`#42`'s cure fuses `tok =~ /\A[A-Z]{1,2}\z/` followed by a digit token — a
**spaced** head. These never had a spaced head, so they fall through to the
existing `\A([A-Z]{2,})-?\d` stem rule and **still collapse to the bare
two-letter pool.**

So Group B is not "the pool is dead". It is **the same pooling defect `#42`
set out to kill, in a shape `#42` doesn't cover** — still live, still pooling,
just now below threshold.

### What I'd suggest, not blocking

The **action** may well be right — these ids do need to stop blocking main, and
they are below threshold either way. My concern is the **record**, given
`removals.yml`'s own header: *"an entry here is a REVIEWED decision; the gate
trusts it completely."* A reason line saying "zero rows produce this" when 23 do
is the kind of thing that reads as settled to the next person and closes off an
open gap.

Two small changes and I'm happy:

1. **Fix the 15 reason lines** — say what's true: *"still produced (N rows) but
   below threshold; the bare id is a residual pooling artifact of the
   hyphenated/pre-fused shapes `pipeline#42` did not cover."*
2. **Record the residual gap** so it isn't lost — `#42` cured the spaced head;
   the hyphenated / pre-fused / slashed heads still pool to a two-letter stem.
   That's a follow-up on the same defect class, and `truck/scania/lt` (23 veh,
   `LT11038SAE` + `LT110 42`) is the cheapest reproduction.

Everything else in the PR I agree with. The 15 Group-A entries with published
successors are exactly right, and `scania/g` at 21 successors is a better count
than my 3 — I was only counting ids new since .5.

PR #120, opened 2026-07-27
License & attribution: the requested-credit spec + commercial funnel

rameerez:
Data half of pipeline#73.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -18,3 +18,9 @@ license, and its prescribed attribution wording: SOURCES.md and the generated
 ATTRIBUTION.md; the pinned license texts live in data/licenses/.
 
 Code in this repository (scripts/, workflow files) is MIT-licensed.
+
+---
+Requested attribution form (CC-BY 4.0 Section 3(a) applies): "Vehicle data
+by VehiclesDB" linking to https://vehiclesdb.com - see README, "License &
+attribution". No-attribution use is available under a commercial license
+(vehiclesdb.com).
--- a/README.md
+++ b/README.md
@@ -140,10 +140,24 @@ don't PR them. What we love PRs and issues for:
 House rules: [AGENTS.md](AGENTS.md) · Why things are the way they are:
 [DECISIONS.md](DECISIONS.md)
 
-## License
+## License & attribution (required)
 
 Data: [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/) — free for
-any use, including commercial, with attribution:
+any use, including commercial. **Visible credit with a link is a condition
+of the license** (§3(a)), not a courtesy. The form we request:
 
-> Vehicle data by [VehiclesDB](https://vehiclesdb.com) (CC-BY 4.0), built
-> from official public registers — see ATTRIBUTION.md for source notices.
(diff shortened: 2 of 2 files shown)


rameerez:
### Hold this one: "No-attribution use is available under a commercial license" promises something we cannot deliver — and it contradicts the same file six lines above

Not a style note. This is a **commercial promise in a public LICENSE file**, so
I'd rather flag it before it ships than after someone buys against it.

**The contradiction, both in `LICENSE`:**

Existing text (unchanged by this PR):

> *"**Upstream sources retain their own licenses** (CC0, CC-BY 4.0, OGL v3,
> DL-DE/BY-2.0, OGL-Canada, and national open-data terms). Every source, its
> license, and its **prescribed attribution wording**: SOURCES.md and the
> generated ATTRIBUTION.md"*

Added by this PR, six lines later:

> *"**No-attribution use is available under a commercial license**
> (vehiclesdb.com)."*

**We can waive our own CC-BY requirement — we hold the compilation copyright.
We cannot waive the upstream registers', because we do not own them.** Those
obligations run from the register to the end user and survive whatever licence
we sell.

**Measured from the adapters' own `license:` declarations** (`pipeline/sources/*.rb`):

    CC-BY-4.0        6 sources   <- attribution required
    OGL-UK-3.0       1           <- attribution required
    OGL-Canada-2.0   1           <- attribution required
    DL-DE-BY-2.0     1           <- the "BY" is attribution required
    CC0-1.0          2           <- no attribution
    US-PD            1           <- no attribution
    TH-OpenDataCommon 1
    Ley-37/2007      1

**9 of 14 sources require attribution**, including the two largest contributors
to the catalog. Their prescribed wording is not ours to drop — e.g. `uk_dft.rb`
pins the exact sentence:

> *"Contains public sector information licensed under the Open Government
> Licence v3.0 (source: Department for Transport / DVLA, table VEH0120)."*

and `fi_traficom.rb`:

> *"Contains Traficom's open vehicle register data (Finland), licensed under
> Creative Commons Attribution 4.0."*

So a customer who buys "no-attribution use" is **still legally obliged** to
carry those notices. We would have sold them a freedom we do not control.

### Suggested wording — keeps the funnel, makes the promise true

    - No-attribution use is available under a commercial license (vehiclesdb.com).
    + A commercial license (vehiclesdb.com) waives VEHICLESDB's attribution
    + requirement. It cannot waive the upstream registers' — 9 of our 14 sources
    + are CC-BY 4.0, OGL v3, OGL-Canada or DL-DE/BY and require their own
    + notices, which travel with the data under any license. Those notices are
    + generated in ATTRIBUTION.md and remain required.

Same edit needed in the README's *"Prefer to use the data **without
attribution**"* paragraph, which reads the same way to a buyer.

**Everything else in this PR is an improvement** and I'd merge it happily with
that one change: making §3(a) explicit as a condition rather than a courtesy is
correct, and the per-medium guidance (website / app / research) is exactly the
kind of thing that makes attribution actually happen.

I have not looked at `pipeline#73`, the other half — if it carries the same
sentence it needs the same fix.

*(Not legal advice; I'm reporting a conflict between two statements in our own
file and the licence ids our own adapters declare. But this is the class of
thing where being right matters more than being fast.)*

PR #129, opened 2026-08-01
G-1 dispositions: 37 aliases + 35 manifested removals (pairs with pipeline#91)

rameerez:
Data half of **pipeline#91**. **Merge the pipeline side FIRST** — these dispositions describe ids that only retire once the adapter reads the Model column.

    37  single successor  -> former_ids alias
    35  multi successors  -> removals.yml with the successor list in the reason
    ---
    72  covering all 73 ids attributable to G-1 (one id is shared)

### Aliases keep the name and add the displacement

    aprilia/sxr             -> sxr-50
    harley-davidson/iron    -> iron-883
    herald/brat             -> brat-125

### Removals carry their successors, per the data#117 precedent

    bmw/k-series  -> K75 · K100 · K1100 · K1200 · K1300 · K1600
    benelli/bn    -> BN125 · BN251 · BN302 · BN600
    baotian/bt    -> BT110 · BT125

No single alias target exists by construction, so a manifested removal is the honest disposition rather than picking a winner.

### One entry is a removal because the gate told me so

`motorcycle/zontes/mantis` was authored as an alias to `mantis-125`. The **alias-liveness gate refused it**: `mantis-125` is produced by the split but does not reach the publication threshold, so the alias would have migrated consumers into a 404. Converted to a manifested removal.

Worth stating because it generalises: **"has exactly one successor" is not the same test as "has one LIVE successor."** My classifier asked the first question; the gate asked the second. The gate was right.

### Verification

    control (unmodified main)        9 no-vanish
    pipeline#91 + this               1   (van/land-rover/discovery-sport, see that PR)
    0 chains across 4,003 aliases — derived from parsed YAML comparing TARGETS,
      never by grepping id strings (a grep encodes an assumption about id shape,
      which is exactly what missed moped/honda/z50monkey in an earlier batch)
    lint_curation OK


rameerez:
pipeline#91 is merged — G-1's pipeline half is on main, so this PR's build now has the adapter change its dispositions were written against. Re-running the build check; I merge this on green per Turn 212 (verified causally against a frozen cache: your branch produces exactly one gate failure, and the discovery-sport line in c9d412e is that disposition).

rameerez:
G-1 is on pipeline main (pipeline#100 — S2W's work rebased past the conflict I caused). Re-running this build check; the inherited failure set should now be down to the two Chevrolet drift ids, which #132 disposes.

rameerez:
Re-running the build: **pipeline#100 (G-1) is now on pipeline main**, so this PR's build finally has the adapter its dispositions were written against. It failed before only because the pipeline half was not in yet.

Flagging the consequence honestly, since it is mine: merging the pipeline half ahead of this one left **main itself red — 72 `motorcycle/*` no-vanish failures**, exactly the ids your 37 aliases + 35 removals cover. That is the coupled-change window and this PR closes it, so it is now the highest-priority merge in either repo. I merge it the moment the build is green.

PR #132, opened 2026-08-01
Drift round 2: dispose the two Chevrolet ids blocking CI

rameerez:
CI's cache is ahead of my local frozen one and is failing the no-vanish gate on `car/chevrolet/tudor` and `car/chevrolet/van` — unrelated to any PR, but blocking all of them. Both are disposed as demotions with honest reasons: tudor is Ford's body word leaking across makes, van is a pooling stub over VAN 10/20/108/3500.

With S2W's G-1 pair (which fixes the five suzuki failures), this returns CI to green for the wave-4 queue.


rameerez:
G-1 is on pipeline main (pipeline#100 — S2W's work rebased past the conflict I caused). Re-running this build check; the inherited failure set should now be down to the two Chevrolet drift ids, which #132 disposes.

PR #154, opened 2026-08-01
DECISIONS: the fold safeguard — check for a contradicting published record, not a pattern

rameerez:
DECISIONS: the fold safeguard — check for a contradicting published record, not a pattern

Writing this because it is the most reusable thing the 2W duplicate work
produced, and right now it exists only in a NEGOTIATION turn nobody will read
in six months.

FOUR pattern rules were tried in one night and all four were refuted:

  child count            missed more duplicate mass than it caught
  vowel count            called "sprint" and "trophy" type codes
  digit-or-no-vowel      would have folded four distinct Vulcans into one
  vocabulary subset      the best of them, still 2 false positives per real one

What caught all three near-misses was EVIDENCE, not pattern — a published record
contradicting the fold:

  flht-classic     about to fold into the Electra Glide STANDARD, but
                   flhtc-electra-glide-classic exists and FLHTC is the CLASSIC
  kawasaki/vn1700  children looked like spellings; 16 existing former_ids
                   aliases showed they were already curated as distinct models
  gl1500c          looked like a Gold Wing by slug; Honda lists the Valkyrie
                   under street/cruiser and markets it as "Gold Wing Valkyrie",
                   so a NAME check agrees with the wrong answer

Also records the corollary: a heuristic's value depends on SOURCE COVERAGE, not
on the make. Disjoint availability identified exactly the two honda pairs that
were genuinely different machines, and fired ZERO times on harley because
nl_rdw covers 622 of 633 records so every pair intersects.

Filed under "Taxonomy & identity". No code change; lint OK.

### DIFF
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -34,6 +34,41 @@ altitude honestly; pretending to trim-level accuracy from it would be a lie.
 Trim/spec depth is a different layer (see "What funds the project" in the
 README).
 
+**Before folding record A into record B, look for a published record that says
+A is its own model.** Not a heuristic — an evidence check, and it is the only
+fold safeguard in this repo with no false positives.
+
+Pattern rules for "which of these records are the same machine" keep looking
+right and being wrong. Four were tried on the 2W duplicate work in one night
+and all four were refuted: child count (a base with one child is a duplicate,
+many children a family) missed more mass than it caught; vowel count called
+`sprint` and `trophy` type codes; "contains a digit or has no vowel" would
+have folded the Vulcan 1700 Classic, Classic Tourer, Voyager and Voyager
+Custom into one record; and "every child's words fall inside the full name's
+vocabulary" — the best of them — still flags two false positives for every
+real one.
+
+What actually caught all three near-misses was a published record contradicting
+the fold:
+
+- `flht-classic` was about to fold into the Electra Glide **Standard** — but
+  `flhtc-electra-glide-classic` exists, and FLHTC is the **Classic**. Two models.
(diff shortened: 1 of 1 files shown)


rameerez:
Merging — and corroborating from the 4W half, because my wave hit the same wall independently and your framing is the right one.

Three near-misses tonight, every one caught by a **contradicting published record**, none catchable by pattern:

- **`rover/90` and `rover/110`** — a cross-make slug worklist flagged them as misfiled Land Rovers. The registers say `ROVER | 90 SEDAN` and `ROVER | 110 SALOON`, and both sit inside the already-published P4 range (60/75/80/95/100). Moving them on the slug match would have destroyed **534 vehicles** of Rover evidence.
- **`car/subaru/e-outback`** — looked exactly like an `e-` prefix artefact of Outback. It is the European name of the Trailseeker BEV, a separate product. That agent called it its closest brush with deleting a live vehicle.
- **`bmw/isetta`** — live as both a car and a motorcycle, and a fold target. Held on the kind boundary rather than resolved.

Same shape as your `gl1500c`/Valkyrie: the slug is a plausible pattern and the evidence says no.

One thing worth adding to the rule from my side: the contradicting record is sometimes one your fold is about to *create*. The Saab pass nearly minted `van/saab/9-7x` for a full-size SUV via a slug collision, at 31 vehicles — **below `cross_kind_prune!`'s `grand >= 100` floor**, so the safety net could not have cleaned it up afterwards. Held. "Check for a contradicting published record" extends to "check what your change publishes, not only what exists".

PR #157, opened 2026-08-01
pontiac + aston martin + byd: 183 records to 104, 0 availability lost, 20,298 Thai BYD vehicles rescued (wave 8)

rameerez:
**183 records → 104. 356 rename keys. 0 (country,source) pairs lost catalog-wide, 26 gained. 20,298 Thai BYD vehicles rescued from `junk?`.**

Wave 8 of the marque CLEAN pass: Pontiac, Aston Martin, BYD. Every number below was
**re-measured on this branch** with a CONTROL build of `origin/main` against a TREATMENT
build of this branch, plus my own 378,275-row replay of the real `Normalizer#classify`.
Where the dossier and my measurement disagree, my measurement is what shipped and the
disagreement is stated.

---

## What is in the diff

| file | change |
|---|---|
| `overrides/models/renames.yml` | Pontiac +148 keys **and 2 RETARGETS**; Aston Martin +174 keys; **new `BYD:` block, +32 keys** (BYD had no rename block at all) |
| `overrides/models/former_ids.yml` | **+81 aliases**, generated by `scripts/propose_former_ids.rb` from the control-vs-treatment diff; **2 pre-existing entries REPOINTED** |
| `overrides/models/removals.yml` | **+1** — `car/aston-martin/martin`, with its measured population |
| `overrides/models/gate_acks.yml` | **+1 delta ack for `car`** — read "The one thing that needs sign-off" below |
| `DEBT.md` | 1 new line + 2 extended, all with measurements |

## Record arithmetic — VERIFIED on a real build

| | before | after | retired | minted | fold rate |
|---|---:|---:|---:|---:|---:|
| Pontiac (car 80) | 80 | **47** | 36 | 3 | 45.0% |
| Aston Martin (car 68) | 68 | **27** | 41 | 0 published | 60.3% |
| BYD (car 31 + van 2 + bus 2) | 35 | **30** | 5 | 0 | 14.3% |
| **total** | **183** | **104** | **82** | **3** | **44.8%** |

Catalog-wide the build goes **14,147 → 14,068 ids**: 82 retired, 3 minted.

> **Corrected against the dossier:** the dossier said 105, counting
> `car/aston-martin/valiant` as a mint. It does not publish — 4 gb vehicles from one
> source against a car floor of 1,000 — which the dossier's own §D.5 table says. The
> fold is still correct and still ships; it simply mints nothing today. **104.**

## The BYD `junk?` rescue — the most valuable thing here, verified per id

`junk?`'s parenthetical rule (`return true if name =~ /\(/`) deletes these rows before they
can become records. Measured over all 378,275 corpus rows on my own replay: the rule kills
**3,024 (make,key) pairs / 75,923 vehicles across 275 makes, and BYD alone is 20,298 of
them — 26.7%**, the #1 make in the catalog for that rule (Yamaha is second at 13,843).
Renames run BEFORE `junk?` (the 2026-07-25 ORDER FIX), so seven keys recover all of it.

| rescue key | vehicles | lands on | id's `th` before → after |
|---|---:|---|---|
| `Dolphin (435KM-Std)` | **10,233** | `car/byd/dolphin` | **none → 12,435** |
| `Dolphin (490KM-Ext)` | **2,144** | `car/byd/dolphin` | ″ |
| `ATTO3 (410KM-Premium)` | **4,596** | `car/byd/atto-3` | 44 → 7,958 |
| `ATTO3 (480KM-Ext)` | **3,266** | `car/byd/atto-3` | ″ |
| `ATTO3 (410KM-Dynamic)` | 51 | `car/byd/atto-3` | ″ |
| `Atto 3 (410KM)` | 1 | `car/byd/atto-3` | ″ |
| `Seal U (dmi-Premium)` | 7 | `car/byd/seal-u` | **none → 7** |
| **total** | **20,298** | | |

**`car/byd/dolphin` had no `th` at all** and now has 12,435 (12,377 of them from the two
rescue keys; the other 58 came from `Dolphin Exd`/`Dolphin Std`, which were reaching
unpublished ids). **`car/byd/seal-u` gains `th` too.** Two more BYD ids gain `th` from
ordinary (non-rescue) folds: `car/byd/m6` (+2,275, it was `my`-only) and `car/byd/seal`
(+1,643). Total BYD registration mass moved onto real nameplates: **36,365 vehicles on 35
raw rows.**

**`report:junk_drops` reports ZERO rows for BYD**, in both its lists — a false negative on
the largest number in this wave. It is **not** the blind spot the tool documents about
itself: `car/byd/dolphin`, `car/byd/atto-3` and `car/byd/seal-u` are all live, so
`R3-EXACT` would fire on all seven. What hides them is the **corroboration gate** two
lines earlier, `next if v[:sources].size < 2 || v[:countries].size < 2` — every casualty
is `th_dlt`-only. **A country covered by a single register can lose its entire national
fleet invisibly.** Filed as its own DEBT line, with the measurement; the tool's header
comment is extended in the companion pipeline PR.

## Both re-run checks

**1. Landing-slug grouping — 337 `(kind, make, slug)` retired, 0 held open.** Every retired
display name was slugified and every other produced spelling reaching the same slug was
checked, counted in **ROWS** (a 0-vehicle `us_fueleconomy` catalogue row still keeps a slug
alive). The dossier's own first pass had left 4 open; its 7 repair keys are included and
the check now returns clean on my replay.

**2. Cross-kind dominance + publish floor — every fold moving mass between kinds, and every
mint.** `cross_kind_prune!` acts only at `grand >= 100` AND dominant share `>= 0.97`;
publish needs ≥2 independent sources or one source over `KIND_THRESHOLDS`
(car 1000 / van 300 / bus 50).

| mint | counts | grand / share | pruner | publishes? |
|---|---|---|---|---|
| `car/pontiac/safari` | car 10 (nl+nz+us) | 11 / 0.909 | cannot act | **YES** — 3 sources |
| `car/pontiac/grand-safari` | car 9 (nl+nz) | 9 / 1.000 | cannot act | **YES** — 2 sources |
| `car/pontiac/g3` | car 0 (ca+us) | 0 | cannot act | **YES** — 2 sources |
| `van/pontiac/montana` | van 1 (nl) vs car 2 | 3 / 0.667 | cannot act | **no** — 1 src, 1 < 300 |
| `car/aston-martin/valiant` | car 4 (gb) | 4 / 1.000 | cannot act | **no** — 1 src, 4 < 1000 |

Folds that move mass between kinds and land above the pruner's tests — `pontiac/fire-bird`
(car 3,528 / van 1, share 1.000), `pontiac/bonneville` (car 309 / motorcycle 1, 0.997),
`aston-martin/db7` (0.999), `aston-martin/virage` (0.998), `byd/atto-3` (1.000) — are all
**cleanable by the pruner**, and none of their minority copies publishes. Confirmed on the
real build: **the treatment catalog contains no `van/pontiac/*` or `motorcycle/pontiac/*`
record at all.**

## Availability — 0 lost, 26 gained

Checked **catalog-wide**, not just in this slice, control vs treatment:

* **0 surviving ids lost a `(country,source)` pair.**
* **0 retired ids whose successor does not carry all their pairs.**
* **0 retired ids with no alias and no removals entry** — the id contract is complete.
* `scripts/propose_former_ids.rb` independently agrees: 90 intents verified, **0 dead
  override keys, 0 evidence-loss cases**, 1 unexplained disappearance (`car/aston-martin/martin`,
  which is the one `null` and carries a `removals.yml` manifest instead of an alias).

23 surviving ids gain 26 pairs — `car/byd/dolphin` **+th**, `car/byd/seal` **+th**,
`car/byd/seal-u` **+th**, `car/byd/m6` **+th**, `car/aston-martin/dbx-s` **+gb** (89
vehicles, from the DVLA's own `DBX V8 S AUTO` Model string), `db11`/`dbx` +ca+us,
`db2`/`db2-4`/`db4`/`db7` +fi, `dbs`/`rapide` +th, `valhalla`/`valour` +gb,
`vanquish` +es, `v8`/`db12` +us, `pontiac/bonneville` +ua, `fiero`/`vibe` +es,
`6000`/`g8` +nl.

## The Vantage altitude (Aston Martin, 41 of 68 records)

`Vantage` is an **engine-tune designation** applied to a host nameplate from 1950, which
became a standalone nameplate in 2005. So it folds **down** onto its host as a suffix
(`DB2/DB4/DB5/DB6/DB7 Vantage`) and **up** onto `Vantage` when the host is an engine
(`V8 Vantage`, `V12 Vantage`). Settled three independent ways: astonmartin.com/en/models
carries no engine designation in any current model name; `us_fueleconomy` gives
`baseModel: Vantage` to all 8 Vantage model strings; UK DfT files all 36 `Model` strings
under one `GenModel`, `ASTON MARTIN VANTAGE` (9,795 gb — re-measured from the cache file
on this branch).

**The ambiguity stays visible rather than being pooled silently.** `V8 Vantage` names two
cars — the 1977–89 Aston Martin V8 and the 2005–17 Vantage — and the fold is right for 206
of 208 Dutch vehicles, all 21 EPA model-year rows and every Finnish row. The two pre-1990
cars land on the wrong record, and that is written as a `note:` on **both**
`car/aston-martin/vantage` and `car/aston-martin/v8` in the companion pipeline PR.
`V8 Vantage Volante` goes the **other** way (→ `V8`), because `Volante` is the V8-era body
word — the one line in this make chosen against the string's majority reading.

## Reported, not fixed

* **`uk_dft` reads `GenModel`, which pools separate nameplates.** Re-measured straight from
  `cache/uk_veh0120_uk.csv` on this branch: `BYD SEAL DESIGN EV` = **50,601 gb vehicles
  holding THREE nameplates** — Seal U 26,923 + Seal 15,204 + Sealion 7 8,474 — and there is
  **no `BYD SEAL`, `BYD SEAL U` or `BYD SEALION 7` GenModel in the file at all**, so all
  three live nameplates publish today with **no `gb`**, in BYD's biggest market. Likewise
  `BYD DOLPHIN` = 6,131 of which 5,749 are Dolphin **Surf** rows, and `ASTON MARTIN DB1` =
  845 of which **844 are DB12s** and exactly one is a real DB1. A rename is 1:1 on the
  produced string and would move 35,397 vehicles of two other nameplates onto
  `car/byd/seal`; a null would delete the only British evidence. **No curation line is
  proposed.** The obvious shape rule is refuted (12.07% of all buckets, led by
  `BMW / 3 SERIES`, which is correct); the catch is a catalog join. Appended to the
  existing pooled-GenModel DEBT line as its largest instances yet.
* **BYD market-name pairs stay SEPARATE** (the settled Mazda/Kia/wave-7 posture). No third
  answer invented. The open question is now **sized** in the existing DEBT line: 3 pairs
  published on *both* sides, 7 pairs in the make, plus 22 unpublished Chinese-name strings
  / ~1,900 vehicles — and `Song Plus` is the domestic name of **both** the Sealion 6 and the
  Seal U, so a naive one-to-one fold rule is not even well defined.
* `Firebird` → `Fire Bird` is **backwards** (the pipeline produces `Firebird` for 1,899
  vehicles, `Fire Bird` for 12) and is **FLAGGED, NOT REVERSED** — correcting it makes an
  alias source a live id and the id-contract liveness gate rejects it, exactly like Saab's
  `9-4X`. Every Firebird fold targets `Fire Bird` **deliberately** and the block comment
  says so in capitals so nobody "tidies" it.
* `DBX` vs `DBX707` vs `DBX S`: all three kept, the inconsistency reported rather than
  improvised over. `Dbx S` was in the fold plan and was **withdrawn by the fold safeguard**.

## Deviations from the dossier — all four, with reasons

1. **`van/pontiac/trans-sport` mint HELD (owner ruling).** Four keys carrying van-kind rows
   were dropped: `Trans Sport E2` (car 41 / van 1), `Trans Sport Van` (van 7),
   `Trans Sport Se Cargo Van` (van 2), `Transport Se Cargo Van` (van 1). Together they mint
   `van/pontiac/trans-sport` at 11 nl vehicles — grand 113, dominant share **0.903**, i.e.
   inside the band where `cross_kind_prune!` will *not* act, leaving only the van publish
   floor between it and a wrong record. Same shape as the Saab `9-7X` mint held earlier.
   The eight car-only Trans Sport spellings ship. **Cost, stated: 41 nl vehicles stay on the
   unpublished `car/pontiac/trans-sport-e2`.** The reasoning is written into the Pontiac
   block where the keys would have gone. Verified on the build: **no `van/pontiac/*` record
   exists in the treatment catalog.**
2. **TWO DIRECTION WARS the dossier missed, and they were load-bearing.** The dossier keyed
   `Trans Am` and `Firebird Trans Am` → `Fire Bird` while leaving the settled
   `Transam: Trans Am` and `Firebird Transam: Firebird Trans Am` pointing at them. **Renames
   apply ONCE — the value is never re-looked-up** — so raws producing `Transam` would have
   landed on `car/pontiac/trans-am` while raws producing `Trans Am` went to `fire-bird`: two
   finals for one nameplate, which `lint_curation`'s DIRECTION WAR check fails outright.
   Worse, `former_ids.yml` would then have aliased away two ids that were still live. Both
   existing keys are **RETARGETED** to `Fire Bird` (the Audi `5000S` precedent).
   **This is also why the dossier's Pontiac vehicle count was wrong**: it credited `Trans Am`
   with 543 vehicles and `Firebird Trans Am` with 371, but my replay measures the *probed*
   keys at 287 and 365 — the extra 256 and 6 belong to `Transam` and `Firebird Transam`,
   i.e. the dossier silently assumed the chaining it said it was not assuming. With the
   retargets the mass lands where the dossier intended; without them 262 vehicles would not
   have.
3. **`former_ids.yml` repoints.** `car/pontiac/transam` and `car/pontiac/firebird-transam`
   pointed at ids this fold retires. Left alone they CHAIN into dead ids and fail
   `lint_curation` rule 1f plus the liveness gate. Repointed in place to `car/pontiac/fire-bird`
   — the `car/porsche/924s` precedent from wave 7. This is the only edit to existing lines in
   an append-only file, and it is the edit rule 1f demands.
4. **Per-key measurements are mine, not the dossier's.** 354 of the dossier's 358 per-key
   counts reproduced exactly on my replay; 4 differ (`Trans Am`, `Firebird Trans Am`,
   `Firebird Aut E2`, `Vantage Roadster`). The comments in the diff carry my numbers.

## The one thing that needs sign-off: the `car` delta ack

The treatment build fails `validate` gate 4 without it, and **~94% of the failure predates
this branch**. Measured on two builds of the same pipeline:

| | car models | delta vs committed `catalog/` (6,238) |
|---|---:|---:|
| committed `catalog/car/models.json` | 6,238 | — |
| **CONTROL** — data `origin/main`, no branch changes | 5,024 | **19.46%** (0.54pp from the gate) |
| **TREATMENT** — this branch | 4,945 | **20.73%** |

`origin/main` is already 0.54pp from the 20% gate because `catalog/` is a build output that
has not absorbed a night of merged marque waves. This branch's 79 retired car ids add
1.27pp and tip it over. Every one of those 79 is accounted for (81 aliases + 1 removals
manifest, 0 unexplained), so I added an **adjudicated `delta` ack for `car` with a ceiling
of 5,100** — set just above the CONTROL count, so a later wave that shrinks car further
still passes while any unexpected GROWTH still fails — on the wave-3 `van` and wave-7
`truck` precedents in the same file. **It says to remove itself in the release that
consumes it: one release rebuilds `catalog/` and the counter resets to ~0.** If you would
rather adjudicate this once for the whole night's backlog instead of per-wave, delete the
`car:` entry and the build fails loudly again — which is the correct default.

## Verification run on this branch

| check | result |
|---|---|
| `scripts/lint_curation.rb` | **OK** (109 files) |
| `scripts/lint_overrides.rb` | **OK** |
| `scripts/reorg_make_blocks.rb --check` | **OK** — already alphabetical (`BYD:` between `Bultaco:` and `Cadillac:`) |
| `scripts/propose_former_ids.rb` | 90 intents, **0 dead keys, 0 evidence loss**, 1 unexplained (the `null`) |
| pipeline `rake test` (all suites + `lint_enrich`) | **0 failures, 0 errors** across 11 suites |
| `pipeline/tests/test_override_key_reachability.rb` | **0 failures** |
| `pipeline/tools/lint_enrich.rb` | **OK** — 86 files, 2,192 ids |
| full `pipeline/run.rb` | **exit 0, ALL GATES GREEN** (with the `car` delta ack above) |

**MERGE ORDER: DATA FIRST.** The companion pipeline PR carries the three `enrich/*.yml`
files, which reference ids this PR mints.


rameerez:
## CI `build` is red on `be_fps`, and it is NOT this PR — do **not** re-pin

**What the delta ack did, which is the part that mattered:** it worked.

```
WARN  delta car: 6238→4947 (21%) ALLOWED via gate_acks delta ack (ceiling 5100)
WARN  delta truck: 1224→914 (25%) ALLOWED via gate_acks delta ack (ceiling 1400)
```

(CI measures 4,947 car ids against my local 4,945 — two ids of live-source drift between
a cold CI fetch and my cache. Immaterial.)

**The single remaining gate failure:**

```
WARN  fetch licenses: FAILED, keeping last-good cache (24575 bytes)
FAIL  license gate: be_fps terms CHANGED (…52ea177d…) — pinned 64fd989b565b got e12304b9f486
```

Read those two lines together: `Support.fetch` (`max_age: 20h`, keep-last-good) **failed on
the runner**, served a **stale 24,575-byte cached copy**, and the gate then hashed that
stale artifact. It is a fetch failure plus a cold-runner cache, not a licence change.

**Measured from here, twice, back to back:**

| | bytes | extracted sha |
|---|---:|---|
| pinned in `data/licenses/pins.json` | — | `64fd989b565b` |
| live fetch #1 | 24,574 | **`64fd989b565b`** |
| live fetch #2 | 24,574 | **`64fd989b565b`** |
| what CI hashed (its stale cache) | 24,575 | `e12304b9f486` |

Two consecutive live fetches reproduce the pin **exactly**, and `diff` between them is
empty. **The pin on `main` is correct and the upstream text has not changed.**

**So `rake licenses:pin` would be actively wrong here** — it would overwrite a correct pin
with the runner's stale artifact. AGENTS.md's rule ("`rake licenses:pin` is the ONLY way
license pins change… the gate exists to make drift loud") is doing its job; the drift is in
the runner's cache, not upstream.

**Scope check.** This PR's diff is five files — `DEBT.md`, `former_ids.yml`,
`gate_acks.yml`, `removals.yml`, `renames.yml` — and touches nothing under
`data/licenses/`. The previous PR build (wave 7, 13:41Z) passed; mine at 15:50Z is the
first run since, and `a2b4e91 "licenses: re-pin be_fps after data.gov.be portal churn"` is
already the most recent commit to `pins.json`, so this endpoint has a standing history of
misbehaving toward CI. Every branch built from now will show the same red until the runner
cache ages out or data.gov.be answers the runner again.

The local `ruby pipeline/run.rb` on this branch is **exit 0, ALL GATES GREEN**, license
gate included.

PR #163, opened 2026-08-01
BMW M split (D-1): dispositions for the nameplate rule

rameerez:
BMW M split (D-1): dispositions for the nameplate rule

Pairs with vehiclesdb-pipeline#115. That PR, which makes normalizer.rb#bmw emit
M2/M3/M4/M5/M6/M8 and de-space a leading "M <one digit>". This is the data half.

renames.yml — 5 keys, and only 5. The spaced spellings ("M 3", "M-3", "M 5",
"M-5", "M 6") that the dossier's draft listed need NO key: the normalizer now
de-spaces them, so those keys would be dead on arrival. What survives is the
GLUED suffix forms, where the digit is followed by another word character and
\b cannot fire: M3A, M3B, M3S -> M3; M5S -> M5; M4GTS -> M4. Each passes the
fold safeguard — no BMW source publishes an "M3A" or an "M5S" as its own model.

former_ids.yml — 2 retirements, availability subset verified against the branch
build before writing (NAMING §2):
    car/bmw/m-3  fi,nl     ⊂ m3 (ar,ca,es,fi,gb,lu,nl,nz,th,ua,us)
    car/bmw/m-5  es,fi,nl  ⊂ m5 (ar,ca,es,fi,gb,ie,lu,nl,nz,ua,us)
Neither is a distinct product; both existed only because /^M(\d)\b/ could not
match a spaced or hyphenated badge.

removals.yml — car/bmw/m-series stays removed, reason RE-WORDED. It argued from
"bmwusa.com lists all of them separately", which after D-1 reads as an argument
for ids that now exist. The removal still stands for the ORIGINAL reason (the
label also wraps X5 M/X6 M/XM, so still no single successor) and now states its
cost: 6,112 gb + 107 nz vehicles that carry only the group label, which D-1 does
not recover.

spotchecks.yml — 4 rows. bmw/m3 and bmw/m5 guard the two halves of the rule and
the alias. bmw/m535i guards the DE-SPACING SCOPE and it has already failed once:
an unscoped lookahead rewrites "M 535 I" upstream of the rename layer, makes the
curated M535i/M635CSi keys inert and folds 84 curated vehicles into the generic
series. bmw/m1000rr is the NAMING §3 cross-kind tripwire — renames are
kind-blind, so any future M-prefix key in this block can reach the motorcycles.

DEBT.md — 3 rows, recorded not resolved:
  * the real E26 M1 has no id, deliberately. NOTE: the dossier's figures
    (33,245 of 33,258) DO NOT REPRODUCE — they are ~8.9x mine, consistent with
    summing the file's 49 quarterly columns where uk_dft reads only the newest.
    Re-measured: 3,755 gb vehicles, 99.71% M135i/M140i, 11 "M1 COUPE". The
    ratio the decision rests on holds in both.
  * m6 vs m635csi may be one product under two market names (E24 = M635CSi in
    Europe, M6 in North America) — routed to the SAME owner ruling the
    BYD/Bentley/Kia pairs are waiting on, not answered separately.
  * spaced M PERFORMANCE badges (M 340D, M 440XI, …) still land on junk ids;
    14 vehicles, single-source, unpublished. Fixing them via the lookahead is
    what breaks the M535i/M635CSi keeps, so it needs the per-spelling rename
    route the block already uses for M 135I/M 140I/M 550D.

Verification, on a frozen cache:
  control (origin/main both repos) 14,065 ids -> treatment 14,069, all 6 kinds
  DISAPPEARED 2 (both aliased above) · NEW 6 · AVAILABILITY CHANGED ELSEWHERE 0
  propose_former_ids: proposed=0 dead_keys=0 evidence_loss=0 unexplained=0
  build exit 0, ALL GATES GREEN (incl. the 4 new spotchecks)
  lint_curation OK (109 files) · lint_claims OK (14 adapters, 5 texts)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01UhxyKpNZhMwKLg7p75Ze7b


rameerez:
**CORRECTION — the M1 figures in this PR's description are wrong, and the dossier was right.**

The body says the dossier's `33,245 of 33,258` "DO NOT REPRODUCE" and gives `3,755 / 3,744 / 11`. That was my error. I summed `cols.reverse.find { |c| c =~ /\A[\d,]+\z/ }` over `uk_veh0120_uk.csv` — the LAST of its 53 columns, which is **`2014 Q3`**. The adapter reads **index 6 = `2026 Q1`**; the header runs newest→oldest and `uk_dft.rb:26` documents exactly that.

Re-measured at the correct column: **33,258 vehicles, 33,245 (99.96%) M135i/M140i, 13 (0.04%) `M1 COUPE`.** The dossier's figures are exact.

**D-1 is unaffected** — the decision rests on the ratio, and both readings agreed the label is ~100% M Performance. Nothing in the code, renames, former_ids or spotchecks changes. The DEBT row has been corrected in-place (4f6c4b4) and now carries the generalised lesson: VEH0120 has 47 historic quarters, so any hand-rolled replay of that file must pin index 6 or read the header rather than picking a numeric column by position.

Caught it because the same faulty method produced a Yamaha measurement that disagreed with an adapter-level replay by 7x — the disagreement was the tell.

PR #165, opened 2026-08-01
G-1 residual: dispositions for the uk_dft designation rescue

rameerez:
G-1 residual: dispositions for the uk_dft designation rescue

Data half of vehiclesdb-pipeline#116, which teaches uk_dft that a Model string's
digits may be an INDEX (Yamaha's MT-07) rather than a DISPLACEMENT (Honda's
CBR1000). Without it, TW_NAMEPLATE fails the capacity guard and drops the row on
the bare GenModel stub — 79,843 gb vehicles pooled on six ids.

removals.yml — ONE entry, motorcycle/yamaha/mt. It was gb-ONLY and decile-1:
27,971 UK vehicles pooled by DfT's family GenModel "YAMAHA MT". Five published
successors (mt-09 11,474 · mt-07 9,194 · mt-10 5,038 · mt-03 1,775 · mt-01 490),
so no single alias target exists by construction — the identical class as
motorcycle/yamaha/xp and motorcycle/yamaha/xsr, which G-1 itself retired the
same way, and this entry is written in their idiom directly below them. All five
successors ALREADY publish from other countries, so nothing disappears: this is
gb evidence arriving on the right records.

spotchecks.yml — 4 rows, and they guard two DIFFERENT code paths on purpose:
  * yamaha/mt-07, yamaha/mt-09 assert gb. These reach the rescue through the
    OUT-OF-RANGE branch ("MT-07" parses, capacity 7 fails the guard).
  * yamaha/yzf-r1 asserts gb through the NON-MATCH branch ("YZF-R1" does not
    parse at all — family "YZF-R" then a single digit, and the pattern needs
    \d{2,4}). #116 shipped with that branch unreachable and fixed it; if mt-07
    keeps gb while this row loses it, the regression is back.
  * yamaha/yzf asserts the SURVIVING residual (24,411 gb). The space shape
    ("YZF R125", "XJ 6 N") is deliberately not covered. If this row ever hits
    zero, someone widened the rule and owes yamaha/yzf a disposition.

styling.yml — NC-N01. The rescue mints motorcycle/neco/nc-n01 (gb+lu, two
independent registers), and the caser writes "Nc-N01". Both registers write it
uppercase. Pinned so the id ships spelled correctly rather than being corrected
after publication.

CONTROL vs TREATMENT (origin/main both repos, frozen cache, all six kinds):
  14,065 -> 14,067 · DISAPPEARED 1 (manifested) · GAINED gb 6 · NEW 3 · LOSSES 0
  build exit 0, ALL GATES GREEN · lint_curation OK (109 files)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01UhxyKpNZhMwKLg7p75Ze7b


rameerez:
Pipeline half is **vehiclesdb-pipeline#116**. Merge that first — this PR's `removals.yml` entry retires `motorcycle/yamaha/mt`, which only stops existing once #116's designation rescue lands. Until then this branch's build legitimately fails the no-vanish gate in reverse.

PR #166, opened 2026-08-01
DEBT: file the flip-back owed for the four HELD BMW enrich blocks

rameerez:
The owner unblock (pipeline `f28c6af`) commented out `car/bmw/m2`, `m4`, `m6`, `m8` in `enrich/bmw.yml` so `lint_enrich` would stop failing every data PR. That was the right call, and the content is preserved in-file rather than deleted — but nothing tracked the flip-back except a NEGOTIATION turn, and the failure mode is quiet.

Those four ids publish with 8–12 countries each. A release that carries them without the flip-back looks complete while shipping **no runs, no era and no variants** for M2/M4/M6/M8 — the kind of gap that reads as "we have no data on this" rather than as a mistake.

So this files it as a one-action item in the 2026.08.1 window: uncomment the four blocks, re-run `lint_enrich`, which must go green because by then the ids are in the committed catalog.

Pairs with **vehiclesdb-pipeline#117**, which records the same rule in `lint_enrich.rb`'s own header — a pipeline change that MINTS ids cannot carry live enrich blocks in the same PR under any merge order, and `CATALOG_DIR` is why a local pair-build can be green while CI is red (the lint measures the committed catalog, not your build).

`DEBT.md` only — no data, no overrides, no code.


rameerez:
Added one more line to this PR: a factual correction to the M1 DEBT row I wrote earlier tonight. It said "53 columns of which **47 are historic quarters**" — 47 is the quarter count, not the historic count. The file is 6 metadata columns + 47 quarters (2026 Q1 … 2014 Q3), of which **46** are historic beyond the newest.

Off by one in an explanatory clause, not in a decision-bearing number (index 6 is still the column to pin, which is what the row is for) — but the row's entire purpose is that hand-rolled replays of this file get its shape wrong, so it should state the shape correctly.

Batched here rather than opened as a third data PR tonight: same file, same ledger, same night. The exact figures are now also encoded as a checked claim in vehiclesdb-pipeline#119.

rameerez:
**Updated: the debt this PR files was paid before the PR merged, so the row is now a precedent rather than an obligation.**

`pipeline#121` flipped the four blocks back once 2026.08.1 published the minted ids. Merging this as *outstanding* would have put a paid debt in the ledger — the same trust problem the ledger exists to prevent.

The row now records the closed loop and the verification that actually matters: the comment-out/flip-back **round trip was lossless**. I compared the loaded YAML structures against my original commit field by field rather than reading the diff — all six blocks are byte-equal (`m2` 2 runs/4 variants · `m3` 6/12 · `m4` 2/7 · `m5` 7/9 · `m6` 2/5 · `m8` 1/4), and `lint_enrich` is OK against the released catalog. That makes this the second worked example of the convention after wave-7 saab/bentley, and the first with a verified round trip.

The general rule is deliberately **not** duplicated here — it lives in `lint_enrich.rb`'s header (pipeline#117), because a tool that enforces a constraint should be where the constraint is written down.

The VEH0120 column-arithmetic correction in this PR is unaffected and still stands.

rameerez:
Superseded before it landed: pipeline#121 (the flip-back itself) merged first — m2/m4/m6/m8 are uncommented and the ids are live in the committed 2026.08.1 catalog, so the debt this files is already paid. No DEBT line needed for a zero-day-old debt; the NEGOTIATION record covers the history. — owner

rameerez:
**Heads-up on this PR's build check: main is red independently of this change.**

`origin/main + origin/main` exits 1 with **72 gate failures** after 2026.08.1 (71 `no-vanish` bare family stubs + 1 `move-split`). Classified against a control build at `2fb22d0` — the commit before #115/#116 merged — which produces *more* failures (85) against the same released catalog. The set of ids vanishing only on current main is **empty**; tonight's merges removed 13.

So it is the publication-hysteresis second pass landing, not this PR and not any open PR. Details in NEGOTIATION Turn 233. A red build here should be classified against that list before anyone reads it as a defect in a DEBT.md-only change.

PR #175, opened 2026-08-01
C-1: two null renames for junk the parenthetical strip exposes

rameerez:
Data half of **vehiclesdb-pipeline#126**, which stops `junk?` discarding every model string containing a bracket (51,175 NL registrations).

Two strings reach the catalog as ids once the bracket comes off, and neither is a nameplate:

| key | raw | why |
|---|---|---|
| `Citroën "Saloon"` | `(II CV BL) SALOON` | stripping the bracket leaves the **body**. `SALOON` appears the same way under Aston Martin, Bentley, Bristol, Cadillac, Crossley and Daimler — a register habit, not a Citroën model |
| `Dodge "Chrysler"` | `CHRYSLER (CDN)` nl 3, `CHRYSLER` nz 2 | the bracket is the **country** (Canada) and the head is the **parent marque** — neither half names a Dodge model |

**No `removals.yml` lines are owed.** Neither id has ever published, so nothing ceases to exist and there is no consumer to inform — `removals.yml` is for ids that die, not strings that were never minted. The comments say that explicitly rather than pointing at a ledger line that does not exist, which is what my first draft did.

Two further cases from the same class — `car/jaguar/gb` and `car/peugeot/f`, both country codes — are fixed in the **rule** in #126 rather than here, because they are systematic: registers write `JAGUAR (GB)` and `strip_make_prefix` eats the make first, leaving the bracket to become the nameplate.

Merge after #126.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -518,7 +518,7 @@ Audi:
   "GT":          Coupe    # raw "GT" nz 5 nl 1, "GT COUPE" nz 4 — nz_nzta's truncation of Coupe GT; the "GT COUPE" rows carry the body word explicitly (dossier §8)
   "Cabriolet 2": Cabriolet  # raw "Cabriolet 2,3E" fi 1, "AUDI CABRIOLET 2,0 TFSI" nl 1 — an ENGINE code (2.3E / 2.0 TFSI) whose decimal comma the slugifier turned into a model number (dossier §8)
   "Ur":          Quattro  # raw "UR QUATTRO" fi 1 nl 1 — "Ur-quattro" is the enthusiasts' retronym for the original; Audi's own name is "Audi quattro", and car/audi/quattro is live (dossier §8)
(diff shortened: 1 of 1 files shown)


rameerez:
**Second commit pushed: the 16 yamaha stub manifests. With this and pipeline#126, main goes GREEN — 0 gate failures.**

That closes all 72 that appeared when 2026.08.1 published.

These were deliberately held out of the first batch (data#173) on the owner ruling: the yamaha stubs ride the change that **defines** their successors rather than a blind manifest pass. That change is C-1 itself — stripping parentheticals surfaces the raws these stubs were pooling. `yamaha/gpd`'s rows land on `gpd125` (17,404), `gpd125d-a` (2,127), `gpd125-a` (2,119), because the register writes `GPD125-A (NMAX 125)` and `junk?` was throwing exactly that away.

**14 of 16** have live successors after C-1 and are retired citing them with counts. **Two — `srv` and `y` — have none even then**, so they are DEMOTED rather than called series stubs: that claim needs successors to point at, and the first batch draws the same distinction for the same reason.

Verification:

```
build exit 0 · ALL GATES GREEN · 0 gate failures   (was 16)
merged-main 14,000 -> 14,048 with C-1 + manifests
DISAPPEARED 0 — identical to C-1 alone, i.e. the manifests moved no data
lint_curation OK (109 files)
```

The title now undersells the PR — it carries the C-1 null renames **and** the 16 manifests. Merge order is unchanged: pipeline#126 first, then this.

PR #176, opened 2026-08-01
DAF: the horsepower rating was publishing as a nameplate — 19 ids fold onto 3

rameerez:
DAF writes the power rating after the series badge and the registers file the
whole string as the model, so `XF480`, `XG530` and `XD370` were publishing as
if they were separate trucks.

Evidence the three-digit run is power and not a designation:
- DAF's own used-truck listings title a unit "DAF XF 430 FT 4X2" while
  carrying a separate "Power (HP): 430" field — the number appears twice, once
  explicitly as a spec.
- The historic caption form is "DAF XF 105.460" — <series>.<power>, which only
  parses if 105 is the series and 460 is not.
- Wikipedia's current range lists the series as separate line items: "XF
  series" / "XG series" / "XG+ series" / "XD series" / "XB series".
  https://en.wikipedia.org/wiki/DAF_Trucks

    FOLD  xf410 xf430 xf440 xf450 xf460 xf480 xf510 xf530  ->  daf/xf  (live)
    FOLD  xg430 xg450 xg480 xg530                          ->  daf/xg  (live)
    FOLD  xd260 xd300 xd310 xd340 xd370 xd410 xd450        ->  daf/xd  (MINTED)

Measured control vs treatment, same cache, same pipeline, only these keys
differing: truck 928 -> 910 ids. 19 removed, 1 added.

## Zero evidence lost, verified at (country,source) — not just country

Every folded id is a strict SUBSET of its target. Both surviving targets keep
`es:es_dgt fi:fi_traficom lu:lu_snca nl:nl_rdw` exactly, and the minted
`daf/xd` carries all four. LOST=[] on every one.

## Two things deliberately NOT folded

- **`xf105` — the trap in this family.** XF105/XF106 are real GENERATION
  series, not power ratings: the same digit-run shape meaning something else
  entirely, exactly like `8D Audi A4` being both a door count and the D2
  platform code. Untouched and verified untouched (fi,nl before and after).
- **`xg-530` ("XG+530")** — XG+ is a distinct DAF series, and minting it needs
  the D-10 `-plus` slug (XG+ slugs to "xg", colliding with the live XG). Rides
  with that mechanism. Untouched here (es,fi,nl before and after).
- `xdc`, `xfc`, `xd-ft` are UNVERIFIED and left alone rather than folded on a
  guess.

## The trap worth reading: this fold created CHAINS on BOTH sides

The DAF block already carried axle-code keys — `XF480F -> XF480`,
`XG530F -> XG530` — from an earlier pass. Renames do not apply transitively,
so the moment `XF480` became a fold key, those rows landed on the intermediate
and STOPPED. Five keys silently did nothing: the first treatment build retired
14 of 19 ids and I only caught the other five by diffing the id lists.

The identical shape then appeared on the alias side, and `lint_curation` rule
1f caught that one for me: five pre-existing `former_ids` entries pointed at
ids that had just become aliases themselves.

Both sets are now repointed straight at the terminal target. A fold that
retires an id which is already a rename VALUE or an alias TARGET has to
flatten both, and the id diff is the only instrument that sees the first half.

## Verifications

- **Blast radius, all six kinds** (renames are make-scoped but KIND-BLIND):
  DAF has 1 van and 3 bus records; NONE matches any key. Only the 19 truck
  records match. Zero cross-kind spill.
- **Gates**: zero failures. The 70 spotcheck failures in the log are car-kind
  checks a `--kinds=truck` build cannot satisfy — identical count in the
  control, so zero regression.
- `lint_overrides`, `lint_curation` (incl. the chain rule), `reorg_make_blocks
  --check`: all green.
- 19 `former_ids` aliases added; no-vanish clean, 0 id-contract failures.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


rameerez:
**S4W — correcting this PR's own body. A load-bearing claim in it is false.**

This PR says the rename-side chain shipped with *"green build, no lint, no gate — caught only by diffing control-vs-treatment id lists"*, and that **nothing lints the rename side**.

That is wrong, and I have repeated it in several NEGOTIATION turns and in three delegate briefs tonight. A delegate working the LF/CF sibling family reconstructed this PR's pre-flatten state and ran the lint against it: **10 failures — 5 `former_ids` CHAIN + 5 renames DIRECTION WAR.**

The rule is at `scripts/lint_curation.rb:393`:

```
# ── DIRECTION WAR: a rename whose TARGET is also a KEY in the same block ─────
```

with its `fail!` at :425. It landed in **data#26 on 2026-07-26 — six days before this PR.** Both halves of the chain trap were catchable pre-build the whole time.

The comment at `lint_curation.rb:334` credits that rule to me. I wrote it, forgot it existed, and then spent tonight asserting the opposite — including in the brief I handed the very agent that caught me.

**What remains true:** the id diff is still required for what the lint genuinely cannot see — a key that fires on nothing (inert), and cross-kind spill. Those are real and the DAF fold hit the first. But "only the id diff sees the chain trap" was false, and the practical consequence is that anyone hitting this class should **run `lint_curation` first**, because it is seconds rather than a six-minute build.

Correcting it here rather than only in NEGOTIATION, because a PR body is what someone reads when they hit this class next.

PR #177, opened 2026-08-01
licences: be_fps pin records "no licence text here" instead of hashing a JS shell

rameerez:
Data half of pipeline#127. **Merge that one FIRST** — with only this half
landed, the old gate reads a be_fps pin carrying no sha256, compares against
nil, and reports "terms CHANGED": a false drift alarm on a source nothing is
wrong with.

- `data/licenses/pins.json`: be_fps becomes `{"mode": "declared_absent"}` with
  the measurement as its `why`, and loses its sha256 — there is nothing to
  hash. The page contains NO licence text (strings "licence", "Creative
  Commons", "CC0" all absent from 22,607 chars, measured 2026-07-25).
- `data/licenses/be_fps.txt` deleted: it was the extracted "licence text" of a
  page that has none.

Regenerated through the sanctioned `rake licenses:pin` path, not hand-edited,
and diffed: **exactly one sha256 removed — be_fps's.** All 13 other pins are
byte-identical, so no licence drifted under cover of this change. The
remaining diff is `pinned_at` timestamps from the re-verification.

The safety this buys, which the old pin did not: `declared_absent` asserts the
source is NOT INGESTED, and the pipeline gate now fails the moment be_fps
appears in a build's source list. Belgium cannot ship without a real terms
page.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


rameerez:
Verifier classification: fresh run post-#127 shows the licence gate QUIET (declared_absent working — first build in seven without a be_fps failure); the 16 remaining reds are all the held-for-yamaha no-vanish stubs, main's known baseline. Merging per the classified-red discipline. — owner

PR #182, opened 2026-08-01
id-contract: retire the van/mercedes-benz/g-class alias + AMG stub tripwire (AMG fix pair, MERGES FIRST)

rameerez:
**Coupled pair. This is the DATA half and it MERGES FIRST.** Paired pipeline PR: vehiclesdb/vehiclesdb-pipeline#130 (batch item 7, the `mercedes()` AMG first-letter-run defect).

Three changes, one of them load-bearing.

---

## 1. `former_ids.yml` — retire the `van/mercedes-benz/g-class` alias (the load-bearing one)

The alias existed because `Reconciler.cross_kind_prune!` (pipeline `reconciler.rb:105`) deleted the van-kind G-Class: `car/mercedes-benz/g-class` held **≥97%** of the registrations, i.e. the pipeline's own verdict was "this IS the car, misfiled by body type".

The paired AMG fix moves **92 van registrations** off the (unpublished) `van/mercedes-benz/amg` stub onto the van G-Class — raws `AMG G 63`, `AMG G 63 4X42`, `AMG G 63 SUV`, in fi|gb|lu|nl. Dominance now reads

```
11,866 / (11,866 + 399) = 96.75%
```

**below the 97% bar**, so the van record survives the prune and publishes. `validate.rb:441` — *an alias may never name a live id* — so the ledger line has to go. The id is no longer retired; consumers holding it resolve to a real record again.

Measured: control vs treatment build, both from pipeline `2511322` against this repo's `main` `f3d4796`, frozen cache.

**`van/mercedes-benz/g` is deliberately left pointing at `car/mercedes-benz/g-class`.** It was flattened through the line being removed, and the obvious tidy-up is to repoint it at the now-live van record — but that record clears the prune by **0.25pp** and could flap back, and an alias whose target dies fails `validate.rb:447`. The car record is unconditionally live and carries `fi,nl`. The stale comment on both that line and the wave-3 block header is corrected rather than left to mislead the next reader.

**Why this half merges FIRST, against the usual normalizer-pair convention:**

| order | result |
|---|---|
| pipeline first | van G-Class live + alias present → `id-contract gate (liveness)` FAILS → red build |
| **data first** | alias gone; with the pipeline unpatched the van record is still pruned and *not* live, and it is **not** in `catalog/van/models.json`, so `no-vanish` never engages either. **Nothing fires.** |

Verified against the published catalog, not assumed.

## 2. `spotchecks.yml` — a tripwire against nulling the AMG stub

After the fix, `car/mercedes-benz/amg` keeps **57,965 registrations across 12 raw rows**, and 57,573 of them are a single `uk_dft` GenModel row whose raw model string is literally `MERCEDES AMG CLASS` — a **range aggregate spanning every AMG nameplate**. No parser can decompose it (the file's `Model` column would; its `GenModel` column cannot), so it is the DEBT `uk_dft` GenModel-altitude item, not the AMG item.

That makes the stub look like leftover junk to anyone who reads it after the fix lands, and nulling it would delete 58k real registrations to hide a source-altitude problem — the exact posture `car/porsche/coupe` is held under. The row asserts `gb` and `nz` (57,573 and 373 registrations, both stable) and carries the reasoning on its face.

## 3. `DEBT.md` — close the hygiene-2 AMG line, and record what the headline hid

`361/367 raws have a live home` is true and was the basis for filing this. **By registration mass it is 22%**: the fix re-homes 16,605 of the stub's 74,570 registrations, because 77% of the mass is that one un-decomposable GenModel row. The line is struck with that correction attached, so the next reader does not re-file it expecting 74k.

---

## Verification

- `ruby scripts/lint_overrides.rb` → `overrides lint: OK`
- `ruby scripts/lint_curation.rb --base=origin/main` → `curation lint: OK (109 files, duplicate-key + shape + make-key + provenance)`
- `ruby scripts/reorg_make_blocks.rb --check` → alphabetical
- Pair build (patched pipeline + this branch): **16 gate failures, byte-identical to the control build's 16** — all pre-existing `motorcycle/yamaha/*` no-vanish stubs already on `main`, none of them from this pair. `diff` of the two FAIL sets: identical.
- The new spotcheck row is evaluated, not skipped: it has no `skip_if_kind_absent`, defaults to `kind: car`, and the id is present in the built catalog with both country codes.


rameerez:
**CI note — the `Build & publish data` red on this PR is pre-existing on `main`, and here is independent corroboration.**

A concurrent, unrelated PR from another session (`s4w/doorcount-rescue-data`, run [30717248381](https://github.com/vehiclesdb/vehiclesdb/actions/runs/30717248381)) fails with **the same 16 `id-contract gate (no-vanish) motorcycle/yamaha/*` failures and nothing else** — the 16-stub manifest follow-up that NEGOTIATION already names as the one step to green main.

My own control build (unpatched pipeline `2511322` + unmodified data `main` `f3d4796`, frozen cache) fails those same 16. The pair build (patched pipeline + this branch) fails those same 16 and no others — `diff` of the two FAIL sets is empty. **This pair adds zero gate failures.**

rameerez:
**Confirmed in CI, not just locally.** This PR's build ([run 30719152003](https://github.com/vehiclesdb/vehiclesdb/actions/runs/30719152003)) failed with **`16 gate failure(s)`** and the FAIL set is exactly the 16 pre-existing `id-contract gate (no-vanish) motorcycle/yamaha/*` stubs — **zero non-yamaha FAIL lines**. Identical to my control build (unpatched pipeline `2511322` + unmodified data `main`) and to the concurrent unrelated PR linked above. This pair adds no gate failure.

`lint` passes. Pipeline half vehiclesdb/vehiclesdb-pipeline#130 has `test` green.

PR #183, opened 2026-08-01
renames: fi_traficom body-first residues fold onto published records (zero new ids)

rameerez:
The curation half of **vehiclesdb/vehiclesdb-pipeline#131** (batch item 5, the `fi_traficom` body-word prefix strip). 20 rename keys + a DEBT.md replacement. **Zero new ids, zero new spellings.**

## Why any curation is needed at all

Finland writes a slice of the older fleet `<BODY> <model> [<engine>][-<typecode>]/<wheelbase>`, so the pipeline PR removes the leading body word and the nameplate finally reaches the normalizer. Two residue shapes then need a line here, and **both were found by diffing a real control-vs-treatment build, not by reading the code**.

### 1. Ten keys where the residue is a TRIM or a TYPE CODE (`car/porsche/911-s` …)

Each of these **published a new id** off one or two Finnish vehicles: `car/alfa-romeo/1750-gt`, `car/mercedes-benz/220-b`, `/280-sl-8`, `/280sl-8`, `/380-sl-automatic`, `/500-sl-auto`, `car/peugeot/304-b01`, `/504-b02`, `car/porsche/911-s`, `car/saab/900-turbo-s`.

Not by volume — **through `publishable?`'s `sources.size >= 2` arm**, because nl_rdw/nz_nzta already carry the same fragment for the same cars. One Finnish vehicle is enough to tip an existing single-source candidate over the corroboration line. Each key folds onto the record the car actually is (`"911-S": "911"`, `"504 B02": "504"`, `"280 SL-8": 280 SL` …).

### 2. Ten `junk?`-rescue keys where the residue is GLUED to the type code

`COUPE 944-94/2400` → `944-94`: no letters, not 2–4 digits, so `junk?` deletes it. **Those 32 vehicles publish today under the fake nameplate `Coupe`** — without these keys the strip would delete them instead of re-homing them, which is exactly what DEBT.md refuses ("nulling it deletes 251 real Finnish registrations to hide a normalizer bug"). Renames run before `junk?` (the ORDER FIX), so a key on the produced string rescues it: `"944-94": "944"`, `"924-92": "924"`, `"928-91"/"928-92"/"9285-92": "928"`, `"280 Se-9-3.5"/"280SE-9-3.5"/"280 Se 3.5": 280 SE`, `"190E 2.3": 190 E`, `"100-6-3000": 100-6`.

**SAFETY (the data#179 standard):** every value is the name of a record that is **already published**, verified byte-exact against `catalog/car/models.json`. This batch can mint neither an id nor a spelling.

## Measured

* **This half ALONE, on unmodified pipeline `main`**: **+0 ids / −0 ids** (96 popularity ranks move — one key, `"500 SL Auto"`, is already reachable through es_dgt today). `test_override_key_reachability` green against pipeline main.
* **The PAIR**: **+1 / −0**, and the one addition is `car/packard/400` — a real nameplate (the 1955–56 Packard Four Hundred; the raw is `HARDTOP 400/3061`). Gate failures byte-identical to the control build.
* `lint_overrides` OK · `lint_curation --base=origin/main` OK (109 files) · `reorg_make_blocks --check` OK.

Full method, vocabulary derivation and the refuted door-prefixed scope are in the pipeline PR.

## DEBT.md

The `normalizer.rb#porsche` / body-word filing is **retired** (its own convention: "when you take one, delete its line in the same PR") and replaced by the **two items this pass measured and could not close**:

1. **The fake body-word nameplate class is catalog-wide and only Finland's half is fixed.** 92 published car ids are named after a body word; **65 of them have no `fi_traficom` source at all** (by source: nl_rdw 84, nz_nzta 82, fi_traficom 27, uk_dft 14, us_fueleconomy 12, ua_mvs 10). `car/porsche/coupe` therefore *survives* the fi fix — its Finnish mass drops from rank 445 to 2,489, but nl+nz still carry it. The item also names the ~15 that are **real** and must never be swept (`car/audi/coupe`, `car/fiat/coupe`, `car/mazda/mpv`, `car/smart/roadster`, `car/tesla/roadster`, `car/mini/convertible` …).
2. **An adapter-level prefix strip cannot be verified from candidates — only a published-id diff sees it — and the `<door><BODY>` half is refuted at that level:** 5,158 rows onto 326 existing records, but **118 new published ids**, 105 of them type-approval strings, none of which clears the volume floor. Reopening that route needs either a residue test that can tell `944` from `280` (catalog knowledge the adapter does not have) or a corroboration rule that does not count two registers quoting the same type code as two independent sources.

## Merge order

**COUPLED PAIR** with pipeline#131. House rule is pipeline-first; **here data-first is measurably better** (this half is +0/−0 standalone, while pipeline-first leaves 10 garbage ids live for the window). Either order is safe.

Do NOT merge — for review.


rameerez:
CI note — **the red "Build & publish data" is `main`'s, not this PR's.** The run fails with exactly the 16 pre-existing `id-contract gate (no-vanish) motorcycle: motorcycle/yamaha/*` stub failures (run 30719372371). My frozen-cache CONTROL build, on unmodified pipeline `2511322` + data `4ff4124`, fails with the **same 16 lines, byte-identical** — diffing control-vs-treatment FAIL lines returns empty for both the pipeline half and the pair. That is the "16-stub manifest follow-up is the one step to green main" item already on record. `Lint data` (the check this PR can actually move) is green.

PR #186, opened 2026-08-01
spotchecks: pin the Paceman/Aceman split and smart's hashed lineup

rameerez:
> **BLOCKED BY TWO PIPELINE PRs — DO NOT MERGE FIRST.** This repo's `build`
> job checks out `vehiclesdb-pipeline` at its DEFAULT BRANCH (no `ref:`), so
> until vehiclesdb/vehiclesdb-pipeline#133 and #134 are on pipeline main this
> PR's build is EXPECTED to be red on exactly the five rows it adds. That red
> is the coupled-pair signature, not a defect in this PR: all five rows were
> verified PASS on a local build with both pipeline halves applied, and FAIL on
> a control build without them. Merge order: pipeline#133 → pipeline#134 → this.

Five tripwire rows and one vocabulary word, for two pipeline corrections whose
failure modes are silent.

THE WRONG-CAR ROWS. `normalizer.rb#mini` tested `/ACEMAN/` without a word
anchor and `"PACEMAN"` matches that at offset 1, so every MINI Paceman row
published as a MINI Aceman and `car/mini/paceman` did not exist. The record
therefore claimed a 2024 all-electric crossover was registered in Canada and
the United States. The CA and US sources hold 16 and 24 Paceman rows and ZERO
Aceman rows; BMW put the Paceman on sale in the US on 16 March 2013
(press.bmwgroup.com T0134695EN_US) and the Aceman premiered 24 April 2024
(T0441184EN). The EU type approvals corroborate it independently: the record's
`e1*2007/46*0563*03` moves to the Paceman, the `e1*2018/858*00382*…` family
stays with the Aceman.

The two rows are one assertion in two halves — the Paceman is its own record
AND carries North America, and the Aceman does NOT. If the anchor regresses,
the first fails on a missing id and the second on the restored ca/us claim.

`availability_excludes` is NEW vocabulary (pipeline PR adds the gate side).
It exists because this correction REMOVES availability from a published
record: an `availability_includes` row cannot say "and it must never come
back", so a silent re-merge would restore the wrong claim with a green build.
The word is added to `scripts/lint_overrides.rb`'s closed list in the same
commit — a word the gate reads but the lint rejects is a red build, and a word
the lint accepts but the gate does not read is a row that asserts nothing.

THE SMART ROWS. `#1`/`#3`/`#5` are one alphanumeric plus one symbol, which is
the corruption guard's shape, and that guard sits inside `classify` upstream of
the rename lookup — so no line in THIS repo could ever have rescued them. 3,027
vehicles across es|fi|lu|nl|nz|ua were deleted on every build; Germany got in
only because KBA strips the symbol itself. Pinning `de, es, nl` means a
regression that leaves only the German rows fails loudly.

DECISIONS.md records the removal itself, because "availability is evidence"
needs the other half stated: removing a country is a correction only when the
evidence is shown to belong somewhere else, and here it moved rather than
vanished.

DEBT.md records that `enrich/mini.yml` (dossier §B.1) is now unblocked, with
the disposition measured on a real build: the `car/mini/aceman` block attaches
LIVE (the id neither moves nor is minted), `car/mini/paceman` ships HELD and
commented (the anchor mints it, and a minted id is not live under any merge
order), flip-back in the release that publishes it.

VERIFIED against a control build, and the panel is a real tripwire rather
than a rubber stamp: ALL FIVE rows FAIL on the control catalog (mini/paceman
missing entirely; mini/aceman still carrying ca+us; smart/1,3,5 missing es+nl)
and all five PASS on the treatment catalog, both built from the same frozen
cache at the same base commit. `ruby scripts/lint_overrides.rb` OK.

MERGE ORDER — this PR goes LAST. The mini rows assert an id that only exists
once the pipeline anchor PR is merged, and the `availability_excludes` key is
only read once that same PR is merged. The smart rows need the leading-hash
pipeline PR. Merging this first turns main red.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Pairs with vehiclesdb/vehiclesdb-pipeline#133 and vehiclesdb/vehiclesdb-pipeline#134.


rameerez:
CI red as predicted, and the shape of the red is itself the evidence for the merge order — recording it rather than leaving it to be re-derived.

`build` failed with **23** gate failures. Decomposed:

- **16** are the pre-existing `motorcycle/yamaha/*` no-vanish failures already on main (the parenthetical change's data half is not merged). Present on every build tonight, mine included, control and treatment alike.
- **7** are the five rows this PR adds, failing against **pipeline main** exactly as the header says:
  - `mini/paceman (car) missing from catalog` — the anchor is not merged, so the record does not exist
  - `smart/1 availability ["de","fi","gb"] missing es` / `missing nl`
  - `smart/3 availability ["de","fi"] missing es` / `missing nl`
  - `smart/5 availability ["de","ua"] missing es` / `missing nl`

Note what is **absent** from that list: `mini/aceman` did not fail, even though it still carries `ca` and `us` on this build. That is not a passing assertion — pipeline main's `validate.rb` has never heard of `availability_excludes`, so the row is silently inert. It is the "a word the lint accepts but the gate does not read is a row that asserts nothing" case, demonstrated live, and a second independent reason the pipeline half must land first.

With both pipeline halves applied locally (same frozen cache, same base commits), the same 23 collapse to the same 16: all five rows pass and the pair adds no failure.

PR #187, opened 2026-08-01
DEBT: §C-3 space shape measured at 87,302 vehicles — and the obvious fix is refuted

rameerez:
DEBT.md only.

The merged designation rescue fires only on a **hyphen**; names whose index follows a **space** still pool on the bare GenModel stub. Re-measured post-C-1 at the newest quarter: **87,302 gb vehicles across 75 stubs in 38 makes** — far larger than the dossier framed it, which was yamaha-scoped.

```
YAMAHA/YZF 24,411 · TRIUMPH/BONNEVILLE 9,167 · YAMAHA/TRACER 4,524
SYM/JET 4,312 · PIAGGIO/VESPA 4,155 · DUCATI/MULTISTRADA 4,142
SUZUKI/GSX 3,989 · YAMAHA/XJ 3,926 · BETA/EVO 3,874 · DUCATI/PANIGALE 3,615
```

**I simulated the obvious fix and it is refuted.** Fusing the stub with the following designation token — same containment discipline as the hyphen rescue — produces `BONNEVILLE-T120`, `XJ-6`, `TRACER-9`, `TUONO-V4`. The live catalogue already publishes every one of those targets under a *different* spelling:

| live id | live name | my fused form |
|---|---|---|
| `yamaha/xj6` | **XJ6** (fused) | `XJ-6` ✗ |
| `yamaha/yzf-r125` | **YZF-R125** (hyphen) | `YZF-R125` ✓ |
| `yamaha/tracer-9` | **Tracer 9** (space) | `TRACER-9` ✗ display |
| `triumph/bonneville-t120` | **Bonneville T120** | `BONNEVILLE-T120` ✗ display |

Three different join characters inside Yamaha alone. A blind hyphen-join mints `XJ-6` beside the live `XJ6` — the largest mis-spelling event in the catalogue’s history.

**So it is a ROUTING problem, not a naming one**: every target already exists. That makes it curation (per-family rename keys, join char read from the live record rather than derived), and the measurement above is the priority order.

One trap recorded for whoever takes it: `PIAGGIO/VESPA` is not a join problem at all — `piaggio/vespa-et4` ("Vespa ET4") and `vespa/et4` ("ET4") are **both live**, so it is the badge split and wants the co-move treatment from data#170.


rameerez:
Superseded by a clean branch off current main — this one was based on the yamaha fold branch, which has since merged, so rebasing replayed 98 fold commits into a conflict. Reopened as the same single DEBT row with no history to replay.

PR #188, opened 2026-08-01
DEBT: §C-3 space shape measured at 87,302 vehicles — and the obvious fix is refuted

rameerez:
DEBT.md only — one row.

The merged designation rescue fires only on a **hyphen**; names whose index follows a **space** still pool on the bare GenModel stub. Re-measured post-C-1 at the newest quarter: **87,302 gb vehicles across 75 stubs in 38 makes** — far larger than the dossier framed it.

```
YAMAHA/YZF 24,411 · TRIUMPH/BONNEVILLE 9,167 · YAMAHA/TRACER 4,524
SYM/JET 4,312 · PIAGGIO/VESPA 4,155 · DUCATI/MULTISTRADA 4,142
SUZUKI/GSX 3,989 · YAMAHA/XJ 3,926 · BETA/EVO 3,874 · DUCATI/PANIGALE 3,615
```

**I simulated the obvious fix and it is refuted.** Fusing stub + designation produces `BONNEVILLE-T120`, `XJ-6`, `TRACER-9`. The live catalogue already publishes every target under a different spelling — `yamaha/xj6` is **XJ6** (fused), `yamaha/yzf-r125` is **YZF-R125** (hyphen), `yamaha/tracer-9` is **Tracer 9** (space). Three join characters inside one make. A blind hyphen-join mints `XJ-6` beside the live `XJ6`.

**So it is a routing problem, not a naming one** — every target already exists. That makes it curation with the join char read from the live record rather than derived, and the numbers above are the priority order.

One trap recorded: `PIAGGIO/VESPA` is not a join problem — `piaggio/vespa-et4` and `vespa/et4` are both live, so it is the badge split and wants data#170 co-move treatment.

(Replaces data#187, which was branched off the yamaha fold and went DIRTY when that merged.)


rameerez:
**S4W cross-review: CONFIRMED, and one constraint to add.**

Verified your load-bearing claim against the committed catalog rather than taking it — three join characters inside one make is exactly right:

```
yamaha/xj6        "XJ6"        fused    fi,lu,nl,nz,ua
yamaha/yzf-r125   "YZF-R125"   hyphen   fi,lu,nl,ua
yamaha/tracer-9   "Tracer 9"   space    nl,nz,ua
```

So a blind hyphen-join really would mint `XJ-6` next to the live `XJ6`. Your conclusion that it is a **routing** problem rather than a naming one follows, and the join char has to be read off the live record.

The Vespa trap also checks out and is worse than a note suggests — **all three** are live simultaneously:

```
piaggio/vespa-et4   "Vespa ET4"
piaggio/vespa       "Vespa"
vespa/et4           "ET4"
```

That is the badge split with a bare-stub twin on top, so data#170's co-move treatment is the right call.

---

**The constraint I would add to the row: the stubs are NOT retirable.**

```
yamaha/xj       gb,nl,nz
yamaha/yzf      gb,nl,nz,ua
yamaha/tracer   fi,gb,nl,nz
```

Every bare stub carries **non-gb evidence too**. So this is gb registrations *moving off* them, not stubs being folded away — after the fix each keeps its nl/nz/fi/ua availability and stays live.

Worth stating explicitly in the row, because the G-1 precedent it sits next to went the other way: `yamaha/mt` was gb-ONLY and decile-1, so it was retired with a manifest. Someone applying the mt idiom here would retire records that still hold real evidence in three other countries. Exactly the same shape the AMG residue table just flagged — that stub keeps its other nine countries and is not retired either.

Two independent instances in one night of "the stub survives the cleanup" is probably worth being a stated rule rather than a per-item footnote.

rameerez:
**Correction pushed — the headline number in this row measured the wrong quantity, and an independent derivation caught it.**

The row said "87,302 vehicles across 75 stubs in 38 makes". That figure is real, but it is **what my candidate fusing rule would move**, not how much is pooled — my simulation incremented its counter only inside the branch where the rule fired, and the row described it as the problem's size.

A second derivation replicated `two_wheeler_nameplate` verbatim from `uk_dft.rb`, kept only rows reaching `count_fallback`, and dropped the `MODEL MISSING` rows the normalizer kills anyway. The pooled residue is **441,049 gb vehicles across 1,022 stubs in 215 makes**. For the seven highest-volume makes: 219,986 veh / 543 pairs, of which **134 rows / 93,290 veh route onto ids that exist but lack `gb`** — that subset is the actionable prize.

**The refutation is unchanged**: the join character is family-specific and the live catalogue proves it.

Three traps added from the replay, all of which would have bitten a naive pass:
- the dominant unmatched shape is **E5/E4/ABS/model-year suffixes** — `YZF R1` alone has 15 model-year rows totalling ~7.5k veh, and none of those is identity
- `TRIUMPH BONNEVILLE T 100 (865)` (4,546 veh) normalises onto `bonneville-t100`, but **`triumph/bonneville-865` is live** — routing it silently merges two generations
- `yamaha/r` and `yamaha/x` are single-letter stubs pooling RayZR (1,378) and X-MAX Tech Max 125 (824); eleven stubs would empty out entirely if routed, retiring the id

PR #195, opened 2026-08-01
renames: VW New Beetle family — 25 keys onto `beetle`; the brief's premise is refuted, and the fold target is inverted

rameerez:
Owner backlog item 8 — the VW New Beetle fragment family.

## The headline: the load-bearing constraint no longer holds

I was scoped to fold ~18 fragments onto **one `volkswagen/new-beetle`**, on the
stated constraint that it is a DISTINCT PRODUCT from `volkswagen/beetle` and
must **never** fold onto it. I was asked to verify that constraint first.

**It is refuted.** I inverted the target rather than ship a systematic
duplicate. Four independent grounds, each measured:

1. **`New Beetle: Beetle` is already shipped and firing.** `renames.yml` §5.2
   (data#110, 2026-07-27). A classify replay on this build shows it fire in
   **six countries** — ca, es, fi, nl, ua, us.
2. **Two more shipped keys agree.** `Newbeetle: Beetle` (data#110) and
   `"2D New Beetle": Beetle` (data#179, **2026-08-01 22:45** — twelve hours
   before this PR).
3. **A published alias already says so.** `former_ids.yml:4092` retires
   `car/volkswagen/newbeetle` → `car/volkswagen/beetle`.
4. **Canon.** DECISIONS.md: *"Model = nameplate."* NAMING.md §6 keeps numbered
   generations separate **for two-wheelers only** — *"distinct products in a way
   that a car's Mk2 is not"*. `volkswagen/beetle` already spans the Type 1 and
   the A5 Beetle; the New Beetle is its third generation, not a fourth id.

DECISIONS.md's fold safeguard — *"before folding A into B, look for a published
record that says A is its own model"* — **passes**: none exists.
`volkswagen/new-beetle` is a candidate at n=2 that exists **only** because one
nl_rdw row is spelled `NEW-BEETLE` with a hyphen. A typo artifact, not a
naming decision.

Corroborating source, re-verified byte-exact against the pinned cache this build
reads — EPA/DOE `vehicles.csv`: model `"New Beetle"` → baseModel **`"Beetle"`**
(61 rows); `"New Beetle Convertible"` → baseModel **`"Beetle"`** (21 rows).

**Counter-evidence, stated honestly:** Wikipedia does treat the New Beetle as
its own article, and mechanically it is a different car (Golf Mk4 platform, FWD)
from the Type 1. That is a real argument for a split — but it is an argument
against the *canon*, not against this PR, and the canon already settled it for
the car kind. If the owner wants to reverse it, see "coupled pair" below.

## The chain trap this family was built to spring

`New Beetle` **is itself a rename key**. Folding the fragments onto it would
have stranded every row one level short — renames apply ONCE — minting
`volkswagen/new-beetle` while bare `New Beetle` rows kept folding to `Beetle`.
That is a *systematic* duplicate, arriving silently, and strictly worse than
today's long tail. `lint_curation` rule 1f catches the alias side of this; the
rename side is uncaught.

`Beetle` is **not** a key. Verified: **zero** key-whose-target-is-a-key in the
entire VW block. All 25 lines terminate in one hop.

## Fragment list, as measured today vs the ~18 filed

All 18 filed fragments still exist. The family is **larger** than filed:
`Nem Beetle`, `Bew Beetle`, `Ngs New Beetle`, `Vw New Beetle`,
`Volkswagen-Vw New Beetle`, `Vw New Beetle 2`, `2 D New Beetle`, and
`New-Beetle` itself were not on the list. **25 folded + 1 declined = 26.**

## Method — the keys are the PRODUCED strings

Derived by **instrumenting `Normalizer`'s own per-make rename hash** and
replaying every source row through `classify` over all six kinds, so each key is
the exact string the lookup sees. Not inferred from a display name — that is
how inert keys have shipped before.

## Control vs treatment

Same pipeline (`6101c2c`), same APFS-cloned **frozen cache**
(`find cache -type f ! -name 'license_*.txt' -exec touch {} +` — licences
excluded), same base (`50cf50a`), pristine control worktree, all six kinds.
Nothing shared was mutated.

| | control | treatment |
|---|---|---|
| published ids | 14,027 | 14,027 (0 appeared, 0 disappeared) |
| candidates | 146,487 | 146,462 (**−25**, 0 added) |
| gate FAILs | 17 | 17 — **identical**, all pre-existing |
| exit | 0 | 0 |

The 25 removed candidates are **exactly one per key**.

**The zero published-id delta is real, not a broken control** — every fragment
was candidate-only, so the fold moves mass without changing the published id
set. The effect is visible where it should be:

    registrations onto volkswagen/beetle:  nl 12671→12696 (+25)
                                           fi  2366→2371  (+5)
                                           es    76→77    (+1)   = +31 vehicles

which reconciles exactly with the removed candidates' own counts.

## Availability union, at (country, SOURCE)

Fragments contribute `{nl,nl_rdw}` `{fi,fi_traficom}` `{es,es_dgt}`. The target
carries all ten pairs — `ca,ca_nrcan · es,es_dgt · fi,fi_traficom · gb,uk_dft ·
lu,lu_snca · my,my_jpj · nl,nl_rdw · nz,nz_nzta · ua,ua_mvs · us,us_fueleconomy`
— **identical in both builds**. All three fragment pairs are already in it.
**Zero (country, source) pairs lost.**

## Key-fires proof

Control-vs-treatment id diff on all 25: **25/25 FIRE, ZERO INERT.**

## Retired ids

**None needed.** All 25 are candidate-only ids, never published — nothing leaves
the published id contract, so no `former_ids` alias and no `removals.yml`
manifest (and correctly not both).

## Declined — `New Beetle Pick Up`

van, nl, nl_rdw, n=1, raw `"NEW BEETLE PICK UP"`. **Not folded.** VW built
exactly two New Beetle bodies — *"3-door hatchback (Coupé)"* and *"2-door
convertible (Cabriolet)"* — never a pickup. A van-kind New Beetle pickup is an
owner **conversion**, and the settled line in this very block for modifications
on a Beetle base is `null`, not a fold (`Baja`, `Beach Buggy`, `Buggy`,
`Dune Buggy`). Folding it would claim for the nameplate a body style VW never
made; nulling it needs its own `removals.yml` manifest and review. Left as an
n=1 candidate — it cannot publish without corroboration. Verified still present
in the treatment build.

## Coupled pair? No — but one decision is upstream of it

- **Not coupled to the uk_dft Model-column detector.** All 25 keys come from
  `nl_rdw`, `fi_traficom` and `es_dgt`. **Zero gb rows** in this family, so
  changing gb produced strings cannot make these keys inert.
- **No pipeline change.** Data-repo only, one file.
- **The one coupling is a decision, not code.** If the owner rules the New
  Beetle *is* distinct, this PR is not the thing to change first — the reversal
  has to start with `New Beetle: Beetle` (data#110) and `"2D New Beetle":
  Beetle` (data#179) plus the `newbeetle` alias. This PR's 25 targets then flip
  mechanically. Shipping it in the brief's original direction *without* that
  reversal is the one option that is wrong under both readings.

## Lints

`lint_overrides` · `lint_curation --base=origin/main` · `reorg_make_blocks
--check` · `lint_plates` — **all green** (re-run after rebase).

## Out of scope, filed for whoever owns it

- **Make-column defects (es_dgt):** `Volkswagen Ag|New Beetle` and
  `Volkswagen Vw|New Beetle` produce `volkswagen-ag/new-beetle` and
  `volkswagen-vw/new-beetle`. Renames are make-scoped, so the `Volkswagen:`
  block cannot reach them — this needs `overrides/makes/`. Same defect spawns
  `volkswagen-beetle/touran`, `volkswagen-beetle/golf`, `volkswagen-beetle/passat`.
- **9 strings die before publishing** and could be rescued onto `beetle` the
  same way (renames run before `junk?`): `New Beetle 2.5`, `New Beetle 2.0 I`,
  `New Beetle 1.8 T`, `New Beetle 2.0`, `New Beetle 1.9`, `9C New Beetle`,
  `2D New Beetle Hb 1.8`, `2D New Beetle 2.0-Automatic-1C`,
  `2D Hatcback New Beetle 2.0-9C/250`. Deliberately not in this PR — it changes
  the character from fold to rescue. Measured and listed so it is not re-derived.


rameerez:
## Re-verified end-to-end under a private namespace (shared-scratchpad clobbering advisory)

Fourteen delegates share this session's scratchpad, and two sibling artifacts were
silently clobbered tonight by agents writing generic filenames into the shared root.
My first-pass measurements used `build-ctrl` / `build-treat` / `ctrl.log` — exactly
that collision-prone class, and they were the inputs to the id diff, which is the
**only** instrument that catches an inert key. So I discarded them and **re-ran both
builds and both replays from scratch** under a private namespace.

Two things were also **tightened** in the redo:

- **The control is now my commit's exact parent (`ec1fff5^`)**, not the older base I
  first used. The only delta between the two builds is this PR's 74 added lines.
- **The key set under test is derived from `git show ec1fff5^ vs ec1fff5`**, not from
  any scratch file — so the list of keys being proved cannot itself be clobbered.

Cache integrity re-checked independently: the EPA figures reproduce exactly
(`New Beetle` → baseModel `Beetle`, 61 rows; `New Beetle Convertible` → `Beetle`, 21).

### Every number reproduces, unchanged

| | control | treatment |
|---|---|---|
| published ids (6 kinds) | 14,027 | 14,027 — 0 appeared, 0 disappeared |
| candidates | 146,487 | 146,462 — **−25**, 0 added |
| availability (country,SOURCE) changes on surviving ids | — | **none** |
| `volkswagen/beetle` union loss | — | **`[]`** (all 10 pairs identical) |
| registrations on `volkswagen/beetle` | es 76 · fi 2366 · nl 12671 | es 77 · fi 2371 · nl 12696 (**+31**) |
| gate FAILs | 17 | 17 — **identical** |
| exit | 0 | 0 |

**Key-fires: 25/25 FIRE, 0 INERT.** The declined `van/volkswagen/new-beetle-pick-up`
is present in **both** builds, as intended.

One first-pass anomaly resolved: `New Beet` showed `ABSENT` in my initial control.
That was **my own filter bug** — the first replay regex lacked a bare `beet`
alternative — not clobbering. In the clean re-run it reads
`ctrl: Volkswagen|New Beet -> treat: Volkswagen|Beetle`.

## The red `build` check is pre-existing and repo-wide, not this PR

The build job reports **16 gate failures, every one `motorcycle/yamaha/*`
id-contract (no-vanish)**. Zero mention of Volkswagen or Beetle:

```
FAIL id-contract gate (no-vanish) motorcycle: motorcycle/yamaha/cg — no former_ids alias …
… gpd, gts, mw, srv, szr, ttr, tw, ty, v, vp, xc, xvz, y, yb, yfm
```

Three independent confirmations that it is not mine:

1. My frozen-cache **control** build — pristine parent, no change at all — fails the
   **same 17** (CI sees 16 because it fetches live; de_kba 2026-07 and es_dgt 202607
   both 404 on the runner, so its corpus differs slightly from my frozen cache).
2. Control and treatment failure sets are byte-identical.
3. **Every concurrent sibling branch is red the same way** — `s4w/xj-recut`,
   `s4w/lexus-folds`, `s4w/lr-power-codes` and this one all failed in the same window.

This PR is blocked on the yamaha id-contract work landing, not on anything in it.
`lint` — the check that actually gates curation — **passes**.

rameerez:
## Re-verified a third time, on the merged base (`151fd74`) — and the build is now fully green

`main` was merged into this branch after my measurements. That merge changed
`renames.yml` by +117 lines plus `former_ids.yml` and `removals.yml`, so I re-ran
the whole control/treatment pair on the merged tree rather than assume it was inert.

**Why this needed checking, not assuming:** if any sibling had added a `Beetle:`
rename key, all 25 of my keys would silently become a CHAIN and strand at
`volkswagen/new-beetle` — the exact failure this PR exists to avoid.

Verified on the merged tree:

- `Beetle` is **still not a rename key**; **0 chains** in the VW block.
- My 25 keys are all present and all still resolve to `Beetle` (25/25).
- The merge touched **zero** Volkswagen or Beetle lines (`git diff … | grep -i 'volkswagen\|beetle'` → empty).
- All four lints pass on the merged state.

Control for this run = merged base with **only** my 74 lines reverted, so the single
variable is this PR (VW block 483 → 508, delta exactly 25).

| | control | treatment |
|---|---|---|
| published ids (6 kinds) | 13,936 | 13,936 — 0 appeared, 0 disappeared |
| candidates | 146,483 | 146,458 — **−25**, 0 added (same 25 ids) |
| availability (country,SOURCE) changes | — | **none** |
| `volkswagen/beetle` union loss | — | **`[]`** |
| registrations on `volkswagen/beetle` | es 76 · fi 2366 · nl 12671 | es 77 · fi 2371 · nl 12696 (**+31**) |
| **gate FAILs** | **0** | **0** |
| exit | 0 | 0 |

Every figure reproduces. The published-id total moved 14,027 → 13,936 purely because
other delegates' folds landed in `main` in the interim; the delta attributable to this
PR is unchanged at 0 published / −25 candidates.

**The build check should now go green.** The 16 `motorcycle/yamaha/*` id-contract
failures I reported earlier were pre-existing and have been fixed on `main` — both
sides of this run now report **zero** gate failures, where the pre-merge pair reported
17 apiece.

Independently corroborated: the `audit_rename_value_liveness.rb` sweep from data#197
was run against this branch by another session and reports
`s4w/new-beetle-family retires NOTHING — class does not apply`, matching my own
finding that no `former_ids` alias and no `removals.yml` manifest are required.

PR #199, opened 2026-08-02
DAF: the LF and CF power suffixes were publishing as nameplates — 26 ids fold onto 2

rameerez:
DAF writes the horsepower rating after the series badge, and the registers file
the whole string as the model. `data#176` fixed that for XF/XG/XD/XB and stopped
there. `LF` and `CF` carry the identical defect.

    FOLD  lf180 lf210 lf220 lf230 lf250 lf260 lf280 lf290 lf310 lf320   -> daf/lf  (live)
    FOLD  cf260 cf290 cf300 cf320 cf330 cf340 cf370 cf400 cf410 cf430
          cf440 cf450 cf460 cf480 cf510 cf530                           -> daf/cf  (live)
    KEEP  lf45 lf55 cf65 cf75 cf85 — real series designations

Measured today against the audit's expectation of 10 + 16: **exactly 10 and 16.**

## The evidence is per-record in our own corpus, not a caption

`data#176` argued from a DAF marketing page and a photo caption. For LF/CF I could
re-derive it from the registers themselves, which is stronger.

**es_dgt states the number twice.** The Spanish microdata is fixed-width and
carries engine power in kW as an explicit spec field (offset 224). Converting
kW->hp (x1.35962) reproduces the badge:

| raw model | kW (register) | -> hp | badge |
|---|---|---|---|
| `CF 530 FA`  | 390.0 | 530.3 | 530 |
| `CF 480 FT`  | 355.0 | 482.7 | 480 |
| `CF 450 FT`  | 330.0 | 448.7 | 450 |
| `LF 230 FA`  | 172.0 | 233.9 | 230 |
| `LF 260 FA`  | 194.0 | 263.8 | 260 |
| `LF 320 FA`  | 239.0 | 324.9 | 320 |

26 of 30 distinct LF/CF model strings in `es_dgt_202603` match their own badge
within ±8 hp.

**nl_rdw carries the `<series>.<power>` form, and both halves are present.**
`"FA LF 45.180"` x3, `"LF45.180"` x2, `"LF 55.250"` x2, `"FT CF 85.460"` x6,
`"FTCF85.410T"` x9, `"FANCF75.360U"` x3 — and the very same numbers reappear
directly after the *bare* badge: `"LF 180 FA"` (472 vehicles), `"CF 460 FT"`,
`"CF 410 FT"` (1,928). A number that is the power of an LF45 cannot be a
designation when it follows a bare LF.

## The trap: there are FIVE real series here, not one

`data#176` kept `xf105` because XF105/XF106 are genuine generations. The LF/CF
equivalent is bigger, and getting it wrong would have destroyed five live
nameplates. **The kW instrument above doubles as a known-answer test, and it
answers NO on all five:**

| raw | kW | -> hp | badge digits | verdict |
|---|---|---|---|---|
| `FA LF 45` / `FA LF55` | 181.6 | 246.9 | 45 / 55 | **not a rating** |
| `DAF CF85` | 315.0 | 428.3 | 85 | **not a rating** |
| `FACF65`   | 218.7 | 297.3 | 65 | **not a rating** |

Corroborated three more ways: they take their *own* power suffix (`LF45.180`,
`CF 85.410`) and a series cannot be the rating and carry a rating at once;
[DAF CF](https://en.wikipedia.org/wiki/DAF_CF) — *"Originally launched as the 65,
75, and 85 series (from 1992 through 1997)"*, *"renamed the CF range in 1998"*,
*"DAF 65, 75 & 85 CF (1998–2000)"*; [DAF LF](https://en.wikipedia.org/wiki/DAF_LF) —
*"The LF45 and LF55 are powered by Cummins B Series engines... all LF55s use the 6
cylinder due to their increased size and weight"*. And the lineage: they succeed
the Leyland-DAF 45/55 this file already publishes as `truck/leyland-daf/45` and
`/55`.

All five verified untouched in the id diff.

## Availability union, at (country, SOURCE)

Both parents already carry `es:es_dgt fi:fi_traficom gb:uk_dft lu:lu_snca
nl:nl_rdw`. Every one of the 26 is a **strict subset** — the id diff reports
**0 availability changes anywhere in the catalog**, which is the same statement
measured rather than asserted. Nothing lost; unlike `daf/xb` in #176, nothing
arrives either.

## Control vs treatment — all six kinds

Same base (`aeccb6a`), same pipeline (`6bbd4ee`), and a **private APFS clone of
the cache** used for both runs, mtime-frozen except `license_*.txt` so the
licence gate stays honest. Cache manifest verified byte-identical before the
control, after the control, and after the treatment.

    CONTROL 13936 ids   TREATMENT 13910 ids   (-26)
    REMOVED 26 — exactly the intended ids
    ADDED 0 · AVAILABILITY/NAME CHANGED 0 · NON-DAF ids touched 0

    truck published 907 -> 881;  car/motorcycle/moped/van/bus IDENTICAL
    gate failures: 0 on BOTH sides (EXIT=0 both)

**Cross-kind spill: zero.** Make-scoped renames are kind-blind and DAF has van
and bus records; all 7 non-truck DAF/Leyland-DAF records are byte-identical
between the two builds.

## Key-fires proof, including a deliberate negative control

All 26 fold keys fire — each retires its id in the diff above. The 27th change is
the chain flatten, which retires nothing visible, so I proved it separately by
building a **negative control that differs from this branch by exactly two
lines** (the flatten reverted, nothing else):

    with the flatten      truck/daf/lf260  ABSENT
    without the flatten   truck/daf/lf260  SURVIVES  [fi, nl]

`LF260F -> LF260` was already in the DAF block. Renames are not transitive, so
the moment `LF260` became a fold key those rows landed on the intermediate and
stopped — and `daf/lf260` lives on as a *different, smaller* record (fi+nl
instead of es+fi+nl). That is nastier than #176's version: a check that only asks
"did the id list shrink?" passes while the id survives.

`Fascf`/`Fatcf -> CF85` needed **no** change, precisely because CF85 is kept. The
trap and the chain are the same fact seen twice.

## Correction to data#176's permanent documentation

`data#176`'s body states the rename-side chain shipped with *"green build, no
lint, no gate — caught only by diffing control-vs-treatment id lists"*. **That is
wrong, and it is wrong in permanent documentation.** I reconstructed #176's
pre-flatten state on current main and ran the lint:

    10 failure(s) — 5x former_ids CHAIN, 5x renames DIRECTION WAR

`lint_curation`'s direction-war rule landed in `data#26` on 2026-07-26, **six days
before #176**. Both halves were always catchable pre-build. The id diff remains
necessary for what the lint genuinely cannot see — a key that fires on nothing,
and cross-kind spill — but it is not the only instrument, and the next reader
should not skip the cheaper one.

## Gates

`lint_overrides`, `lint_curation --base=origin/main`, `reorg_make_blocks --check`,
`lint_plates`: all green. 26 `former_ids` aliases added, 1 repointed; no-vanish
clean, 0 id-contract failures. `daf/cf` now carries 19 aliases, `daf/lf` 13.

## Not coupled

Data-only. No pipeline change, no merge-order constraint.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -4581,7 +4581,7 @@
 "truck/daf/xf480f":                        "truck/daf/xf"                         # FLATTENED 2026-08-01 by the power-suffix fold (renames.yml DAF §X-1): was -> truck/daf/xf480, which is now itself an alias; XF480F => XF480, 16 vehicles. inbound 0 — nothing chains through this id
 "truck/daf/xf510f":                        "truck/daf/xf"                         # FLATTENED 2026-08-01 by the power-suffix fold (renames.yml DAF §X-1): was -> truck/daf/xf510, which is now itself an alias; XF510F => XF510, 3 vehicles. inbound 0 — nothing chains through this id
 "truck/daf/xf530f":                        "truck/daf/xf"                         # FLATTENED 2026-08-01 by the power-suffix fold (renames.yml DAF §X-1): was -> truck/daf/xf530, which is now itself an alias; XF530F => XF530, 5 vehicles. inbound 0 — nothing chains through this id
-"truck/daf/lf260f":                        "truck/daf/lf260"                      # FOLD onto a LIVE survivor (es+fi+nl, already covers the child's fi+nl); LF260F => LF260, 8 vehicles. inbound 0 — nothing chains through this id
+"truck/daf/lf260f":                        "truck/daf/lf"                         # FLATTENED 2026-08-02 by the LF/CF power-suffix fold (renames.yml DAF §X-2): was -> truck/daf/lf260, which is now itself an alias; LF260F => LF260, 8 vehicles. inbound 0 — nothing chains through this id
 "truck/daf/fa45":                          "truck/daf/45"                         # FOLD onto a MINTED survivor (fi+gb+nl); FA45 => 45, 1911 vehicles. inbound 0 — nothing chains through this id
 "truck/daf/fa55":                          "truck/daf/55"                         # FOLD onto a LIVE survivor, which GAINS es+fi (gb+nl -> es+fi+gb+nl); FA55 => 55, 317 vehicles. inbound 0 — nothing chains through this id
 "truck/daf/ad85":                          "truck/daf/85"                         # FOLD onto a MINTED survivor (fi+nl); AD85 => 85, 12 vehicles. inbound 0 — nothing chains through this id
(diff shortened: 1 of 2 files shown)


rameerez:
ADJACENT FINDING, deliberately NOT shipped here — `truck/daf/530`.

While bounding the LF/CF families I checked the two bare-number DAF truck ids that sit next to them. `truck/daf/530` (fi, nl) is the **same power-suffix defect in its worst form**: a bare horsepower number with the series badge gone entirely. Its nl_rdw raws:

```
2  TB 160 DS 530     1  FAC 530     1  FT-95XF-530
2  TB 160 DT 530     1  FAD 530     1  XF 530
                     1  FAQ 530     1  XG+ 530
                     1  FAW 530     1  XG-SERIEN FTS 530
```

That pools **at least four different series** — 95XF, XF, XG+ and XG — onto one id, so there is no single honest survivor and it is NOT a fold. It is the `Oprij` shape from §A-D4: a NAMING §8.1 mixed-block case needing a `removals.yml` manifest (or per-raw keys), with its own measurement.

Out of scope for this PR by construction: it is neither `lf*` nor `cf*`, it needs a different remedy from a fold, and a fold decided twice is worse than a fold deferred. **Filing it rather than shipping it.** `truck/daf/230` is a separate thing again (fed by `MB 230`, plus uk_dft) and is not this class — unverified, left alone.

PR #205, opened 2026-08-02
curation: the data half of D-2 (Opel electric split) — one styling pin, four annotations, 13 tripwires

rameerez:
Pairs with the pipeline's ELECTRIC_NAMEPLATES change and MUST MERGE SECOND.
No fold key is added and none is deleted: the split is a normalizer change,
because collapse_variant's -[A-G] strip ran BEFORE the rename lookup and no line
in this repo could ever have reached those rows.

overrides/styling.yml — ONE pin, and it is load-bearing, not cosmetic.
AMPERA-E: Ampera-e. The normalizer pin makes collapse_variant emit "Ampera-e"
for 2,856 of car/opel/ampera-e's 2,861 vehicles, and smart_case turns that into
"Ampera-E" (measured). Today's correct display comes from the renames.yml
`Ampera EV: Ampera-e` VALUE used verbatim — the path the pin now bypasses.
Measured on a pipeline-only build: without this line the published display
regresses to "Ampera-E" with ALL GATES GREEN, so nothing else would catch it.
Blast radius 1 — car/opel/ampera-e is the only id in dist/vehicles.csv whose
model_name upcases to AMPERA-E. Slug unchanged either way. The other four pin
values need no styling entry: "Combo Electric"/"Mokka Electric"/"Movano
Electric" survive smart_case unchanged, and "Rocks-e" round-trips to "Rocks-E",
exactly what moped/opel/rocks-e publishes today.

overrides/models/renames.yml — annotations only, in the Opel block.
* `Corsa E: Corsa` — CONFIRMED AND WIDENED. The dossier asked for the
  hyphenated CORSA-E to be split off as "Corsa Electric" (37,469 vehicles, its
  largest item). It is the same generation E: nl_rdw's 32,533 first-register
  2014-2019 and ZERO after, beside that register's own CORSA-A/-B/-C series;
  fi_traficom's 4,794 are 99.4% pre-2020. A car still on sale cannot have a
  curve that stops in 2019. So CORSA is absent from the pipeline pin, this key
  stays load-bearing, and former_ids.yml "car/opel/corsa-e" keeps the published
  meaning PR #97 gave it. Nothing in former_ids.yml or removals.yml changes.
* `Ampera EV: Ampera-e` — NOW INERT (2 probing rows -> 0, measured by a
  probe-count diff over all 377,986 corpus rows). KEPT, not deleted: flap
  insurance if the pin is ever narrowed, and retiring a key needs its own
  dead-key replay (the wave-7 `Astra Twin Top` precedent).
* `Combo Cargo: Combo` and `Combo Life: Combo` — comments CORRECTED. Both cited
  the "COMBO-E CARGO" raws as their evidence; those raws are the electric van
  and now resolve ahead of this lookup. Both keys stay LIVE on the ICE
  spellings (8->4 and 3->2 probing rows), so neither is dead.

spotchecks.yml — 13 rows. Renames are make-scoped but KIND-BLIND and the pin
fires on car AND van, so both kinds are asserted, and the block includes the
NEGATIVE ruling, which is the most important part of it:
* opel/corsa-electric `exists: false` in BOTH car and van — the assertion that
  the 37,300-vehicle generation-E reading is never re-litigated silently.
* opel/corsa keeps all nine countries; opel/kadett and opel/rekord keep theirs
  (the -E-is-a-generation-letter collision tripwires).
* opel/combo-electric fi|lu|nl and opel/mokka-electric fi|ua — the two ids the
  split mints and publishes.
* opel/combo, opel/mokka, opel/ampera and opel/movano keep every country.
* opel/ampera-e must carry es and ua — they only arrive if the pin fires.
* opel/rocks-e (moped) — untouched, and the tripwire for the day
  PROPOSAL-kind-boundary Q2 moves L6e/L7e into :car.
* vauxhall/mokka — DECLARED EVIDENCE LOSS under NAMING §2, asserted in BOTH
  directions (availability_includes gb,ua + availability_excludes nz). It
  published gb|nz|ua and its ONLY nz row was nz_nzta "VAUXHALL MOKKA-E" x1, a
  Mokka Electric. That row moves to car/vauxhall/mokka-electric, which is
  single-source at 1 vehicle and stays a CANDIDATE, so nz leaves the published
  catalog. The ICE Vauxhall Mokka has no nz evidence, so the reading is right —
  but it is a loss, and availability_excludes stops a silent re-merge restoring
  the wrong claim.

VERIFICATION — control and treatment from the same pinned base (f9d5637 data,
6bbd4ee pipeline), pristine control worktree, frozen cache with licence files
excluded, manifest verified byte-stable across both builds.
  car 4967->4968, van 618->619, motorcycle/moped/truck/bus unchanged.
  NEW 2, DISAPPEARED 0, availability changed 2. ALL GATES GREEN both sides.
  lint_overrides OK · lint_curation OK (109 files) · reorg_make_blocks --check
  already alphabetical · lint_plates OK · 14 pipeline suites, 0 failures.
Merge order proved by measurement: this PR alone fails 5 spotchecks, because
four of them assert records only the pipeline change creates.

---

**MERGE SECOND**, immediately after vehiclesdb/vehiclesdb-pipeline#140. Measured both directions:
| order | result |
|---|---|
| this PR alone | **RED** — 5 spotcheck failures (4 assert records only pipeline#140 creates) |
| pipeline#140 alone | GREEN, but `car/opel/ampera-e` display silently regresses `Ampera-e` → `Ampera-E` |
| both | GREEN, display correct |


rameerez:
Verifier classification: sole failure = ua_mvs licence UNVERIFIED (data.gov.ua fetch failed on the runner; honest-gate class, zero relation to this PR's content — the standing classify-and-proceed rule for disposition PRs applies). The AMG spotcheck cleared on this run, confirming #206's amendment. Merging. — owner

PR #206, opened 2026-08-02
curation lockstep for pipeline#144 (uk_dft 4W Model-column router)

rameerez:
Lockstep half of **vehiclesdb-pipeline#144** (`uk_dft` 4W Model-column router, G-1c). That PR has the measurement, the design and the full control-vs-treatment evidence; this one is the curation the routing change makes true or false.

## ⚠ MERGE AFTER pipeline#144 — proven, not asserted

I built **this branch against pipeline `main`** (i.e. what the data repo's CI actually does — it runs the pipeline at its *default branch*):

```
FAIL  id-contract gate (liveness): van/mercedes-benz/g-class is ALIVE in this build
      yet aliased to car/mercedes-benz/g-class — an alias may never name a live id
1 gate failure(s)
```

With pipeline#144: **0 gate failures.** Control (main+main): **0**. Base `f9d5637` + `6bbd4ee`, frozen cache (`find cache -type f ! -name 'license_*.txt' -exec touch {} +`), VEH0120 2026 Q1.

---

### 1. `spotchecks.yml` — `mercedes-benz/amg`: `availability_includes: [gb, nz]` → `[nz]`

The row called itself *"the tripwire against that cleanup"*, and its own author saw this coming — *"uk_dft's GenModel aggregate 'MERCEDES AMG CLASS' (gb 57,573, a range label spanning every AMG model, **which the file's Model column would split**)"*.

It has been split. All **198** Model values decompose to **19 already-live records**, **57,573 / 57,573** gb vehicles, **zero** ids minted, **zero** unresolvable strings. Reproduced three ways: the owner swarm's two independent derivations (`aux/research/2026-08-owner-swarm/item11-amg-residue.md` §4, Ruby vs hand-transcribed Perl, diffed row by row), pipeline#144's own replay, and a known-answer case pinned in `pipeline/tests/test_uk_dft_pools.rb`.

**This is an amendment, not a bulk edit to go green.** A spotcheck is amended when the thing it guards has deliberately changed, and the change is stated in the row. What is *retained*: `nz`, and the refusal to null the stub — the bare `AMG` rows from `nz_nzta`/`nl_rdw` (nz 373) carry no model string to decompose and still have no honest target. The record keeps ar/ca/es/fi/lu/nl/nz/th/us and is **not retired**.

### 2. `former_ids.yml` — RESTORE `"van/mercedes-benz/g-class": "car/mercedes-benz/g-class"`

Removed on 2026-08-01 because the AMG sub-brand fix pushed car dominance to 96.75%, under the 97% `cross_kind_prune!` bar, so the van record survived. **That line predicted this reversal in writing:**

> NOTE THE MARGIN — 0.25pp. If a later build pushes car dominance back over 97% the van record prunes again and will then need this alias back PLUS a removals.yml manifest.

pipeline#144 sends 2,702 gb `AMG G 63 …` registrations onto `car/mercedes-benz/g-class`. Measured, control vs treatment: **11,866/(11,866+399) = 96.7468% → 14,568/14,568 = 100.00%**. The van record prunes again and the alias is legal again (`validate.rb:441` only forbids aliasing a **live** id).

**One correction to the prediction: no `removals.yml` manifest alongside it.** The two are alternatives, not a pair (`validate.rb#no_vanish` accepts an alias *or* a removal), and the repo rule is never both. The alias is the right one of the two because the pipeline's verdict is still *"this IS the car, misfiled by body type"* — there is a single honest successor, which is precisely the case a manifest is **not** for.

Worth stating: the gate never engaged in either direction. `van/mercedes-benz/g-class` is **not** in `catalog/van/models.json` at v2026.08.0, so no consumer holds it. This line is defensive — it costs nothing and covers the case where a release ships between the two PRs and publishes the record just before #144 retires it.

### 3. `former_ids.yml` — the two ledger comments that cited the vanished 96.75% margin

Both now carry the **third state** rather than being rewritten to look like they were always right. Three states in three days is not churn: each is the pipeline's own dominance verdict under the evidence it had, and the alias is why consumers never see the difference.

### 4. `renames.yml` + `removals.yml` — the `M Series` / `car/bmw/m-series` cost citation

Both cited *"uk_dft GenModel 6,112"* as vehicles the removal **costs**. It no longer does. `removals.yml` said:

> COST, stated plainly: this drops the 6,112 gb + 107 nz vehicles that carry only the group label … a row that says nothing more than 'M SERIES' cannot be assigned to M2 rather than M5.

The gb rows never said "nothing more than M SERIES" — the Model column names the car: `M135 XDRIVE AUTO` 3,765 · `M5 PHEV AUTO` 1,209 · `M235 XDRIVE AUTO` 730 · `M760E XDRIVE AUTO` 213 · `M3 CS XDRIVE AUTO` 137 · `M4 CS XDRIVE AUTO` 58. #144 reads it, so all **6,112 land on live BMW records** instead of being discarded. The removal itself is unchanged and still right for nz.

**The key stays LIVE** — nz_nzta's 107 model-less rows still produce it. Verified against a build. `test_override_key_reachability` would **not** have caught it either way: it is kind-blind and accepts a key if any kind reproduces it.

### 5. `styling.yml` — `CLE: CLE` whole-string pin

`car/mercedes-benz/cle` publishes as **"Cle"** while its entire cohort (`CLA`, `CLS`, `GLC`, `GLE`, `GLS`) is already an `acronyms:` token. Proposed by the owner swarm as DRAFT-2; **re-verified here against a fresh six-kind build rather than inherited**:

- blast radius **1** — `car/mercedes-benz/cle` is the only record catalog-wide whose name contains the token `Cle` (car+van+truck+bus+motorcycle+moped);
- **no** key in `renames.yml` / `moves.yml` / `aliases.yml` / `body_types.yml` contains the token, so nothing goes silently inert;
- slug unchanged — `slugify("CLE") == slugify("Cle") == "cle"` — no id churn, no retirement, no alias.

Whole-string pin per the `XDC`/`XFC` precedent two lines below it, deliberately **not** an `acronyms:` token (NAMING §9 wants a full-catalog sweep for those; the ES-token orphaning recorded on the `480 ES` line is what skipping one costs). Raised now because #144 puts **680 more gb vehicles** behind the misspelling; the defect predates it.

**This pin nearly shipped a silent 680-vehicle regression**, and the fix is in #144: `FW_POOLS` declares **both** `"Cle"` and `"CLE"`, because an undeclared nameplate *falls back to the stub* rather than failing. Declaring one spelling would have made the whitelist order-dependent across two repos. `test_normalizer.rb`'s AMG case also stopped pinning the casing — it failed exactly that way on the first `rake test` of the pair, on an assertion about the sub-brand strip, which is casing-independent.

---

## Verification

- `scripts/lint_overrides.rb` **OK** · `scripts/lint_curation.rb` **OK** (109 files).
- Full build with pipeline#144: **0 gate failures**, 0 new ids across all six kinds, 1 removed (`van/mercedes-benz/g-class`, aliased above).
- Full build with pipeline **main**: **1 gate failure**, the alias-liveness one quoted at the top. That is the merge order, demonstrated.
- Pipeline suite green with **and** without this branch (`rake test`, both `VDB_DATA_REPO`s).

## Merge order and the concurrency caveat

**pipeline#144 first, then this.** And **before** the nine concurrent make-scoped curation PRs — #144 changes which raw gb strings reach the normalizer for every make, so a curation key written against pre-G-1c strings may not survive it (that is exactly what happened to the `M Series` cost citation above). Cheaper to rebase curation onto a settled string set than to derive it twice.

**#144's dead-curation audit is valid as of `f9d5637` and must be re-run at merge time** — it enumerates keys against a snapshot of `overrides/` while nine delegates are writing to those files.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


rameerez:
## ⚠ The red `build` check is the merge order, reproduced by CI — do not "fix" it here

```
NormalizerTest#test_mercedes_amg_subbrand_is_stripped_not_published
  [pipeline/tests/test_normalizer.rb:126]:
Expected: ["Mercedes-Benz", "Cle"]
  Actual: ["Mercedes-Benz", "CLE"]
```

This job runs **the pipeline at its default branch**, so it is testing this branch's `styling.yml` against `test_normalizer.rb` as it exists on pipeline `main` — where that assertion still pins the old spelling. The `CLE: CLE` pin (line 5 of the description) makes `classify("MERCEDES-BENZ", "MERCEDES-AMG CLE 53")` return `"CLE"`, which is the point of the pin.

**pipeline#144 already contains the fix**, and it is not "update the expected string": the assertion is about the AMG **sub-brand strip**, which is casing-independent, so #144 asserts it casing-independently and stops the line being a cross-repo tripwire for a cosmetic it does not guard. I hit this locally on the first `rake test` of the pair, which is how it got into #144.

So: **merge pipeline#144, then re-run this check.** Verified locally, frozen cache, base `f9d5637` + `6bbd4ee`:

| pipeline side | data side | result |
|---|---|---|
| main | main | 0 gate failures, `rake test` green *(control)* |
| main | **this branch** | ❌ `test_normalizer.rb` (above) **and** ❌ `id-contract gate (liveness): van/mercedes-benz/g-class is ALIVE in this build yet aliased to car/mercedes-benz/g-class` |
| **#144** | main | 0 gate failures, `rake test` green |
| **#144** | **this branch** | **0 gate failures**, `rake test` green, 0 new ids across all six kinds |

Two independent lockstep failures, both predicted in the description before CI ran, both fixed by the same ordering. The `lint` check — which does not depend on the pipeline — passes.

rameerez:
**S2W — the `M635 CSI` gap pipeline#144 noticed cannot be closed with a rename key. I verified the mechanism; it needs a normalizer exception or a router carve-out.**

BMW is my block, so this is mine to fix — but I want the shape right before anyone writes a key, because the obvious key is catastrophic.

Ran the real normalizer against the real overrides:

```
M635 CSI    -> "6 Series"     <-- what the router will produce
M635CSI     -> "6 Series"
M635 CSi    -> "6 Series"
M 635 CSI   -> "M635CSi"      <-- the SPACED form, which is why the curated key works
```

**Why a rename cannot intercept it.** `normalizer.rb#bmw`'s `/^M?(\d)\d\d/` fires on the unspaced form and resolves it to `6 Series` **before the rename layer is consulted**. The existing curated key `"M 635 Csi": M635CSi` works only because the *spaced* string escapes that regex (`M` then a space, so `(\d)\d\d` cannot match at position 1).

So the only rename key that could catch the router's output would be keyed on **`"6 Series"`** — which would fold the entire 6 Series into `m635csi`. That is the direction-war shape at catalogue scale.

**Two honest options, both normalizer-side:**

1. **Exception before the series rule** — `return "M635CSi" if m =~ /\AM635\s*CSI/i`, and the same shape for `M535`/`M535i`, which has the identical problem (`renames.yml` keys `"M 535"` and `"M 535 I"` only). Narrow, two closed badges, no blast radius beyond BMW.
2. **Router carve-out** — pipeline#144 does not route `BMW M6`/`BMW M5` rows whose Model begins `M635`/`M535`, leaving them on the GenModel path they take today.

I lean (1): it fixes the string wherever it appears rather than only on the path that currently produces it, and it is inside the make block I own.

**Impact if neither lands with #144:** 261 gb vehicles route to `car/bmw/6-series` instead of the curated `car/bmw/m635csi` — a KEEP decision with its own `former_ids` arms, quietly reversed. Small, but it is exactly the class of silent reversal the curated keeps exist to prevent.

Say which you prefer and I will write it; if you would rather not block #144 on it, I can land (1) as a standalone pair immediately and you merge in either order — it is safe on today's main because the unspaced string does not occur yet.

PR #208, opened 2026-08-02
hyundai: retire the i80 stopgap the regex fix makes inert, and pin the truncation

rameerez:
hyundai: retire the i80 stopgap the regex fix makes inert, and pin the truncation

Second half of a COUPLED PAIR. Merge vehiclesdb/vehiclesdb-pipeline FIRST.

`renames.yml` carried `Hyundai: "i80": "i800"` — a stopgap whose own comment
said what to do next: "Stopgap for the pipeline regex fix (filed); retire with
a comment when that lands". It has landed. `normalizer.rb#hyundai` now reads
`/^I\s?(\d+0)(?!\d)/` and returns "i800" for raw "I800" directly, so nothing
downstream ever produces the nameplate "i80" again and the key is corpus-INERT.

## Why the key must go rather than be left as a documented dead key

The reachability test cannot see this one. `test_override_key_reachability`
replays the KEY as a raw string, and "I80" still produces "i80" under the
fixed regex — so the key would report GREEN FOREVER while being probed by
zero corpus rows. That is the dead-curation trap in its purest form: a check
that says a line is fine because the line can be reached in principle, not
because anything reaches it in fact.

The `Astra Twin Top` disposition (wave-7, in this same file) is the right
precedent, not QASHQAI+2: QASHQAI+2 was kept because it has a live SIBLING
key mapping to the same target. "i80" has none.

## Dead-curation audit — corpus probe of ALL 47 keys in the Hyundai block

Method: wrap `@o.model_renames["Hyundai"]` so `key?` records every nameplate
the pipeline actually hands to the lookup (report_junk_drops.rb's ProbeBlock —
zero behavioural change), then replay the whole corpus, all six kinds.

BEFORE (pipeline 6101c2c + data 268a6ca), 47 keys:
  "i80"  -> "i800"   LIVE, 7,967 veh, car+van, gb+nz, sole raw "HYUNDAI I800"
  "I45"  -> i45      LIVE, 1,462 veh, car, nz, raw "I45"
  DEAD: "Inster 01 EV", "Kona Advance HEV S"

AFTER (both halves), 46 keys:
  "i80"              GONE
  "I45"  -> i45      LIVE, 1,462 veh — byte-identical probe
  DEAD: "Inster 01 EV", "Kona Advance HEV S"  (unchanged)

The probe-report diff between the two runs is ONE line: the removal of "i80".
Every other key's vehicle count, kind set and country set is identical. No
new inert key, and no key lost its rescue.

The two remaining DEAD keys are PRE-EXISTING and deliberate, not defects:
both are UK DfT genmodel-drift folds whose comments say they exist to keep a
retired id dead in EVERY cache state (CI pre-drift vs local post-drift, the
#67 liveness lesson). Zero probes in the current cache state is their
designed condition. Left alone.

## Other stopgaps for the same rule

`"I45": i45` is the only other one, and it is STILL LIVE after the fix —
`(\d+0)(?!\d)` requires the digit run to end in 0, so "I45" still falls
through to it. The pipeline half explicitly declined the wider `(\d+)` for
this reason. Its comment now records that it survives and why.

## The published id is not at risk

`car/hyundai/i80` retired into `overrides/models/former_ids.yml:3839`
("car/hyundai/i80": "car/hyundai/i800"), a separate file this change does not
touch. No `removals.yml` entry exists for it and none is wanted — alias OR
manifest, never both.

## Spotchecks: the tripwire, and proof it is armed

Two rows, one assertion in two halves. `exists: false` is the load-bearing
one: an `_includes` row cannot say "and the wrong name must never return",
which is how this defect stayed invisible.

Built the WRONG merge order on purpose (this half against unpatched pipeline
main) to check the tripwire is not vacuously passing. It fires:

  FAIL spotcheck[...]: hyundai/i800 (car) missing from catalog
  FAIL spotcheck[...]: hyundai/i80 should NOT exist
  FAIL id-contract (no-vanish) car: car/hyundai/i800 — no former_ids alias...
  FAIL id-contract (liveness): car/hyundai/i80 is ALIVE yet aliased to i800
  FAIL id-contract (liveness): car/hyundai/i80 -> i800: target is not live

16 gate failures -> 21. In the CORRECT order the failure set is byte-identical
to the control build's 16 (all pre-existing yamaha/daf id-contract failures on
main from other sessions' in-flight work — this pair neither adds to nor fixes
them).

## Measurements

Frozen cache (`find cache -type f ! -name 'license_*.txt' -exec touch {} +`;
the licence exclusion matters — a blanket touch silently disables the licence
gate). Control from the SAME base in a PRISTINE worktree, and every replay
output carries a provenance stamp (both repo HEADs, both dirty states, the
verbatim regex, whether the i80 key is present) so a control secretly pointed
at the treatment is impossible to miss.

  paired end state vs control: 0 vehicles move, 0 of 14,027 ids change
                               (six-kind build diff: id, name, availability,
                                former_ids — all identical)
  this half ALONE vs control:  7,967 vehicles regress to Hyundai/i80

Lints green: lint_overrides OK, lint_curation OK (109 files), reorg
--check OK, lint_claims OK. lint_dataset's 15 name-shape suspects are
byte-identical to data main's and contain no Hyundai record.

Pipeline half: https://github.com/vehiclesdb/vehiclesdb-pipeline/pull/146 — MERGE THAT FIRST.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


rameerez:
**CI status — read before treating the red as mine.**

`lint` **pass**. `build` **fail**, and the failure is not this PR:

```
TestUkDftPools#test_amg_class_reproduces_the_owner_swarm_decomposition
  [pipeline/tests/test_uk_dft_pools.rb:206]
  -{"A-CLASS" => 16759, "C-CLASS" => 12473, ...}
  +{"A-CLASS" => 15995, "C-CLASS" => 12459, ...}
```

A Mercedes-AMG pool decomposition. This PR touches three things, all Hyundai: one `renames.yml` key inside the `Hyundai:` block, and two `spotchecks.yml` rows for `hyundai/i800`/`hyundai/i80`. Nothing it changes can move a Mercedes count.

Three sibling data PRs on unrelated makes fail on the **identical** test in the same window — `s4w/bodyword-92-class` (run 30725663517), `s4w/uk-model-4w-data` (30725503084), `s4w/opel-electric-split-data` (30725501995). It is a repo-wide red on the Build workflow, which checks out pipeline **main**; the AMG test's expected pools have drifted from what main now produces.

Locally, with the frozen cache and the same base, this pair's build failure set is **byte-identical** to the control build's (16 pre-existing yamaha/daf id-contract failures), and the spotcheck gate passes. The wrong-merge-order build is the only configuration where anything Hyundai-shaped fails, which is the point of the two new rows.

I have not touched the AMG test — it is another session's artifact and outside this task's make scope.

rameerez:
**Re-verified on an advanced base, and the earlier red is gone.**

The base moved twice under this PR while it was open. Everything has been re-measured against the current mains — pipeline `3d57846`, data `7e03a40`, both controls PRISTINE at exactly those commits, frozen cache — and **every number is unchanged**:

| | vehicles moved | six-kind id diff |
|---|---|---|
| pipeline half alone vs unmodified data main | 0 | 0 of 13,869 |
| **this half alone vs unmodified pipeline main** | **7,967 onto `Hyundai/i80`** | `car/hyundai/i800` REMOVED, `car/hyundai/i80` ADDED |
| paired end state | 0 | 0 of 13,869 |

Both the control build and the paired build now exit **0 — ALL GATES GREEN** (the AMG spotcheck that was failing every data PR in the queue was fixed by pipeline#150, which is in main now). The two builds' failure sets are identical, which is to say both empty.

Dead-curation audit re-run on the new base, all 47 Hyundai keys, corpus probe:
before 47 keys with `"i80"` live on 7,967 veh; after 46 keys, `"i80"` gone, `"I45"` still live on 1,462 veh with a byte-identical probe. **The entire probe-report diff is one line — the removal of `i80`.** The two remaining zero-probe keys (`Inster 01 EV`, `Kona Advance HEV S`) are pre-existing, documented DfT cache-state insurance, and are untouched.

Merged `origin/main` in rather than rebasing: AGENTS.md forbids force-push everywhere, and a rebase of an already-pushed branch cannot land without one. Verified the merge result's tree against the rebased tree I had measured — identical apart from the four upstream commits, none of which touch `pipeline/lib/` or any Hyundai line.

rameerez:
**Expect this PR's `build` check to be RED until vehiclesdb/vehiclesdb-pipeline#146 merges. That redness is the coupling, not a defect.**

`.github/workflows/monthly-build.yml:87-92` checks the pipeline out with **no `ref:`**, so it builds this branch against the pipeline repo's DEFAULT BRANCH:

```yaml
- name: Checkout pipeline
  uses: actions/checkout@v4
  with:
    repository: vehiclesdb/vehiclesdb-pipeline
    path: pipeline-repo
```

That is precisely the "data half alone against unpatched pipeline main" configuration — the one measured above as 7,967 vehicles landing on the fabricated `Hyundai/i80`. Reproduced locally in that exact configuration, the build raises six gate failures, five of them mine by design:

```
FAIL spotcheck[...]: hyundai/i800 (car) missing from catalog
FAIL spotcheck[...]: hyundai/i80 should NOT exist
FAIL id-contract (no-vanish) car: car/hyundai/i800 — no former_ids alias and no removals.yml entry
FAIL id-contract (liveness): car/hyundai/i80 is ALIVE in this build yet aliased to car/hyundai/i800
FAIL id-contract (liveness): car/hyundai/i80 -> car/hyundai/i800: target is not live
```

With #146 in, the same build exits **0, ALL GATES GREEN**, and the six-kind id diff against a pristine control is **0 of 13,869**.

`lint` — the check that measures this repo on its own terms — passes.

This is the cross-repo tripwire that `test_normalizer.rb`'s CLE comment documents ("the data repo's CI runs the pipeline at its DEFAULT BRANCH, so whichever side merged second went red on an assertion about something it was not testing"). Here it is not an accident of casing: it is the ordering constraint itself, and I would rather it be loud.

rameerez:
**Update: pipeline#146 merged (82a817b, by the owner, 01:35Z) and this PR's Hyundai assertions now all pass on CI.**

The build was re-run after the merge. Result:

* **Zero Hyundai failures.** Both new spotchecks pass, and the three Hyundai id-contract failures are gone. `grep -i hyundai` over the failed-step log returns nothing.
* The run that failed *before* the merge is the receipt for the ordering: created 01:28:25Z, it checked the pipeline out at 01:35-minus, and produced **exactly** the five failures predicted in the PR body — the two spotchecks, the no-vanish, and both liveness failures. The prediction and the CI hardware agree.

**The remaining red is not this PR.** The re-run's entire failure set is five `id-contract gate (xref-loss) truck: truck/daf/*` entries, from pipeline#147's brand-new gate 7d meeting the already-merged DAF power-suffix fold (data#199). The identical five failures are on **data `main` itself** — run 30728037076, branch `main`, sha 3bd1a31 — and on sibling branch `s4w/lancia-coupe-data`. It is a repo-wide red that predates and outlives this PR, and it wants the DAF fold's owner, not me: the gate's own message says to re-cut the fold or sign the loss off with `accepted_xref_loss:` in `former_ids.yml`.

`lint` passes. This PR touches one `renames.yml` key inside the `Hyundai:` block and two `spotchecks.yml` rows; nothing it changes can delist a DAF type approval.

Ready to merge whenever the DAF red is dispositioned — this half is no longer gated on anything of its own.

PR #209, opened 2026-08-02
body-word 92: retire 59 fake body-word ids, prove the 21 real nameplates untouched, and REPLACE the DEBT remedy measurement refutes

rameerez:
# The body-word 92 class: 59 fake nameplates retired, 21 real ones proven untouched, and the filed remedy replaced

Applies `aux/research/2026-08-owner-swarm/bodyword-92-class.md` (pipeline `2dc997a`).

> **RE-VERIFIED 3× as main moved under it.** Final measurement: data `20e0cbf`,
> pipeline `eab4ebe` (i.e. INCLUDING `pipeline#144`'s uk_dft Model-column router
> and `data#204`). Class size, per-source counts, id diff and the KEEP proof are
> unchanged across all three bases. §8 has the detail; §11 diagnoses the red CI.
**Not merged. Read §7 (conflicts) before merging — nine ids are deliberately deferred.**

---

## 1. RE-MEASURE — the dossier's disposition reproduces, its numbers do not

The dossier's enumeration was taken on the 2026.08.1 catalog, ~3 h before
`pipeline#131` and `data#183`. I re-derived from scratch on a **green control
build of `origin/main` (`5933b1b`)**, frozen cache, using the dossier's **closed
13-word vocabulary** (`coupe roadster sedan saloon convertible cabriolet
limousine cabrio estate hardtop mpv wagon van` — a looser one gives 132 ids,
which is a different class).

| | dossier (2026.08.1) | re-measure (`5933b1b`) |
|---|---|---|
| car ids in class | 92 | **92** |
| nl_rdw / nz_nzta / uk_dft / us_fueleconomy / ua_mvs | 84 / 82 / 14 / 12 / 10 | **84 / 82 / 14 / 12 / 10** |
| fi_traficom | 27 | **26** |
| es_dgt / lu_snca / ca_nrcan / th_dlt / my_jpj | 7 / 6 / 5 / 2 / 1 | **7 / 6 / 5 / 2 / 1** |
| disposition | 21 KEEP · 3 FOLD · 68 NULL | **21 KEEP · 3 FOLD · 68 NULL** |
| KEEP mass | 26,018 v | **49,837 v** |
| NULL cost | 2,184 v (stated as a floor) | **2,385 v** (complete) |
| non-car members | 4 | **4** |

**Three numbers moved and each has a cause.**

**(a) fi 27 → 26.** Only `mercedes-benz/cabriolet` lost Finland to `#131`. It now
publishes on `nl_rdw` alone, through the publication-hysteresis grace — the
control build says so in as many words: `reconcile car: hysteresis kept 2 edge
id(s): chevrolet/sedan-delivery, mercedes-benz/cabriolet`.

**(b) D-1 IS REFUTED — the other seven did NOT unpublish.** The dossier's
blocking prediction was that eight fi-corroborated ids fall below
`sources.size >= 2` once `#131` propagates. Seven still carry `fi_traficom`
today: `mercedes-benz/{coupe,roadster}`, `caterham/convertible`,
`cadillac/hardtop`, `packard/{coupe,sedan}`, `skoda/coupe`. **Cause**, read out
of the adapter rather than guessed: `#131`'s `BODY_HEAD` requires the body word
to be followed by a separator **and more text** (`fi_traficom.rb`, `BODY_HEAD`).
A **bare** `COUPE` row has nothing after it, so the strip never fires — in
Finland either. `#131` removed the *prefixed* shape, which is why
`mercedes-benz/cabriolet` (fed only by `CABRIOLET 300 SE AUTOMATIC`) moved and
the bare-word ids did not. The NULL count therefore stays 68, not 60.

**(c) KEEP mass 26,018 → 49,837, and the gap is one instrument error.** The
dossier's uk_dft figures match mine on 12 of 14 ids exactly. Two do not:
`mini/roadster` 352 vs **5,284** and `audi/cabriolet` 2,238 vs **2,565**.
`veh0120` is keyed on (BodyType, Make, GenModel, **Model, Fuel, LicenceStatus**)
and `MINI ROADSTER` spreads over eight such rows — `MINI ROADSTER COOPER` 2,372
Licensed + 104 SORN, `COOPER SD` 625+29, `COOPER SD AUTO` 112+1, … The dossier
read one row where the adapter sums them all. My figure is the pipeline's own
aggregate (census replays `source.rows(kind)` exactly as `reconciler.rb:240`
does). **The safety property is ~1.9× larger than filed.**

**Measurement method**, because "which string does a rename key have to be" is
the question that has produced inert keys all night: a whole-corpus replay of
production `classify` with a **recording proxy** wrapping `@o.model_renames`, so
every `renames&.key?(nameplate)` probe is captured **at the consultation site**
with mechanisms active and nothing disabled (the `gen_review_pack.rb` idiom).
That is what produced the keys below; they are the strings the pipeline actually
asks about, not the raw register strings.

---

## 2. What ships

**59 of the 68 NULLs — 2,029 vehicles.** `renames.yml` `null` key + `removals.yml`
entry per id. Every line carries its register, its column, its exact raw
string(s), its vehicle count, and the live-sibling count that makes the row
unassignable.

**4 keys of the 63 are not new ids — they are the RESURRECTION half**, and this
is the part a reviewer should look at first (§4).

**Plus 4 residue rescues (§B.3, corrected).** `Convertible De Ville`,
`Convertible Deville` → `De Ville`; `Convertible Eldorado` → `El Dorado`;
`Coupe 62` → `Series 62`. 7 vehicles, 4 candidate ids retired from the queue,
zero minted ids, zero minted spellings — every value is an already-published
record, matching the shipped `Coupe De Ville`/`Sedan De Ville` keys 70 lines up.

**Plus 2 chained-alias corrections** (§5).

**And the DEBT row's remedy text is REPLACED** (§6).

---

## 3. THE SAFETY PROPERTY — proof the 21 KEEPs are untouched

A blind sweep of this class destroys **49,837 vehicles of real nameplate**. The
proof is a full-record comparison, not an assertion:

```
survivors 13,819; changed in a field OTHER than popularity: 0
audi/coupe · audi/cabriolet · fiat/coupe · hyundai/coupe · maserati/coupe
smart/roadster · mazda/roadster · mg/roadster · morgan/roadster · mazda/mpv
volkswagen/cabrio · volkswagen/cabriolet · daimler/limousine · tesla/roadster
mini/roadster · mini/convertible · eunos/roadster · singer/roadster
triumph/roadster · mitsubishi/van · cadillac/limousine
  => 0 KEEP(s) disturbed        (each identical bar the popularity re-rank)
```

The re-rank is unavoidable and expected: 59 records leaving a 13,878-record
catalog moves everyone's rank by construction. **No KEEP changed `name`,
`sources`, `availability` or `body_types`.**

The four non-car members of the class are also untouched and still live:
`motorcycle/harley-davidson/roadster`, `motorcycle/norton/roadster`,
`motorcycle/triumph/roadster`, `truck/chevrolet/van`. That matters because
**make-scoped rename keys are KIND-BLIND** — see §4.

---

## 4. THE TWO SILENT-FAILURE CLASSES THIS BATCH WALKED INTO

Both were found by an instrument, not by reading. Recording them here so the
next reader does not rediscover them.

### 4a. Rename-VALUE resurrection (S2W Turn 242 / `data#197`) — LIVE in this class

Renames are single-pass. A pre-existing rename whose **value** is a retired
display name keeps firing, resolves to the dead display, and **republishes the
id under a live key**. Four such keys exist inside this class today, and all
four were shipped by my own earlier batches:

| key | was | shipped by | why it resurrects |
|---|---|---|---|
| `Cadillac: "2D Hardtop"` | `→ Hardtop` | `data#179` | target `cadillac/hardtop` retires here |
| `Cadillac: "4D Hardtop"` | `→ Hardtop` | `data#179` | same |
| `Ford: Raodster` | `→ Roadster` | `data#183` | target `ford/roadster` retires here |
| `Ford: Roadster V8` | `→ Roadster` | `data#183` | same |

All four are **rewritten in place to `null`**, not shadowed with a second key —
my first attempt added duplicates and `lint_curation` caught it
(`DUPLICATE KEY "Cadillac → 2D Hardtop" … YAML keeps only the LAST one`). That
lint firing on this batch is worth noting: it is a real defence, and it worked.

The generalisation, which I think is the durable finding: **`data#179` and
`data#183` rescued rows onto a fake nameplate.** Stripping a door count off
`2D HARDTOP SEDAN DE VILLE` is a rescue only if the target is real; here the
target was itself in the 92. A rescue batch should check that its fold targets
are not members of a class someone else is about to retire.

### 4b. Kind-blindness — measured, not assumed

Make-scoped keys apply to all six kinds. The complete blast radius of the 63
keys outside `car` is **two candidate rows**: `motorcycle/honda/coupe` (1 v,
a uk_dft 2W row of the same mis-kinded `HONDA COUPE` genmodel — Honda is
deferred anyway, §7) and **`van/dodge/van` 4 v nl_rdw**, which `Dodge: Van: null`
does kill. That is 4 vehicles of candidate-queue evidence, unpublished, and it
is stated rather than hidden. `truck/chevrolet/van` is safe because
`chevrolet/van` is not in the car-92 at all.

---

## 5. The chained-alias class — found by the gate, on the treatment build

The first treatment build came back with two failures the control did not have:

```
FAIL id-contract gate (liveness): car/ford/raodster    -> car/ford/roadster: target is not live
FAIL id-contract gate (liveness): car/ford/roadster-v8 -> car/ford/roadster: target is not live
```

Both ids were retired by `data#183` with a `former_ids` alias to
`car/ford/roadster` — which this change retires. The aliases became a promise to
migrate consumers **into a 404**. `former_ids` is not chased transitively, so
nothing but a real build sees this.

Fixed per SCHEMA.md (*alias to a live id **or** a removals entry, never both*):
the two `former_ids` lines are removed — with an in-file note saying why and
"do not restore these two lines" — and both ids are ledgered in `removals.yml`.
This is the mirror image of 4a: 4a is a rename whose **value** dies, 5 is an
alias whose **target** dies. **Neither is caught by `lint_curation`, and only 5
is caught by a gate.**

I re-ran both hazard sweeps against the newest `origin/main` after a mid-flight
rebase: still exactly these four keys and these two aliases, no new ones.

---

## 6. DEBT — the filed remedy is REPLACED, because measurement refutes it

The row prescribed *"the same shape of source-adapter strip for nl_rdw and
nz_nzta, each with its own grammar measurement."* That measurement was done
(dossier §C, frozen caches) and **neither register is body-first**: no door
count, no type code, no wheelbase tail. The body word is the *whole* string in
227 nl pairs / 1,491 v and 281 nz pairs / 18,579 v; where it leads, it is
overwhelmingly the first word of a real nameplate.

> A Finland-shaped strip rescues **401 vehicles** across nl+nz while putting
> **12,200 vehicles of real nameplate** behind a pin list longer than `#131`'s
> entire vocabulary. Finland's own ratio was 2,244 rescued / 136 pinned. **The
> ratio is inverted.**

The row now says, in the remedy column: **no strip for nl_rdw, nz_nzta, uk_dft,
us_fueleconomy or ua_mvs — this is CURATION**, keyed on the string the pipeline
*probes* (collapse_variant runs first), with the KEEP list protected and each
NULL carrying its evidence. It also records that **uk_dft is the class's
KEEP-DETECTOR, not a defect source** (13 of its 14 body-word ids are confirmed
real by its own `GenModel` column), and re-states the class with today's
numbers. `truck/mercedes-benz/umpikorinen`, which the old row named in the same
breath, is **outside** the 13-word vocabulary — a separate fi single-source
leftover.

---

## 7. CONFLICTS — nine NULLs and all three FOLDs are NOT shipped

**Eight belong to a make someone else owns or is mid-flight on. Reported, not edited.**

| ids | v | owner / PR | why deferred |
|---|---|---|---|
| `mercedes-benz/{sedan,coupe,roadster,cabriolet}` | 266 | `data#204` (open) | make mid-flight. Also needs 3 co-keys: `4D SEDAN200/2750`, `2D Roadster 190SL/2400`, `2D Roadster 450 Automatic 107/245` — the §4a hazard again. `cabriolet` is hysteresis-only now and may self-retire. |
| `opel/roadster` | 4 | `data#205` (open) | make mid-flight |
| `subaru/{sedan,wagon}` | 2 | `data#198` (open) | make mid-flight. The dossier's §A.4 trap is real and confirmed: EPA `baseModel` is literally `Sedan`/`Wagon` for Subaru, and only the contradicting-record check (`leone`, `dl`, `gl`, `1800`, `1400` all live) gets it right. |
| `honda/coupe` | 5 | **S2W** (`OWNERSHIP.yml honda: s2w`) | not my make |

**`lancia/coupe` (79 v) is REFUTED — do not null it.** The dossier called it NULL
on the `GenModel` column. The `Model` column disagrees: every one of its 61 gb
vehicles is a Beta Coupé —

```
Cars,LANCIA,LANCIA COUPE,COUPE 2000 VX,Petrol,Licensed/SORN   8 + 13
Cars,LANCIA,LANCIA COUPE,COUPE IE 2000,Petrol,Licensed/SORN   5 +  7
Cars,LANCIA,LANCIA COUPE,COUPE VOLUMEX,Petrol,Licensed/SORN   6 + 22
```

`lancia/beta` is live. **61 of 79 vehicles are recoverable**, and the instrument
that recovers them is exactly `data#206` / `pipeline#144` (uk_dft Model-column
router). Handing it there rather than deleting it.

**All three FOLDs are reported, not applied.** The union check the brief asked
for is done and **all three are lossless at (country, SOURCE, evidence) level —
zero pairs uncovered**:

```
smart/cabrio  nl:nl_rdw nz:nz_nzta ua:ua_mvs        -> smart/fortwo    NOT covered: []
smart/coupe   nz:nz_nzta ua:ua_mvs                  -> smart/fortwo    NOT covered: []
suzuki/wagon  nl:nl_rdw ua:ua_mvs                   -> suzuki/wagon-r  NOT covered: []
```

- **`suzuki/wagon`** — Suzuki is **S2W's** make (`OWNERSHIP.yml suzuki: s2w`) and
  `data#192` is open on the same `Suzuki:` block. Theirs to take; the union check
  above is the homework.
- **`smart/{cabrio,coupe}`** — I decline these on **new evidence the dossier did
  not weigh**. D-2 already flagged that they contradict `#131`'s SMART pin. The
  re-measure adds a second, harder objection: **`smart/city` and
  `smart/roadster-coupe` are both LIVE**. The W450 was sold as City-Coupé /
  City-Cabrio until the 2004 fortwo rename, so a bare `CABRIO` row is
  *ambiguous between two published records* — which is the dossier's own §A.1
  test for NULL, not the test for FOLD. Under DECISIONS.md ("before folding A
  into B, look for a published record that says otherwise") this fold does not
  clear. smart is also mid-flight on `data#200`. **Owner call, with better
  evidence than D-2 had.**

**One correction to my own reasoning, recorded per house rule.** I first
concluded from the probe that *all* of §B.3's draft residue keys were inert,
reasoning that collapse_variant destroys the second token. Wrong, and I checked
it instead of shipping it: **5 of the 9 are reachable**. `Hardtop Sedan De
Ville`, `Hardtop Sedan Deville`, `Van 6Cwt`, `Van 8Cwt` are genuinely inert
(the first two collapse to `Hardtop` before the rename site; the last two do not
exist in the corpus at all — and `morris/8cwt` is a live record, so the
dossier's "→ Minor" guess was unsupported anyway). `Sedan Delivery` already
reaches the live `chevrolet/sedan-delivery`, exactly as the dossier cautioned —
writing it would be an identity no-op. The remaining four are real and ship.

---

## 8. CONTROL vs TREATMENT

Same base (`5933b1b`), same **pristine** pipeline worktree, same frozen cache
(`find cache -type f ! -name 'license_*.txt' -exec touch {} +` — the licence
exclusion matters), separate `VDB_BUILD_DIR` and `VDB_SNAPSHOT_STORE` per side
so neither can read the other's artifacts.

| | control | treatment |
|---|---|---|
| gates | **1 failure** (see below) | **the SAME 1 failure, byte-identical** |
| published ids (6 kinds) | 13,869 | 13,810 |
| id diff | — | **−59, +0** |
| GONE set | — | **byte-identical to the intended 59** |
| survivors changed outside `popularity` | — | **0** |
| the 21 KEEPs | — | **0 disturbed** |
| non-car class members | 4 live | **4 live** |

`removals.yml` gains **61** entries — 59 retired ids + the 2 chained ids from §5.
No `former_ids` alias is written for any of them: **alias or manifest, never both**,
and no honest successor exists.

**The one gate failure is PRE-EXISTING and identical on both sides:**

```
FAIL spotcheck[AMG is a SUB-BRAND …]: mercedes-benz/amg availability
     ["es","fi","lu","nl","nz"] missing gb
```

`pipeline#144`'s `FW_POOLS` router splits the `MERCEDES AMG CLASS` GenModel, so
`mercedes-benz/amg` no longer receives gb — and the AMG spotcheck still asserts
it does. **Nothing in this PR touches Mercedes, AMG or uk_dft.** It reproduces
on a pristine control at the same commit, which is the whole point of running one.
It belongs to the AMG / uk-model-4w owners.

**Base discipline, stated because it bit others tonight.** I measured three
times. First on data `f9d5637`; `origin/main` then moved twice (once gaining 489
lines of `renames.yml`), so I re-based the control and re-ran both builds on
`5933b1b`. Then CI showed me the **pipeline** had moved too — +229 lines of
`uk_dft.rb`, +120 of `normalizer.rb`, +34 of `validate.rb` — which is exactly
the kind of drift that makes a curation measurement stale, so I moved the shared
pipeline worktree to `eab4ebe` and ran the whole thing a third time on data
`20e0cbf`. **Class size (92+4), per-source counts, the −59/+0 id diff, the GONE
set and the 0-KEEPs-disturbed proof are identical on all three.** Both §4a/§5
hazard sweeps were re-run after each rebase and returned the same 4 keys / 2
aliases.

Every artifact for this item lives under a private scratchpad namespace
(`bw92-priv/`, all helpers prefixed `bw92-`) after the shared-scratchpad
clobbering report — no generic `control.log`/`iddiff.rb` filenames were used.

---

## 9. Lints

```
overrides lint: OK
curation lint: OK (109 files, duplicate-key + shape + make-key + provenance)
make blocks already alphabetical: renames.yml, aliases.yml
plates lint: OK
```

`lint_curation` **failed first** on the four §4a duplicates and was fixed by
rewriting the originals in place. That failure is the reason to trust the rest.

---

## 10. Merge order

Standalone; no pipeline PR (the remedy is curation — that is the finding).

- **Merges cleanly against any of `data#192/198/200/204/205/206`** — this touches
  only makes none of them owns, by construction.
- Whoever merges last rebases; the `renames.yml` conflicts will be
  alphabetical-block adjacency only.
- If `data#206` (uk_dft Model router) lands first, **`lancia/coupe` should be
  re-examined by that agent**, not by this branch.

---

## 11. CI is RED, and it is not this PR

`build` fails on a **pipeline-repo known-answer test**, on every data PR whose CI
ran after ~00:35 UTC:

```
TestUkDftPools#test_amg_class_reproduces_the_owner_swarm_decomposition
  -{"A-CLASS"=>16759, "C-CLASS"=>12473, "GLC"=>5321, "CLA"=>4717, "E-CLASS"=>4339, …}
  +{"A-CLASS"=>15995, "C-CLASS"=>12459, "GLC"=>5306, "CLA"=>4637, "E-CLASS"=>4391, …}
```

Evidence it is not mine, in three parts:

1. **It reproduces on branches that share nothing with this one** —
   `s4w/hyundai-i800-data` (run 30725649076) and `s4w/opel-electric-split-data`
   (run 30725501995) fail on the same test at the same line.
2. **The same test PASSES locally** against my frozen cache: `1 runs, 4
   assertions, 0 failures`. The pin is on `uk_veh0120_uk.csv` and CI's copy is a
   **restated 2026 Q1** — every AMG figure moved a percent or two.
3. **This PR touches no Mercedes key, no uk_dft code and no test.**

Per I-11 (*when a TEST and the CORPUS disagree, the corpus wins*), the fix is to
re-derive the pin from the current file — **which I am deliberately not doing
here.** Re-deriving it requires fetching a fresh `uk_veh0120_uk.csv`, and that
cache is **shared by a dozen concurrent agents running frozen-cache control
builds**. Moving it under them would silently contaminate their comparisons,
which is a worse outcome than a red pin. Whoever owns the AMG table should
re-derive it in the pipeline repo, on their own cache, and ideally make the
assertion tolerant of a quarterly restatement (a share-of-total or a ±2 %
band) — this will recur every quarter otherwise.

**Consequence for this PR:** its `lint` job is green, its local control-vs-
treatment pair is clean, and its build failure is shared with the rest of the
queue. It should not be merged until that pin is fixed — but fixing that pin is
not this PR's job.

### And one status change worth flagging to a merger

**`data#204` (mercedes-residual) has MERGED since §7 was written**, so
`mercedes-benz/{sedan,coupe,roadster,cabriolet}` are no longer blocked by a
mid-flight sibling. I have **still not added them**, on purpose:
`mercedes-benz/sedan`'s 251 vehicles are almost entirely Finnish *pure-spec*
strings (`Sedan (AA) 4ov 2143cm3 A` ×11, `… 2148cm3 A` ×11, …) that `#131`'s
`SPEC_TOKEN` refuses **by design**. Whether those rows should be nulled at the
data layer or fixed at the adapter is a live pipeline question, not settled
curation, and the make's owner is better placed to answer it than a body-word
sweep. `mercedes-benz/cabriolet` is additionally hysteresis-only
(`reconcile car: hysteresis kept … mercedes-benz/cabriolet`) and may retire
itself. The three co-keys it would need are listed in §7.


rameerez:
## Reviewer's checklist — the five things worth re-deriving, with the exact commands

Posting these so a verifier can re-run the load-bearing claims without rebuilding my reasoning.

**1. The class is still 92 under the CLOSED vocabulary.** A looser vocabulary
gives 132 and every number here becomes wrong. Scan `catalog/<kind>/models.json`
for `slug` exactly in
`coupe roadster sedan saloon convertible cabriolet limousine cabrio estate hardtop mpv wagon van`
across all six kinds → 92 car + 4 non-car. Anything else means a different class
was measured.

**2. The rename keys must be the strings the pipeline PROBES, not the raws.**
`collapse_variant` runs before the rename lookup, so `CADILLAC | HARDTOP SEDAN
DE VILLE` arrives as `Hardtop`. Every key in this PR came from a recording proxy
wrapping `@o.model_renames` at the `renames&.key?(nameplate)` site — zero
behavioural change, the `gen_review_pack.rb` idiom. If you re-derive keys from
raw strings you will write inert keys; that is exactly what §B.3 of the dossier
did and why 4 of its 9 draft keys are dead.

**3. `#131` does NOT strip bare body words** — `BODY_HEAD` in `fi_traficom.rb`
requires a separator **and** more text after the body word. This is the whole
reason D-1's prediction failed and the NULL count is 68 rather than 60. Read the
regex, not the PR title.

**4. Two hazard sweeps, one grep each, and they must be re-run at merge time**
(both are keyed on ids this PR retires, so a sibling PR merging first can add a
new one):

```ruby
# rename VALUE names an id I retire → single-pass renames RESURRECT it
renames.each { |mk, blk| blk.each { |k, v| hit if slug(mk)/slug(v) ∈ retired } }
# former_ids TARGET is an id I retire → alias migrates consumers into a 404
former_ids.each { |old, to| hit if to ∈ retired }
```

On `5933b1b` these return exactly the 4 keys and 2 aliases handled in §4a/§5 and
nothing else. `data#197` ships the first as a script.

**5. Judge the local build by DIFF, never by exit code.** Control and treatment
are both `ALL GATES GREEN` here, but the honest check is the id diff (−59/+0,
GONE set byte-identical to intent) and the KEEP proof (0 of 21 disturbed,
0 survivors changed outside `popularity`). Both builds used separate
`VDB_BUILD_DIR` **and** `VDB_SNAPSHOT_STORE` so neither could read the other's
artifacts, and the cache was re-frozen with the licence exclusion before each.

---

### The one number I would most like a second derivation of

**`mini/roadster` gb = 5,284**, against the dossier's 352. Mine sums every
`veh0120` row under GenModel `MINI ROADSTER` across (Model, Fuel,
LicenceStatus) — eight rows, `MINI ROADSTER COOPER` 2,372 Licensed + 104 SORN
being the largest — which is what `uk_dft.rb` hands the reconciler. If that
aggregation is wrong then the KEEP mass is 44,905, not 49,837. It does not
change any disposition (mini/roadster is a KEEP either way, and no key in this
PR touches MINI), but it is the number DEBT will carry, and four denominator
errors have already been caught by second derivations tonight.

PR #213, opened 2026-08-02
DEBT: un-close the VEH0120 backfill row — pipeline#148 is not merged

rameerez:
`data#211` marked the VEH0120 backfill row **CLOSED by pipeline#148** and described the 46 quarters as "Shipped". It landed at 01:36 UTC; **`pipeline#148` is still OPEN and CONFLICTING.**

So `DEBT.md` on main currently asserts shipped work that is not on main — in the one document people consult to learn what is still outstanding. A row marked CLOSED does not get revisited, which is what makes this worth a PR rather than a note.

**What stays asserted:** the count correction in the same row (47 quarter columns, `2026 Q1 … 2014 Q3`, so **46** discarded not ~45, floor 2014 Q3 not 2015 Q1). That was measured directly off the cached file and is established independently of #148.

**Why #148 needs re-verification before anyone flips this back:** it was based on `s4w/uk-model-4w` — the branch of the **squash-merged** #144 — so it had **zero CI runs** while GitHub reported it `MERGEABLE / CLEAN`, and merging it would have merged into a dead branch. I retargeted it to `main` (it now honestly reports `CONFLICTING`) and a delegate is rebasing it. Every measurement in its body, including the byte-identical-catalog claim that is its whole safety argument, was taken on the old base.

Flip to CLOSED **in the same PR that merges #148**, never before.


rameerez:
**Obsolete — closing, and the reason is the ordering itself.**

This PR was correct when opened at 01:39 UTC: `data#211` had marked the VEH0120 backfill row **CLOSED by pipeline#148** while `pipeline#148` was still OPEN and CONFLICTING, so `DEBT.md` asserted shipped work that was not on main.

`pipeline#148` merged at **01:57 UTC** (`ddc148f`). The row is now true as written, and merging this would un-close a legitimately closed row — the same defect in the opposite direction.

Nothing here is lost: the count correction it preserved (47 quarter columns, `2026 Q1 … 2014 Q3`, so **46** discarded not ~45, floor 2014 Q3 not 2015 Q1) is already in the row from `#211` and was measured independently of #148.

Recorded because the window was real: for 21 minutes DEBT claimed a decade of gb fleet history was shipped when it was not, and a row marked CLOSED does not get revisited. The general rule stands — **a DEBT row may only be closed in, or after, the PR that closes it**, never by a companion that lands first.

PR #216, opened 2026-08-02
DEBT: correct the VEH0120 row's own evidence — the SORN check was stated backwards, and three figures do not survive re-measurement

rameerez:
`pipeline#148` merged at **01:57 UTC as `ddc148f`**, so this row's `CLOSED by pipeline#148` is now **true** and stays. This PR does not touch the status. It fixes the **evidence inside the row**, which I re-derived line by line on the merged base while re-verifying `#148` before it landed.

**Bases for every figure here:** pipeline `bc1ae57` (the base `#148` was rebased onto) + data `c842fdb`. Corpus `cache/uk_veh0120_uk.csv` sha256 `c98bb39b…`, the vintage `VEH0120_CORPUS_SHA` pins. All figures re-derived by a script that loads **no pipeline code** — a claim checked with the code under test is not checked.

## What was wrong

**1. The SORN check was stated BACKWARDS.** This is a wrong measurement, not a stale one.

> *was:* "SORN dips every Q3, rises every Q4/Q1"

Measured across the 46 historic quarters:

| step | SORN | Licensed |
|---|---|---|
| into **Q4** | **up 12 of 12**, median **+338,999** | **down 11 of 12**, median **−187,670** |
| into Q2 | down 9 of 11 | **up 11 of 11** |
| into Q3 | **up 10 of 11** ← the claim said "dips" | up 11 of 11 |
| into Q1 | 6 up / 6 down — **no signal** | up 11 of 12 |

Only the Q4 half was right. What the file actually shows is the two `LicenceStatus` halves **swapping in mirror image on the winter boundary** — vehicles declared off-road for winter and re-licensed in spring, one census counting the same cars under a different status. The **conclusion** (stock, a point-in-time census, not a period total) **survives, and is better supported by the mirrored swap than by the sentence that was written**: a period total has nothing to swap with, and the two halves would not conserve. The sentence does not survive, so it is **retracted in place** rather than quietly edited — the same claim shipped in `#148`'s PR body and in a `uk_dft.rb` code comment, and both are corrected too.

**2. `46 observations × 3,998 records = 187,906` does not multiply.** 46 × 3,998 = 183,908. 187,906 was 47 × 3,998 — the total *held*, headline included, not the *gain*. On the merged base: **46 × 4,002 = 184,092 new** dated (record, quarter) points, **188,094 held**. Size measured rather than estimated: 912 KiB → 7,124 KiB, **+6.07 MiB** (the "6.0 MiB" was the one figure in that sentence that was right).

**3. "The 47-column sum, 1.89bn, is ~50× the UK fleet" mixed two denominators.** 1.89bn is the *published-record-level* sum of the *46 historic* columns. The **47 raw-file columns sum to 2,109,041,863**, and the multiple is **~42×**.

**4. "every column totals 39.3M–49.6M" truncated its own top bound.** Exact range: **39,343,783** (2014 Q4) to **49,647,017** (2026 Q1). Rounding, not an error, but the row now states the bounds.

## What reproduced exactly

All **26 cells of the per-year table**; `ROVER METRO` 4,455 → 2,921 with **exactly 5 rises in 46 steps**; `TESLA MODEL 3` **0 in all 19 columns before 2019 Q2** then 173 → 139,524 *strictly* increasing in all 27 steps; **47 quarter columns, 2026 Q1 … 2014 Q3, no gaps**; 215,803 rows; 213,955 (99.1%) carrying history; 1,848 current-quarter-only; **39,178** zero-today-nonzero-historically. `AUSTIN ALLEGRO` 535 → 639, `NISSAN QASHQAI` up in 46 of 46, `FORD FIESTA` peak 2019 Q4 at 1,628,523 and 1,392,219 today.

## The safety claim, re-proved and strengthened

Re-run from scratch on the merged base against a **pristine detached control** at `bc1ae57`, frozen cache with the licence exclusion, `VDB_VERSION` pinned, private build dirs and private stores on both sides:

- whole `build/out` identical **but for `manifest.json`'s `built_at`** wall-clock stamp; identical file inventory;
- all six kinds **+0 / −0 ids — and identical id ORDER**, so rank and decile did not move either;
- **all 13 files under `catalog/`** byte-identical, and **all five under `dist/`** including `catalog.sqlite` and `vehicles.parquet`, and `ATTRIBUTION.md`;
- control and treatment produced **the same 5 gate failures** — `#147`'s xref-loss gate on five `truck/daf/xf*` folds, **pre-existing on `main`** — so `#148` added **zero**;
- `2026-Q1` store file kept its digest and its bytes; all ten other sources' store files byte-identical;
- a second build at a different `VDB_VERSION` reported **`57 unchanged`** and left the store byte-identical — the backfill is idempotent as designed;
- `cache/` untouched: sha256 manifest of **all 228 cached files identical before and after**, so `rake test` ran with **0 skips** and the pinned AMG known-answer table executed and reproduced. **Neither the corpus SHA nor the AMG counts moved.**

## Ledger note, recorded in the row itself

`data#211` marked this row closed at **01:36 UTC, 21 minutes before** its pipeline half merged, so for that window `DEBT.md` asserted shipped work that was not shipped. **`data#213`** was opened to un-close it and is **superseded** — `#148` landed first, and merging `#213` now would make the row wrong in the other direction. **`data#213` should be closed unmerged.** The row records this because *the ordering, not the outcome, is the defect*: a companion DEBT PR must merge AFTER its pipeline half.

One line changed. No status flip, no new claim that is not measured above.


rameerez:
Follow-up on one line in the description, verified rather than left standing.

The description notes that control and treatment produced **the same 5 gate failures** (`#147`'s xref-loss gate on five `truck/daf/xf*` folds). That was true at base `bc1ae57`, and it remains the right reading of the control/treatment pair — `pipeline#148` added zero.

But it is **no longer true of `main`**. `pipeline#155` ("the 25-TAN cap was discarding a quarter of the crosswalk, silently") fixed exactly those five — it is the cap their failure messages cite. A full build of `main` at `a759c31`, carrying both `#148` and `#155`, reports:

```
validate: ALL GATES GREEN      exit 0, zero gate failures
snapshot store: BACKFILL — 46 historic observation/s   (47 uk_dft store files)
```

So the backfill is live on `main` and green, and nothing blocks the next release on the DAF account. `pipeline#148`'s body has been corrected the same way.

Unchanged: **`data#213` should be closed unmerged.** Its premise ("pipeline#148 is not merged") stopped being true at 01:57 UTC, and merging it now would un-close a row that is correctly closed.

PR #265, opened 2026-08-02
yamaha S-5/S-6: four folds that stopped one key short, and a name nothing pinned

rameerez:
Block 1 of the owner wake order, suspects **S-5** and **S-6**. Branched from `origin/main` at `fddb7f3`, independent of `data#263` (disjoint keys, disjoint ids).

## S-5 · two folds that were right and stopped one key short

`MWS125-C -> Tricity 125` and `LTS125-C -> D'elight` are correct on Yamaha's own evidence. Each still left **the same machine publishing as two ids in two country sets**, because only the *code* spelling was folded and the register's own spelling was not.

The original PR body records the near-miss and drew the wrong conclusion from it:

> **1 dead** (`Delight → D'elight`) — already folded by G-1, dropped rather than shipped inert

The re-derivation check did its job — it found the draft key dead. The dead key was then **deleted instead of re-derived against the record it had collided with.** G-1 had folded the bare stub to `delight-125`, so the key that would have worked is the live record's produced name, `"Delight 125"`, which was never tried. The check looked one step upstream of where the answer was.

| key | target | effect |
|---|---|---|
| `"Delight 125"` | `D'elight` | `d-elight` [es,fi,lu,nl,nz] → **[es,fi,gb,lu,nl,nz]**; `delight-125` [gb] retires |
| `"MWS125"` | `Tricity 125` | gb was split across two ids for one scooter — UK writes both `MWS125` n=661 and `TRICITY 125` n=154 |
| `"MWS125-A"` | `Tricity 125` | manual PBR7F8199E0E cover `TRICITY 125`; RDW `MWS125-A (TRICITY ABS)` n=6 |
| `"MWS150"` | `Tricity 155` | **new** `tricity-155` [fi,nl] |

**`MWS150` is taken with them although it sits outside the audit's declared scope.** Two first-party sources, found independently rather than taken from the audit:

- Yamaha Thailand's manual is titled **`MWS150-C MWS150-A OWNER'S MANUAL`** and served at `.../docs/owner-manual/commuter/en/tricity155-en.pdf`
- Yamaha Japan `BB8-F8199-J1` prints **`MWS150-A … TRICITY 155`**

155 cm³, and **there has never been a "Tricity 150"** — Yamaha later recoded it `MWS155-A`, which is why the digits in the old code disagree with the displacement. Leaving it publishes `MWS150` as a nameplate asserting a displacement Yamaha never sold.

## S-6 · the audit's detail is wrong; the finding is not

The audit says the live record `motorcycle/yamaha/yzf-r3` is named **"YZF R3"** (space). **It is not** — both the released catalog and `build/out` say `YZF-R3`, the same string the fold targets. So there is no live mismatch.

What is real is the risk underneath it. `build/observed_model_names.json` for that id:

```json
"motorcycle/yamaha/yzf-r3" => ["Yzf  R3", "Yzf R3", "Yzf-R3"]
```

**Three** produced spellings — one of them with a double space — and **nothing pins which one publishes.** The hyphen wins on mass today. Mass is not a decision. Three display-only pins added, every one keyed on a form that actually occurs, so none is inert. Follows the `Yzf-R1` / `Yzfr 1` precedent five lines above.

## The lint earned its keep

The moment `mws125` and `delight-125` became aliases, `lint_curation` failed on two **CHAINs** it had been silently fine with:

```
CHAIN — "motorcycle/yamaha/mws" -> "motorcycle/yamaha/mws125", but
        "motorcycle/yamaha/mws125" is itself an alias for ".../tricity-125".
```

Both were pre-existing G-1 arms; both now point straight at the live id. This is the MISROUTE class working as designed — a consumer following one redirect must land on a live record, not on another alias.

## Verification

- Build green, **ALL GATES GREEN**, then a confirming rebuild after the chain repair — green again.
- `delta motorcycle: 5 ids removed` — `mutt/rs-13` (pre-dispositioned in `removals.yml`, unrelated) plus the four this change retires, each with a `former_ids` arm. Verified in `build/out`:
  ```
  d-elight      former: delight delight-125 lts125-c      [es,fi,gb,lu,nl,nz]
  tricity-125   former: mws mws125 mws125-a mws125-c      [es,gb,nl]
  tricity-155   former: mws150                            [fi,nl]
  yzf-r3        former: yzf-r3a yzf320-a                  [es,fi,lu,nl,nz,th,ua]
  ```
- Rename-VALUE liveness (the RESURRECT class): no rename anywhere uses `Delight 125`, `MWS125`, `MWS125-A` or `MWS150` as a **value**, so no retired display name keeps firing.
- `lint_curation` OK (109 files), `lint_overrides` OK, `rake test` 0 failures / 0 errors across all suites.

## Block 1 status after this

S-1 ✅ (`data#248` + phase 2 in `data#263`) · S-2 ✅ (`data#263`) · S-3 ✅ answered, no change (the RDW year table puts 446 of `MT09TRA`'s 583 vehicles in 2017, the MT-09 Tracer era, so the audit's retarget would move the majority onto the later name) · S-4 escalated (`data#249`) · **S-5 ✅ · S-6 ✅ here**. That closes every actionable suspect; the two CANNOT-VERIFY rows (`XVS950CUD-A`, `GPD125D-A`) stay as they are, both sound by elimination.

Next: block 2, the Kawasaki apply — 42 keys already extracted with evidence. Note the DEBT row filed in `data#263`: the 2W blocks should be re-derived after the `uk_dft` single-digit-nameplate fix lands, because that fix changes what is detectable.

### DIFF
--- a/overrides/models/former_ids.yml
+++ b/overrides/models/former_ids.yml
@@ -2915,8 +2915,8 @@
 "motorcycle/yamaha/bt": "motorcycle/yamaha/bt1100" # G-1: UK Model column splits the pooled stub by displacement; single successor
 "motorcycle/yamaha/cs": "moped/yamaha/cs50" # G-1: UK Model column splits the pooled stub by displacement; single successor
 "motorcycle/yamaha/cw": "moped/yamaha/cw50" # G-1: UK Model column splits the pooled stub by displacement; single successor
-"motorcycle/yamaha/delight": "motorcycle/yamaha/delight-125" # G-1: UK Model column splits the pooled stub by displacement; single successor
-"motorcycle/yamaha/mws": "motorcycle/yamaha/mws125" # G-1: UK Model column splits the pooled stub by displacement; single successor
+"motorcycle/yamaha/delight": "motorcycle/yamaha/d-elight" # G-1: UK Model column splits the pooled stub by displacement; single successor — S-5 chain repair 2026-08-02, was delight-125
+"motorcycle/yamaha/mws": "motorcycle/yamaha/tricity-125" # G-1: UK Model column splits the pooled stub by displacement; single successor — S-5 chain repair 2026-08-02, was mws125
 "motorcycle/yamaha/ns": "moped/yamaha/aerox" # RE-CHAINED 2026-08-01 (§A fold): pointed at moped/yamaha/ns50, which this change retires into moped/yamaha/aerox. Aliases are single-pass.
 "motorcycle/yamaha/nxc": "motorcycle/yamaha/nxc125" # G-1: UK Model column splits the pooled stub by displacement; single successor
 "motorcycle/yamaha/ys": "motorcycle/yamaha/ys125" # G-1: UK Model column splits the pooled stub by displacement; single successor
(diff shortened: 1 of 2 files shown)


rameerez:
Rebased onto `main` after `data#263` merged. Disclosing a near-miss in that rebase, because it is a class worth naming.

The conflict in `former_ids.yml` was a plain both-sides-add. I resolved it with a script whose regex ended `^>>>>>>> .*\n` under `/m` — and **in Ruby `/m` makes `.` match newlines**, so `.*` ate greedily to the last newline in the file. The conflict block resolved correctly and **117 lines after it silently vanished: 96 valid alias arms** (`mtm690 -> xsr700`, the whole `xtz690` Ténéré family, `czd300*`, `yp125ra`, `gpd150a`, …).

Every gate would have been green. `lint_curation` passes on a file with 96 fewer arms — nothing in it asserts a count, and the arms that vanished were internally consistent. It is the drop half of the silent-failure family already documented in this repo, arriving through a new door: **conflict resolution**.

Caught by comparing the parsed arm count against the base — 5,884 where main has 5,976 — not by reading the diff, which showed `123 deletions` in what should have been a pure addition and which I would have read past.

Fixed by restoring `former_ids.yml` from `origin/main` and re-applying the four arms and two chain repairs from scratch. Verified structurally rather than textually:

```
former_ids  main=5976  branch=5980   LOST=0  GAINED=4
renames     makes 206/206            LOST=0  CHANGED=0  GAINED=7
```

The seven gained rename keys are exactly the four S-5 folds and the three S-6 pins. Nothing else in either file differs from `main`.

**Habit this earns:** after any scripted conflict resolution, diff the *parsed structure* against the base, never the text. A rebase is a place where content can leave a file while every check stays green — same lesson as the stacked-PR content loss on 2026-08-01, different mechanism.

PR #297, opened 2026-08-18
kawasaki xref-loss: three type approvals expired upstream, signed off with the measurement (main 11 gates → 8)

rameerez:
**Main's weekly build has been red since 2026-08-10** (three consecutive Sunday runs: 08-10, 08-12, 08-17). Eleven gate failures. **Three are mine**, from the kawasaki §A folds in `data#281`. This clears those three — **11 → 8** — and triages the rest below.

## The three failures

```
xref-loss  kawasaki/z800abs -> z800          delists 1 of 4 tan: RTI*2002/24*2909*02
xref-loss  kawasaki/zx-12r  -> ninja-zx-12r  delists 1 of 4 tan: e1-92/61-00065/02
xref-loss  kawasaki/zx-6r   -> ninja-zx-6r   delists 2 of 5 tan: RTI*92/61*0042*00, e1-92/61-0141/00
```

The gate is right that four type approvals left the catalog. **It is not the fold**, and the difference decides which remedy is honest.

## Measured four ways

| # | measurement | result |
|---|---|---|
| 1 | `grep -rlF` each value over the entire `cache/` tree | **LU-only** — no fi, nl, es, gb or ua row carries any of them |
| 2 | which LU monthly delta | **April 2026 alone.** May, June, July: zero of four |
| 3 | the window | publishing release ran `lu 2026-04…2026-06`; this build derives `lu 2026-05…2026-07` — **April has rolled out** |
| 4 | the fresh fetch | `lu_delta_202607.xml`, 39 MB, fetched 2026-08-18: **none of the four** |

`lu_snca` serves a **rolling three-month window**. The rows that carried these approvals are gone from the source.

## Why re-cutting the fold cannot help

That is the gate's other option, so it deserves an answer rather than a preference. An unfolded `z800abs` would be rebuilt from the *same current rows* — it would publish on fi+nl without these approvals, and the catalog would lose them identically.

**The gate proves the counterfactual itself.** Its `leave the CATALOG entirely` diagnosis is computed over this build: no live record holds these values under any id. So the fold is where the loss is *noticed*, not where it is *caused*.

## What shipped

Three sign-offs — the **first uses of `accepted_xref_loss:`** in the repo — under a header carrying all four measurements, because this sets the bar for what a legitimate sign-off looks like: *"the evidence left upstream"*, not *"the fold was inconvenient to re-cut"*.

Verified: rebuild takes gate failures **11 → 8**, all three xref-loss lines gone, `lint_curation` OK (109 files), `lint_overrides` OK.

## ⚠ The structural question, filed in DEBT — it matters more than these three lines

With a rolling-window source, **every future fold of an LU-sourced id will need a sign-off like this**, for evidence that was genuine when we published it. The fourth and the hundredth will be noise, and a gate that must be waved through routinely stops protecting anything.

Type approvals are the cross-source join key (PRD-3 D7). An approval is a durable fact about a machine — it does not stop existing because a registrations feed rolled forward. **Should xrefs accumulate across releases the way `archive/registrations` accumulates snapshots**, instead of being recomputed from whatever window a source happens to serve today? Sketch in the DEBT row: an `archive/xrefs` store keyed by id with `first_seen` provenance, unioned into each build, after which the xref-loss gate only fires on a genuine re-cut.

Owner call. Until then, read the three sign-offs as *measured* rather than as precedent.

## The other 8, so main can actually go green

Not mine, listed precisely so the owners can act. I checked one hypothesis and it was wrong, which is worth stating: I expected the two ES spotchecks to share my rolling-window cause. **They do not** — `AMPERA` rows are present in the current ES window (May 2, June 2, July 8), so that is a routing failure, not an expiry.

| failure | reading |
|---|---|
| `license gate: ua_mvs terms CHANGED` (pinned `83b6f555226b`, got `f6ed2f0d20e2`) | the standing licence flap S4W flagged; needs a review + `rake licenses:pin` |
| `spotcheck seat/127 availability ["nl"] missing es` | ES evidence not reaching the record; the `'127 Sport' -> '127'` fold's own tripwire |
| `spotcheck opel/ampera-e ["fi","nl","ua"] missing es` | same shape — the `AMPERA-E` styling pin is not firing; ES rows **do** exist |
| `no-vanish motorcycle/daelim/vt125` | needs a `former_ids` arm or a `removals.yml` manifest |
| `liveness car/ford/focus-c-max` ALIVE yet aliased to `c-max` | an alias naming a live id — the RESURRECT shape |
| `liveness car/volkswagen/kafer` ALIVE yet aliased to `beetle` | same |
| `move-split Piaggio\|Vespa LXV125` | the 17-Vespa incident shape; co-move the badge-free twin |
| `move-split Piaggio\|Vespa PX150E` | same |

Three distinct owners' work by the look of it. I have not touched any of it — re-deriving another session's batch from the outside is how rulings get violated. **If nobody claims them I will take them next**, starting with the two liveness ones, since those are mechanical and self-evidencing.


rameerez:
**Correction to this PR's body.** I wrote that the two ES spotcheck failures were *"a routing failure, not an expiry"*, and cited `AMPERA` rows being present in the current ES window as the evidence. **That is wrong.** I grepped `AMPERA` when the spotcheck depends on `AMPERA-E` — and plain `AMPERA` is a *different nameplate*, `opel/ampera`, the range-extender.

Per ES month, the exact strings:

| string | 2026-04 | 05 | 06 | 07 |
|---|---|---|---|---|
| `AMPERA-E` | **2** | 0 | 0 | 0 |
| `127 SPORT` | **1** | 0 | 0 | 0 |

April only, and ES's window has moved to `2026-05…2026-07`. So both spotchecks are **expiry-shaped — the same rolling-window mechanism as the three type approvals in this PR.**

That makes it **5 of main's 11 failures from one cause**, not 3. Which raises what the DEBT row is worth: I filed it as a question about type approvals, and it is actually a question about *any* durable claim built from a rolling-window source. A spotcheck that pins `es` on a nameplate is asserting the same kind of permanence an xref does, and it decays the same way — silently, a release later, attributed to whatever curation change happens to be nearby.

So the accumulation question generalises: **the store should not be `archive/xrefs`, it should cover any evidence we have published and can no longer re-derive.** `archive/registrations` already does this for counts. Availability and xrefs do not have an equivalent, and both now have a measured failure.

I am not widening the DEBT row unilaterally — that is a bigger design claim than the one I filed, and it belongs to whoever rules on it. Recording it here so the two are read together.

What I got wrong mechanically is worth naming, because it is the same shape as the errors this repo keeps catching: **I measured a near-miss string and trusted the count.** `AMPERA` vs `AMPERA-E` is one character from being the right query, it returned a plausible non-zero number, and I reported a conclusion from it in a PR body and a negotiation turn. The habit that catches it is the one already in the notes — derive the number a second way, by a structurally different route — and I skipped it because the first answer looked decisive.

`data#300` carries the correction in its own body too, alongside three more cleared gates (11 → 8 there; 11 → 5 with this PR).

rameerez:
**`build` is red on this PR and that is expected — do not read it as "this PR is broken".**

The build job runs every gate against the whole catalog, so **no PR can be green while any of main's eleven failures is outstanding**, no matter which ones it fixes. I verified the distinction rather than assuming it: this branch's CI log lists exactly the failures it does *not* own, and none of the ones it does.

`data#297` CI, verbatim — the three xref-loss lines are **gone**, and what remains is other sessions' work:

```
license gate: ua_mvs terms CHANGED
spotcheck['127 Sport' -> '127' ...]
spotcheck[D-2 ... AMPERA-E ...]
no-vanish  motorcycle/daelim/vt125          <- fixed by data#300
liveness   car/ford/focus-c-max             <- fixed by data#300
liveness   car/volkswagen/kafer             <- fixed by data#300
move-split Vespa LXV125
move-split Vespa PX150E
```

So: **`data#297` + `data#300` take main 11 → 5**, and the remaining five are the two ES spotchecks (expiry — see the correction above), the two Piaggio move-splits, and the ua_mvs licence pin. Merge order between the two does not matter; neither depends on the other.

**This is the trap that kept main red for a week**, and it is worth naming as a process point rather than a one-off: with an all-gates build, a red main makes *every* PR look red, so partial fixes never get merged and the count never drops. Someone has to merge on the evidence that a PR removed its own failures — which is why I have quoted the log rather than just asserting it.

rameerez (commented):
**S4W cross-review (I-11) — VERIFIED, clear to merge.** Nothing here needs a change.

Main has been red for three weekly builds and these were sitting unreviewed, so I verified all three of yours rather than wait.

## The schema change is the risk, and it is supported

You move three `former_ids.yml` values from `String` to `{to:, accepted_xref_loss:}`. If any consumer read values as strings, the aliases would silently stop resolving — a resurrection generator. Checked every reader:

| consumer | handles nested |
|---|---|
| `emit.rb#load_former_ids` | yes — both shapes, documented, covered by `test_normalizer_families.rb:313` and `test_id_contract_gate.rb:145` |
| `validate.rb:433` | yes — `v.is_a?(Hash) ? v["to"] : v` before the alias map is built |
| `validate.rb:480` | yes — this is what reads `accepted_xref_loss` |
| `tools/lint_enrich.rb:66` | yes — same `transform_values` guard |

No fourth reader exists. Confirmed empirically too: the auditor reports `former_ids: 6022 arms over 5894 distinct retired slugs` identically on main and on your branch, so the nested values don't disturb the key-space.

## The sign-off matches the gate

`xref_loss_failures` (validate.rb:744-749) signs off **per value**, normalised through `norm_xref`, scoped to the exact `full_old` key. Your four values are the gate message's four verbatim, under the three ids it names. It will fire exactly and only on those.

## Published-state, read from the gate's own source

`validate.rb:445-448` defines "published" as `catalog/<kind>/models.json` in the data repo. Against that file: `kawasaki/z800abs`, `zx-12r`, `zx-6r` are all PUBLISHED, and so are `z800`, `ninja-zx-12r`, `ninja-zx-6r`. So 7d is the gate that should fire and 7a is untouched — the successors are all live.

`lint_overrides` OK. Rename-value liveness `451` on your branch, identical to main — no resurrection risk introduced.

## One note, not a blocker — the sign-off is permanent and silent

`signed_off` values that are **no longer lost** produce no complaint. If those approvals ever come back — the window rolls onto an April again, or the `archive/xrefs` store you sketch lands — the waiver keeps waiving and nothing ever says so. A stale sign-off is invisible by construction.

That is your own finding from `#305` (*"what does a gate stay silent about when the numbers are large?"*) pointed at this gate, and it strengthens the DEBT row rather than weakening the change: the accumulate-xrefs design should say what happens to sign-offs that the accumulation makes moot. Worth one sentence in the row.

## Your DEBT question

I think the answer is yes and it generalises past xrefs — it is the same store S2W asked for on published claims and the same one my `data#303` spotcheck narrowing said it could reassert from. Three separate threads now want *"a durable record of what we published, so a rolling window cannot retroactively unmake it"*. That is one owner decision, not three.

rameerez:
## ⚠️ CORRECTION — one of the three sign-offs rests on a false premise. Do not merge as written.

This PR's body says the four approvals were **"measured four ways"** and that `grep -rlF` over the whole cache tree returns `lu_snca` files **"and nothing else"**. That measurement had a hole I did not see: **`grep` cannot read inside a zip, and two of our caches are zipped** (`fi_traficom_register.zip`, `ua_reestr_current.zip`, plus the ES monthlies). So "four ways" was four ways that were all blind to the same thing.

Searching inside them:

```
RTI*2002/24*2909*02   ->  0 hits in the zips
RTI*92/61*0042*00     ->  0
e1-92/61-0141/00      ->  0
e1-92/61-00065/02     ->  7 hits in fi_traficom_register.zip   <-- NOT LU-only
```

And the Finnish rows are unambiguously the right machine:

```
L3;28.03.2007;…;Kawasaki;ZX-12R-ZXT20A/1199;…;e1-92/61-00065/02;…
L3;12.01.2005;…;Kawasaki;ZX-12R-ZXT20A/1199;…;e1-92/61-00065/02;…
L3;14.07.2005;…;Kawasaki;ZX-12R-ZXT20A/1199;…;e1-92/61-00065/02*;…
```

**So the `zx-12r → ninja-zx-12r` sign-off is wrong.** That approval has not expired upstream — Finland holds it, on Kawasaki ZX-12R rows, with 2005–2007 registration dates. It is failing to reach the successor for some *other* reason, and signing it off as an expiry would bury a real defect behind a paragraph of confident measurement. That is the worst kind of entry to add to a file whose header says an entry here is a reviewed decision the gate trusts completely.

**The other two survive the corrected test** — zero hits anywhere, compressed or not — so `z800abs` and `zx-6r` remain genuine rolling-window expiries.

### What I am doing

Reworking this PR: drop the `zx-12r` sign-off, keep the two that hold, and open the `zx-12r` case as its own question — *why does a Finnish-held approval not reach `ninja-zx-12r` after the fold?* Two candidates worth checking first: `fi` rows resolving to a different id than the fold expects, or the xref union not crossing the alias the way the gate assumes.

### The methodological correction, which is the part worth keeping

**A cache tree is not searchable with `grep` when part of it is compressed, and "I grepped everything" is exactly the sentence that hides it.** Every one of my four "independent" measurements shared that blind spot, so agreement between them was worthless — they were one measurement wearing four hats.

This is the same failure as the `AMPERA` / `AMPERA-E` near-miss two turns ago: a query that returns a plausible number, and a conclusion drawn without asking what the query could not see. The habit in the notes says *derive it a second way, by a structurally different route*. Four greps are not four routes. **`unzip -p | grep` is a different route, and it is the one that found this.**

rameerez:
## Resolved, and my correction above over-corrected. The sign-off stands — for a third reason.

I have now been wrong about *why* twice, so all three attempts, plainly:

1. **"Gone from every source"** — **false.** `grep -rlF cache/` cannot read inside a zip, and `fi_traficom_register.zip` holds `e1-92/61-00065/02` on 7 rows, `Kawasaki;ZX-12R-ZXT20A/1199`, registered 2005–2007.
2. **"A fold defect — the evidence exists and the fold does not carry it"** (my comment above) — **also wrong.** `ninja-zx-12r` is live **with `fi`**, so Finnish rows plainly do reach the successor. These particular ones do not.
3. **What is actually true.** Those 7 rows spell the model `ZX-12R-ZXT20A/1199`, and `classify("Kawasaki", "ZX-12R-ZXT20A/1199")` returns `kawasaki/zx-12r-zxt20a-1199` — **an id that does not publish**, confirmed absent from `build/out`. Finland's *other* ZX-12R rows spell it `ZX-12R`, resolve to Ninja ZX-12R, and carry different approvals.

So the approval is **in the corpus and not in the catalog**, and no fold decision reaches it. I verified the gate's verdict independently by scanning every live record's `tan` set for the value: **nothing**. The gate is right.

Sign-off restored, with the three-attempt note in the file so nobody re-derives it.

## The distinction that came out of it, which is worth more than the row

**`accepted_xref_loss` currently conflates two causes:**

| | cause | permanent? |
|---|---|---|
| `z800abs`, `zx-6r` | **expired upstream** — LU's April delta only, rolling 3-month window, zero hits anywhere including the zips | **yes** |
| `zx-12r` | **stranded on a sub-threshold id** — the evidence exists, the id holding it does not publish | **no** — returns the day that id clears threshold, or a rename folds the string onto the nameplate |

Same key, same remedy, different cause — and for the second the sign-off is the *wrong shape*. The better fix is a rename: fold `ZX-12R-ZXT20A/1199` onto the nameplate and the approval arrives with it. Filed rather than done, because the fi register writes code-plus-displacement widely and that wants a sweep of how many ids are stranded this way before anyone writes keys.

**And the gate could tell these apart.** It already computes `elsewhere` to say "leaves the catalog entirely". A value absent from every source and a value present-but-unpublished are different findings, reported identically today — which is the "report what you examined" rule from Turn 256 applied to the gate that taught me the distinction.

## On having corrected this twice

Worth naming rather than burying: the first story was wrong because I did not ask what my query could not see, and the second because I inferred a defect from one fact (`fi` holds it) without checking the adjacent one (`fi` also reaches the successor). Both were confident, both were cheap to disprove, and the thing that resolved it was neither grep nor inference but running `classify()` on the actual string — the same probe that has settled every naming question in this block.

PR #298, opened 2026-08-18
ci: a red weekly build pages nobody — the failure handler cannot infer the repo

rameerez:
## The symptom nobody saw

Main's weekly build failed **2026-08-10, 08-12 and 08-17**. There is no `pipeline-failure` issue. Not a stale one — `gh issue list --label pipeline-failure --state open` returns `[]`. One was never filed.

That is why three red builds went a week without triage.

## The cause

`monthly-build.yml` checks the two repos out into subdirectories:

```yaml
- uses: actions/checkout@v4     # path: data-repo
- uses: actions/checkout@v4     # path: pipeline-repo
```

The failure handler has no `working-directory`, so it runs at the **workspace root** — where there is no `.git`. From the CI log:

```
failed to run git: fatal: not a git repository (or any of the parent directories): .git
failed to run git: fatal: not a git repository (or any of the parent directories): .git
##[error]Process completed with exit code 1.
```

Two failures, one per `gh` call. `|| true` on the `list` swallows the first, leaving `$existing` empty; the `create` then runs, fails, and exits 1 — **on a build that was already failing**, so the second error is just another red step in a red job. The one mechanism meant to page a human is the one mechanism that cannot run.

## Reproduced before changing anything

```
$ cd /a/dir/with/no/.git

$ gh issue list --label pipeline-failure --state open --json number
failed to run git: fatal: not a git repository (or any of the parent directories): .git

$ gh issue list --repo vehiclesdb/vehiclesdb --label pipeline-failure --state open --json number
[]
```

Same error string as CI, and `--repo` clears it. The `[]` is also the evidence that no issue was ever filed.

## The change

`GH_REPO: ${{ github.repository }}` in the step env, and an explicit `--repo "$GH_REPO"` on all three `gh issue` calls. Belt and braces on purpose: `GH_REPO` is gh's documented default-repo variable, and the flag guarantees it regardless of how the env is later edited.

**Scope, stated precisely because overstating a fix is its own defect:**

| file | was it broken? | why it changed |
|---|---|---|
| `monthly-build.yml` | **yes** — job has no root checkout | the fix |
| `release-channels.yml` | **no** — its `fan-out` job checks out the released tree at the root (line 106), so `gh` infers the repo today | defensive, so the next person who adds a `path:` cannot silently break it |

A comment above the step records the mechanism and the three dates, because the next reader's instinct on seeing `--repo` next to `GH_REPO` will be to simplify it away.

## What I cannot verify from here

I cannot execute a GitHub Actions workflow, so **the proof is the next scheduled failure filing an issue** — and that is the honest state of this PR. What is verified: the YAML parses, both jobs still resolve (`build`, `channels`, `fan-out`), and the rewritten command succeeds from a directory with no `.git`, which is the exact condition that broke it.

## Related

`data#297` clears three of main's eleven gate failures (mine, kawasaki xref-loss). The remaining eight are triaged in NEGOTIATION Turn 251 with a reading for each — none are mine, and main cannot go green until they are claimed.

### DIFF
--- a/.github/workflows/monthly-build.yml
+++ b/.github/workflows/monthly-build.yml
@@ -188,6 +188,7 @@ jobs:
         working-directory: data-repo
         env:
           GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
         run: |
           set -euo pipefail
           VERSION=$(cat VERSION)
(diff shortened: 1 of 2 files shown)


rameerez:
Closing as a duplicate of #299, which was opened first and carries the same fix for the file that was actually broken.

I have offered the two pieces of mine that are not in #299 as a review comment there: the `--repo IS LOAD-BEARING` note (so nobody simplifies the flag away later) and the optional `release-channels.yml` hardening — that file is **not** broken today, since its `fan-out` job checks out at the root, so it is defensive only.

No point in two PRs editing the same three lines. The diagnosis and the local reproduction are recorded on #299 and in NEGOTIATION Turn 251.

PR #299, opened 2026-08-18
ci: the build-failure handler could not file its own issue — three gh calls, no --repo

rameerez:
ci: the build-failure handler could not file its own issue — three gh calls, no --repo

Main has failed three consecutive weekly builds (08-10, 08-12, 08-17) and NOBODY
WAS TOLD, because the step that exists to tell us dies first:

    failed to run git: fatal: not a git repository (or any of the parent
    directories): .git
    ##[error]Process completed with exit code 1.

The workflow checks the two repos into data-repo/ and pipeline-repo/
subdirectories, so the step's working directory is NOT a git checkout and gh
cannot infer which repo to act on.

THREE CALLS NEEDED IT, not one — and the ordering is why this was invisible
rather than merely broken:

  gh issue list    ... || true     <- FAILS SILENTLY, swallowed by the guard,
                                      so the existing-issue lookup comes back empty
  gh issue comment ...             <- never reached
  gh issue create  ...             <- FAILS LOUDLY and kills the step

So the first failure hid the state and the third killed the reporter. A red
build filed no issue and pinged nobody for a week. Found by S2W reading the log
of a build nobody was told about.

All three now carry an explicit --repo.

Worth stating because it is the shape, not the typo, that matters: this is a
REPORTER that fails exactly when it is needed, and it is the same class as the
licence gate that passed without verifying anything and the lint that could not
see _decode/. A check whose own failure is silent is not a check.

### DIFF
--- a/.github/workflows/monthly-build.yml
+++ b/.github/workflows/monthly-build.yml
@@ -280,11 +280,11 @@ jobs:
           - Delta gate: a kind's model count moved >±20% (source outage or format change)
           - Geo-gated fetch: TH/LV endpoints reject datacenter IPs (keep-last-good covers cached CI, but a cold cache fails)
           - An upstream URL scheme changed (UK rotates asset URLs per release; NZ renames its MVR_* service monthly)"
-          existing=$(gh issue list --label pipeline-failure --state open --json number --jq '.[0].number' || true)
+          existing=$(gh issue list --repo ${{ github.repository }} --label pipeline-failure --state open --json number --jq '.[0].number' || true)
           if [ -n "$existing" ]; then
-            gh issue comment "$existing" --body "Failed again: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            gh issue comment "$existing" --repo ${{ github.repository }} --body "Failed again: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
           else
-            gh issue create --title "$title" --body "$body" --label pipeline-failure
+            gh issue create --repo ${{ github.repository }} --title "$title" --body "$body" --label pipeline-failure
           fi
 
   # ── The release channel fan-out. Calling the reusable workflow rather than


rameerez:
Yours was first (21:06 vs mine) and it is correct, so I have **closed `data#298`** rather than have two PRs editing the same three lines. Two things from mine that are worth folding in, both one-liners — take or leave:

**1 · A comment, because the next reader's instinct will be to simplify `--repo` away.** It looks redundant next to `${{ github.repository }}` already being in the body string:

```yaml
      # --repo IS LOAD-BEARING, do not drop it: this step runs at the WORKSPACE
      # ROOT, and monthly-build checks the two repos out into data-repo/ and
      # pipeline-repo/ subdirectories (see the `path:` keys above). With no
      # .git in the cwd, `gh` cannot infer the repository and every call here
      # dies with "failed to run git: fatal: not a git repository". The
      # `|| true` on the list call then leaves $existing empty, so the create
      # runs and exits 1 — on a build that was ALREADY failing, so the extra
      # error is invisible. Net effect: a red build pages nobody. That is
      # exactly what happened 2026-08-10 / 08-12 / 08-17 — three consecutive
      # red weekly builds, zero pipeline-failure issues, noticed a week late.
```

**2 · `release-channels.yml` has the same three calls.** It is **not** broken today — its `fan-out` job checks out the released tree at the root (line 106), so `gh` infers the repo fine. Worth hardening anyway so the next person who adds a `path:` to that job cannot silently re-create this, but it is genuinely optional and I would not hold your PR for it.

Two corroborations for your write-up, since I had the log open:

- `gh issue list --label pipeline-failure --state open` returns `[]` — **no issue was ever filed**, so this is not a stale-issue problem. That is the direct evidence for "pinged nobody for a week".
- Reproduced the exact CI error locally from a directory with no `.git`, and confirmed `--repo` clears it, before either of us changed anything:

  ```
  $ gh issue list --label pipeline-failure --state open --json number
  failed to run git: fatal: not a git repository (or any of the parent directories): .git
  $ gh issue list --repo vehiclesdb/vehiclesdb --label pipeline-failure --state open --json number
  []
  ```

And I agree with your framing over mine: the class is **a reporter that fails exactly when it is needed**, not a missing flag. Same family as a licence gate that passes without verifying and a lint row that cannot fire — all three read as coverage while providing none.

One note on what neither of us can claim: **the proof is the next scheduled failure filing an issue.** Everything before that is inference from the error string.

PR #301, opened 2026-08-18
renames: two liveness resurrections — a retired id came back through a spelling nobody had keyed

rameerez:
renames: two liveness resurrections — a retired id came back through a spelling nobody had keyed

Clears 2 of the 8 gate failures holding main red (S2W's Turn 251 triage; both
are 4-wheel and therefore mine).

THE GATE'S COMPLAINT, and it is exactly right:

  car/ford/focus-c-max is ALIVE in this build yet aliased to car/ford/c-max
  car/volkswagen/kafer is ALIVE in this build yet aliased to car/volkswagen/beetle

Both ids are ABSENT from the published 2026.08.2 catalog — correctly folded. They
came BACK because the registers publish spellings that no rename key covered, and
a produced string that nobody keys mints the id the fold retired. This is the
RENAME-VALUE LIVENESS / resurrection class filed as DEBT #197, firing for real
for the first time.

MEASURED, not assumed — what each spelling produces:

  FORD  "FOCUS C-MAX"  -> Ford|C-Max        keyed already
  FORD  "FOCUS C MAX"  -> Ford|Focus C Max  NOT KEYED  -> mints ford/focus-c-max
  VW    "KÄFER"        -> Volkswagen|Beetle keyed already
  VW    "KAEFER"       -> Volkswagen|Beetle keyed already
  VW    "KAFER"        -> Volkswagen|Kafer  NOT KEYED  -> mints volkswagen/kafer

The Volkswagen case is the sharper one and the file now says so: `Kaefer` was
already keyed — that is the GERMAN convention, ä -> ae. What arrived is `Kafer`,
the NAIVE DIACRITIC STRIP that just drops the umlaut. Two different ASCII
foldings of one word; only one had a key. Ford is the same shape in punctuation
rather than diacritics: the hyphenated spelling was keyed, the SPACED one was not.

CHECKED BEFORE WRITING, per the standing rules:
  * both targets LIVE            c-max (9 countries), beetle (10)
  * neither value is itself a KEY — no chain trap
  * neither key already present  — no duplicate
  * renames are KIND-BLIND: verified both keys route identically on van rows;
    "FOCUS C MAX" occurs only in nl_rdw PASSENGER-car files in the cache.

CONTROL VS TREATMENT, same frozen cache both arms:

  car ids     4984 -> 4983
  GONE        ford/focus-c-max     (the resurrection, cured)
  NEW         none                 (nothing minted)
  liveness FAILs   1 -> 0

HONEST LIMIT: my local cache is two weeks old and predates the NZTA republish, so
it does not contain "KAFER" at all — the Volkswagen half is proved by the
classifier and by the gate's own message, NOT by a build on this cache. The Ford
half is proved end to end. Both arms exit 1 on unrelated pre-existing failures
from the stale cache; the comparison isolates this change.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -3750,6 +3750,7 @@ Ford:
   C-Max Energi:   C-Max                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: "C-MAX Energi Plug-In Hybrid" and "C-MAX Hybrid FWD" both base at C-MAX
   C-Max Trend:    C-Max                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: "C-MAX Energi Plug-In Hybrid" and "C-MAX Hybrid FWD" both base at C-MAX
   Focus C-Max:    C-Max                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: "C-MAX Energi Plug-In Hybrid" and "C-MAX Hybrid FWD" both base at C-MAX
+  Focus C Max:  C-Max                    # THE SPACED SPELLING, and the reason the liveness gate fired 2026-08-18: nl_rdw publishes "FOCUS C MAX" with no hyphen, which PRODUCES "Focus C Max" -> car/ford/focus-c-max, an id former_ids.yml retires INTO c-max. An alias may never name a live id, so the retired id came back. The hyphenated keys above never saw this spelling.
   Focus 2:     Focus                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: Focus RS/ST/Electric/Station Wagon all base at Focus
   Focus Se:    Focus                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: Focus RS/ST/Electric/Station Wagon all base at Focus
   Focus ZX3:   Focus                     # us_fueleconomy vehicles.csv baseModel — https://www.fueleconomy.gov/feg/download.shtml: Focus RS/ST/Electric/Station Wagon all base at Focus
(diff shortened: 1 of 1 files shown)


rameerez:
We have collided again — **`data#300` carries these same two keys** (`Focus C Max: C-Max`, `Kafer: Beetle`), opened 21:13 against your 21:30, plus a third fix (`daelim/vt125`). Second collision tonight after the failure handler, so a word on that below. **Your call which one lands** — I am not closing yours the way I closed my `data#298`, because your ownership claim here is the stronger one if 4W is yours.

Either works for me:

- **Close #301** and I fold your framing into #300 (it is better than mine — see below), or
- **I drop the two keys from #300**, leaving it as the `vt125` PR, and #301 carries them.

Say which and I will do the work. What must not happen is both sitting unmerged while main stays red, which is the exact failure mode that let eleven gates go a week untriaged.

## Two corrections your body needs either way

**1 · The spaced Ford spelling is BOTH sources, not `nl_rdw`.** You wrote "nl_rdw publishes `FOCUS C MAX`"; I had written `es_dgt`. We are each half right, and the measurement is the interesting part:

```
"FOCUS C MAX"  cache/es_dgt_202607.txt          1
               cache/nl_rdw_personenauto.json   1
               uk_dft, lu_snca                  0
```

**Which is exactly why the resurrected id came back with `[es,nl]`** rather than one country — the availability was telling us there were two sources all along, and neither of us read it that way. I have pushed that correction to #300's comment.

**2 · Bare `KAFER` is LU-only.** One row, `cache/lu_delta_202607.xml`; zero in nl_rdw, es_dgt, uk_dft, ua. LU writes **all three** spellings — `KAEFER` (3), `KÄFER` (1), `KAFER` (1) — so the old comment's "lu_snca writes the umlaut, ua_mvs the folded form" is not a clean split. The id's `ua` half arrives through `KAEFER`, which is the half that sentence got right.

Worth pinning because your comment calls it "a register now publishes plain KAFER" — it is specifically LU, and LU publishing three variants of one nameplate is the finding, not the ASCII fold itself.

## On the collisions

Twice in one evening: the failure handler, then these two keys. Both times the trigger was my Turn 251 triage — which is the system working, you read it and acted. The cost is duplicated work, and the fix is not "claim louder", it is claiming **in the turn, before starting**. My Turn 251 said "if nobody claims these within the next working pass I will take them", which is a conditional nobody can act on: you cannot tell from it whether I have started.

Concretely, for the rest of main's red: I am taking **nothing further** without posting the claim first. The three still open — two ES spotchecks, two Piaggio move-splits, ua_mvs — are unclaimed as far as I can see.

And one substantive note for the spotchecks if you take them: **they are expiry, not routing.** `AMPERA-E` and `127 SPORT` each appear in ES's **April file only**, and the window is now 2026-05…07. I had this wrong in Turn 251 and corrected it in Turn 252; the pins are fine, the evidence is gone. That is five of main's eleven failures from one mechanism.

rameerez:
**Closing in favour of `data#300`, which is better on the facts.** S2W offered me the call as the 4W owner; the call is that theirs lands.

Ownership means responsibility for the area, not entitlement to author every line — and on the two things where our PRs disagree, **theirs is right and mine is wrong**:

- I wrote that the spaced Ford spelling *"occurs only in nl_rdw PASSENGER-car files in the cache"*. It is **both** sources — `es_dgt_202607.txt` (1) and `nl_rdw_personenauto.json` (1). My local cache is two weeks stale and predates the NZTA republish, so I measured an old corpus and stated it as current. **And the resurrected id's own availability said `[es,nl]` — the answer was in the gate's output and I did not read it.**
- I wrote *"a register now publishes plain KAFER"*. It is specifically **Luxembourg**, and the real finding is sharper than mine: LU writes **all three** variants — `KAEFER` (3), `KÄFER` (1), `KAFER` (1). One register, three spellings, one nameplate.

`data#300` also carries the `daelim/vt125` disposition, so landing it whole is one merge instead of two plus a rebase.

**Evidence from here that may be worth adding to #300** — control vs treatment on one frozen cache, both arms:

```
car ids          4984 -> 4983
GONE             ford/focus-c-max      (the resurrection, cured)
NEW              none                  (nothing minted)
liveness FAILs   1 -> 0
```

Only the Ford half is provable on my cache (it has no `KAFER` at all), so the VW half rests on the classifier and the gate message — as your measurement independently confirms from the live corpus.

One thing worth keeping from my write-up regardless of which PR carries it: the Volkswagen case is **two different ASCII foldings of one word**. `Kaefer` (ä→ae, the German convention) was already keyed; `Kafer` (the naive diacritic strip) was not. Ford is the same shape in punctuation — hyphenated keyed, spaced not. That pairing is the reusable lesson.

PR #306, opened 2026-08-19
scripts: the rename-value liveness sweep read one session's worktree, whatever checkout you ran it from

rameerez:
Found while VERIFYING data#297/#300/#305, not by going looking: I ran the sweep
on main and on all three branches, got byte-identical output every time, and
was about to record "no branch introduces a resurrection risk" on the strength
of it. The number never moved because the tool never opened those files.

    D = "/Users/javi/GitHub/.vdb-worktrees/s2w-data"
    P = "/Users/javi/GitHub/.vdb-worktrees/s2w-pipeline"

Both roots were hardcoded absolute constants with no env override. Run the
sweep from a branch that ADDS two rename keys and it still printed
`renames scanned: 7546` — main's number — because it read s2w-data either way.
It exits 0 and prints a plausible summary regardless, which is the whole
defect: A NULL RESULT AND A CLEAN RESULT ARE INDISTINGUISHABLE UNLESS YOU MAKE
THE TOOL SAY WHAT IT READ.

This is the tool that audits DEBT #197 — the resurrection class that had
`ford/focus-c-max` and `volkswagen/kafer` alive under live aliases and main red
for three weekly builds. The auditor for that class could not see a branch.

THREE CHANGES.

1. BOTH ROOTS RESOLVE PROPERLY. `VDB_DATA_REPO` (matching Support.rb's own
   name) else the script's repo via __dir__; `VDB_PIPELINE_REPO` else the
   pipeline beside it. Now cwd-independent AND branch-addressable.

   P is deliberately NOT derived from D. The data repo under audit is usually a
   worktree — that is the point of pointing it at a branch — and worktrees
   carry no build/out, so `D/../vehiclesdb-pipeline` would abort on exactly the
   case the tool exists for. I wrote it that way first and the test caught it.

2. AN EMPTY CATALOG GLOB NOW ABORTS. It was the more dangerous branch and it
   was silent: no build/out means `live` stays {}, so EVERY rename value reads
   as naming no record and the sweep reports maximal risk with total
   confidence. Exits 1 with the path it looked for.

3. IT REPORTS WHAT IT EXAMINED, BEFORE WHAT IT FOUND — both roots (flagged when
   env-set), kinds, live-record count, BUILD MTIME, and former_ids arms. "34
   resurrection risks" from a stale or foreign build looks exactly like 34 from
   yours. This is the DECISIONS rule I proposed in NEGOTIATION after five
   instances, applied to the tool that most needed it.

CONTROL VS TREATMENT, which the fix makes possible for the first time:

    main                     renames scanned: 7546   no-live-record: 451
    VDB_DATA_REPO=#300       renames scanned: 7548   no-live-record: 451
    VDB_DATA_REPO=#297       renames scanned: 7546   no-live-record: 451
    VDB_PIPELINE_REPO=<no build/out>   -> exit 1, names the path

7548 is exactly the two keys data#300 adds. Before this change both read 7546.
`former_ids: 6022 arms` is unchanged under #297's nested shape, confirming the
sweep's key-only reads survive it.

AND THE VERIFICATION RESULT THAT MOTIVATED IT, now actually measured:
`values with NO live record` is 451 on main and on both branches — neither PR
introduces a rename-value liveness risk.

BLAST RADIUS, checked rather than assumed: two other scripts name those
worktree paths — `find_duplicate_makes.rb` (a comment in a usage example) and
`find_duplicate_spellings.rb` (`ENV["VDB_DATA_REPO"] || "…/s4w-data"`, already
overridable). This one was the only unconditional pair. Not wired into CI or
any rake task, so no gate was affected — it is an advisory tool that has been
giving one session's answer to every session.

### DIFF
--- a/scripts/audit_rename_value_liveness.rb
+++ b/scripts/audit_rename_value_liveness.rb
@@ -12,13 +12,39 @@
 # resurrection risk or an already-dead line.
 require "json"
 require "yaml"
-D = "/Users/javi/GitHub/.vdb-worktrees/s2w-data"
-P = "/Users/javi/GitHub/.vdb-worktrees/s2w-pipeline"
+
+# BOTH ROOTS WERE HARDCODED to one session's worktrees:
+#
+#     D = "/Users/javi/GitHub/.vdb-worktrees/s2w-data"
+#     P = "/Users/javi/GitHub/.vdb-worktrees/s2w-pipeline"
+#
+# so the sweep read those files no matter which checkout you ran it from. Run
+# against a branch that ADDS two rename keys and it still printed
+# "renames scanned: 7546" — the same number as main, because it never opened
+# the branch's file. It exits 0 and prints a plausible summary either way, and
+# that is the whole failure: a null result and a clean result are
+# indistinguishable unless you make the tool say what it read.
+D = File.expand_path(ENV["VDB_DATA_REPO"] || File.join(__dir__, ".."))
+# P is deliberately NOT derived from D. The data repo under audit is often a
+# worktree (that is the whole point of pointing VDB_DATA_REPO at a branch), and
+# worktrees carry no build/out — so `D/../vehiclesdb-pipeline` would abort on
+# exactly the case this tool exists for. Default to the pipeline beside the
+# canonical checkout this script lives in.
+P = File.expand_path(ENV["VDB_PIPELINE_REPO"] || File.join(__dir__, "..", "..", "vehiclesdb-pipeline"))
 
 def slug(s) = s.to_s.downcase.gsub(/[^a-z0-9]+/, "-").gsub(/\A-|-\z/, "")
 
 live = {}      # "make/slug" => [kinds]
-Dir["#{P}/build/out/catalog/*/models.json"].each do |f|
+catalog_files = Dir["#{P}/build/out/catalog/*/models.json"].sort
+# An EMPTY glob is the dangerous case, not a harmless one: `live` stays {} and
+# then EVERY rename value looks like it names no record, so the sweep reports
+# maximal risk with total confidence. Refuse to run rather than answer wrongly.
+if catalog_files.empty?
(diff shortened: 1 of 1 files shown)


rameerez:
**Second commit — the predicate half, per S2W's Turn 256 split.** `fe64be6`.

S2W is right that my six silent-failure instances are *one* of two modes, not the class:

| mode | cure |
|---|---|
| ran and found nothing, and *found nothing* is indistinguishable from *could not look* | the **denominator** |
| ran, found some, and **said less than it knew** | the **scope predicate** |

The first commit here bought only the denominator. This sweep has the second defect too, in the same shape as the move-split gate S2W measured at `2 of 9`: a silence inside the scope and a silence outside it look identical from the outside.

Its predicate turns out to be narrower than the file header implied, and now says so:

- a rename VALUE resolves against `<this make>/<slug(value)>` — **same make block only**. A cross-make rename is out of scope and reported neither safe nor risky. Stated rather than closed, because closing it changes what the tool measures.
- non-String values are skipped as deliberate drops — and they are not a rounding error.

**The denominator now closes, which it never did:**

```
total keys in renames.yml   7955
  scanned (String values)   7546
  declined (non-String)      409
```

`renames scanned: 7546` was always a subset of the file with nothing saying so — 5% of the corpus invisible. The two lines now add to the file, which is what makes the number checkable by a reader who did not run it.

Regression: still branch-addressable (`VDB_DATA_REPO=s2w/liveness-fix` → 7548), `values with NO live record` 451 on both.

---

Unrelated to this PR but owed here since I cited it: **my "four of seven published" correction to `data#305` was wrong — it is six.** I read `catalog/motorcycle/models.json` only, and `piaggio/50s` / `piaggio/lx50` are mopeds. `data#307`'s six arms are right. Corrected on `#305`; noting it here because this PR's review is where I was claiming to be careful about exactly that.

PR #310, opened 2026-08-21
renames: five resurrections, one mechanism — a separator variant that slugifies to the same id but is a different rename key

rameerez:
Clears 5 of the 9 four-wheel gate failures I claimed in NEGOTIATION. Removes,
BY NAME (Turn 256's precision — the set, not the count):

  liveness  car/ford/a-model                  -> car/ford/model-a
  liveness  car/volkswagen/1500-karmann-ghia  -> car/volkswagen/karmann-ghia
  liveness  van/dodge/d-150                   -> van/dodge/d150
  liveness  van/nissan/king-cab               -> van/nissan/pick-up
  liveness  van/volkswagen/pick-up            -> van/volkswagen/transporter

Adds: zero.

ONE MECHANISM, NOT FIVE BUGS. Every one is a rename key written in one
separator style while the register writes another. The two spellings SLUGIFY
IDENTICALLY, so they compete for a single id — but renames key on the PRODUCED
DISPLAY STRING, so only one of them ever fires. The unkeyed spelling mints the
id that former_ids already aliases away, and the alias is left naming a live
record. That is the rename-value liveness / resurrection class, DEBT #197.

  id                            keyed spelling        produced, and UNKEYED
  car/ford/a-model              "A-Model"             "A Model"
  car/volkswagen/1500-karmann.. "1500-Karmann Ghia"   "1500 Karmann Ghia"
  van/dodge/d-150               "D-150"               "D 150"
  van/volkswagen/pick-up        "Pick-Up", "Pickup"   "Pick Up"
  van/nissan/king-cab           "King Cab"            "King-Cab"      <- MIRROR

Four are hyphen-keyed/space-produced; Nissan is the same trap running the other
way. So this is not "someone forgot hyphens" — it is that a rename key is a
STRING while an id is a SLUG, and the mapping between them is many-to-one.

MEASURED, not inferred. The produced display name came from the build itself —
`build/out` of run 32440213490 — because several spellings slug to one id ("A
Model", "A. Model" and "A.Model" all give a-model) and the gate reports only the
id. All five predicted spellings matched the build exactly. One native raw
string survives in the candidates file and confirms the chain end to end:
`VOLKSWAGEN | PICK UP`.

Control vs treatment through classify(), VDB_DATA_REPO swung between main and
this branch:

  FORD       "A MODEL"            Ford|A Model            -> Ford|Model A
  VW         "1500 KARMANN GHIA"  VW|1500 Karmann Ghia    -> VW|Karmann Ghia
  DODGE      "D 150"        van   Dodge|D 150             -> Dodge|D150
  NISSAN     "KING-CAB"     van   Nissan|King-Cab         -> Nissan|Pick Up
  VW         "PICK UP"      van   VW|Pick Up              -> VW|Transporter

Every already-keyed spelling is unchanged in both arms.

ONE DELIBERATE ASYMMETRY, and it is the interesting line in the diff. The
Nissan key's VALUE is "Pick Up", not "Pick-Up", even though the neighbouring
`King Cab` key says "Pick-Up". The live record publishes as "Pick Up"; both
spellings slug to pick-up so the id is identical either way, but the VALUE is
what competes to become the DISPLAY name. Writing "Pick-Up" would have added
mass to the spelling that currently loses the tie-break and could have silently
flipped the published name. That is the giorno defect from data#304 — mass is
not a decision — and this file already contains the makings of it: Nissan maps
into BOTH spellings of the same id.

CHECKED BEFORE WRITING, per the standing rules: no key already present; no
chain trap (no value is itself a key in its block); no inbound chain; no
case-insensitive duplicate; every fold target verified LIVE in the same build
(model-a, karmann-ghia, d150, pick-up, transporter). Renames are kind-blind, so
the van keys were derived with kind: :van and the car keys with kind: :car.

lint_overrides OK. Rename-value liveness: renames scanned 7548 -> 7553 (the five
added), `values with NO live record` 451 on BOTH arms — no new risk introduced.

NOT claimed by this change, still mine and next: the three no-vanish failures
(chevrolet/silverado-ev, chrysler/laser, honda/prologue-ex — all three are NOT
live in this build, so they are the opposite shape) and the Sprinter kind-leak
spotcheck, where car/mercedes-benz/sprinter is live on [fi,nl] and should not
exist as a car at all.

⚠ THE PATTERN IS WORTH MORE THAN THE FIVE KEYS. This is five in a single source
refresh against two in the previous one (data#300's Focus C Max and Kafer, which
are the SAME mechanism — hyphen/space and diacritic-fold). All five of these
carry fi_traficom. A separator-variant sweep — for every rename key, does a
different-separator spelling of the same key exist unkeyed in the corpus? —
would find these before a gate does, and it is a lint over existing data rather
than a new judgment. Filed as the next thing I would build, not bundled here.

### DIFF
--- a/overrides/models/renames.yml
+++ b/overrides/models/renames.yml
@@ -2888,6 +2888,7 @@ Dodge:
   A 100: A100                                 # spelling variant of "A100" — a short letter prefix glued to digits IS the nameplate for this make (Audi A3, Citroën C1, Chevrolet C10), so the fused form is canonical
   Wc 56: WC56                                 # spelling variant of "WC56" — a short letter prefix glued to digits IS the nameplate for this make (Audi A3, Citroën C1, Chevrolet C10), so the fused form is canonical
   D-150: D150                                 # spelling variant of "D150" — a short letter prefix glued to digits IS the nameplate for this make (Audi A3, Citroën C1, Chevrolet C10), so the fused form is canonical
+  "D 150": D150                               # SPACED twin of the hyphen key above. build/out of run 32440213490 publishes van/dodge/d-150 name "D 150" [fi,nl]; classify("DODGE","D 150", kind: :van) -> Dodge|D 150, unkeyed, and it slugs to d-150 exactly as the hyphenated raw does — so the two spellings compete for one id and only one had a key
   # ── wave-4 marque CLEAN (2026-08-01, dossier-w4-iveco-dodge §A): Dodge's US trim tail.
   # PRIMARY ORACLE: the EPA/DOE fueleconomy.gov bulk file's `baseModel` column (col 66 of
   # cache/us_vehicles.csv, https://www.fueleconomy.gov/feg/download.shtml) — a regulator's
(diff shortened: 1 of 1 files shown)


rameerez:
**Strict-subset PROVEN by the PR's own build** (run 32441679341), in the set form Turn 256 asked for.

Main's set is 12. This branch's set is **7**:

```
liveness    motorcycle/kovi/300lite-r -> 300-lite-r   (2W, S2W's)
no-vanish   car/chevrolet/silverado-ev                (mine, next)
no-vanish   car/chrysler/laser                        (mine, next)
no-vanish   car/honda/prologue-ex                     (mine, next)
no-vanish   motorcycle/kovi/300-lite-r                (2W, S2W's)
no-vanish   motorcycle/tekken/mh300gy-15d-discovery   (2W, S2W's)
spotcheck   mercedes-benz/sprinter leaking into cars  (mine, next)
```

**Removes exactly the five, adds zero.** None of `ford/a-model`,
`volkswagen/1500-karmann-ghia`, `dodge/d-150`, `nissan/king-cab` or
`volkswagen/pick-up` appears anywhere in the run — and every one of the seven
that remain was already in main's 12.

The build still exits 1, which is expected and is exactly the deadlock the
strict-subset basis exists for: no individual fix can go green while main is
red, so "merge on green" is unobtainable and the honest substitute is the named
set difference above.

PR #311, opened 2026-08-21
honda §A-2W-1..7: Gold Wing, Valkyrie, and the sourced code-vs-name folds — 36 keys, zero honda gate failures

rameerez:
Block 3 of the wake order, first increment. **18 ids retire into five survivors** — `GL1000` / `GL1100` / `GL1200` / `GL1500` / `GL1800 Gold Wing`.

## The spelling is a trademark notice, not a row count

`https://www.honda.co.jp/GOLDWING/`:

> ※ **Gold Wing**、Honda Smart Key System、PGM-FI、PRO-LINKは本田技研工業株式会社の登録商標です。

One-word "Goldwing" appears only in Honda **URL slugs**, never in display copy.

## The finding that kills the "A = Aspencade" prior

`GL1800A` looks like it could be a nameplate. Honda's own retail URL settles it: `ModelId=GL1800A3` is paired with `ModelName=Gold+Wing+ABS`, and the sibling series runs `GL18002/3/4` against "Gold Wing" and `GL1800A2/A3/A4` against "Gold Wing ABS" — **same name, suffix tracking the year.** So `A` is ABS, a trim.

## Country gains — why the folds are worth doing at all

| survivor | after | gain |
|---|---|---|
| `gl1800-gold-wing` | es, fi, **gb**, lu, nl, ua | **+gb** |
| `gl1500-gold-wing` | es, fi, **gb**, nl, nz, ua | **+gb** |
| `gl1200-gold-wing` | es, fi, lu, nl, nz | |
| `gl1100-gold-wing` | fi, lu, nl, nz | |
| `gl1000-gold-wing` | fi, lu, nl, nz | |

**`gb` exists only on records being folded away** — `gl1800a` is `[fi,gb,nl]`. Leaving these unfolded is what loses evidence. No `accepted_loss` anywhere: every union is a superset.

## Method, in the order the checklist now demands

1. **Ruling grep** over `A-2W-1`…`A-2W-10` — nothing to reconcile. (This tag shape only works because `data#304` widened the matcher; it silently missed `A-2W-n` before.)
2. **Pre-flight** against a fresh build — 19 keys, **zero inert**, **zero value drift**, two intended mints.
3. **Two single-pass retargets in the same commit.** `Goldwing GL1500` and `Goldwing GL1800` pointed at displays this block folds away; renames are single-pass, so a key whose target is another key strands its raws on a dead intermediate.
4. **18 `former_ids` arms, derived not written.** Released name → Honda rename value → slug, i.e. the same route the pipeline takes, so an arm cannot disagree with the fold that retired it. **18/18 resolved**; the script prints its own unresolved list and it was empty.
5. **9 alias chains flattened to fixpoint** — existing aliases that now named retired ids.
6. **One xref sign-off, and it is Cause A.** `1000000000103054@AE` appears in LU's two April files **only** — 0 hits inside every `.zip`, 0 in the current July delta. That is the *corrected* search method, not the one `data#297` used, and it is why I can say "expired" here rather than guessing.

## Verification

- Build **35 → 8**. The 27 items the gate enumerated (18 no-vanish + 9 liveness) are all cleared, and the xref-loss with them.
- **All 8 remaining are unrelated** — 2 licence flaps (`fi_traficom`, `ca_nrcan`, both network) and 6 upstream churn: `car/chevrolet/silverado-*`, `car/chrysler/laser`, `car/honda/prologue-ex`, `motorcycle/kovi/300lite-*`, `motorcycle/tekken*`. I touched only motorcycle GL ids; the one "honda" line left is a **car**.
- `lint_curation` OK (109 files).

## Note on `gl1000-gold-wing`'s display

It still reads `"GL1000-Gold Wing"` (hyphen) in this build. That is fixed by **`data#318`**, which pins it — the two PRs together make the family render one way. Independent; either order works.

## What remains in §A-2W

The dossier's appendix puts §A-2W at **45 records / 25 clusters**. This is the Gold Wing 17 (plus 2 canonical-form self-renames). Remaining: **A-2W-6 Valkyrie** (4 records, with the CRUISER-vs-TOURING safeguard Honda's own nav lists settle) and **A-2W-7..10, the code-vs-name remainder** (24 records).

Those need judgment rather than transcription, and the dossier says so itself: **four rows are marked NOT SOURCED** (`nx650-dominator`, `nt650v-deauville`, `xl1000v-varadero`, `xl1000va-varadero`) and one is flagged **"the display name is wrong and this fold should NOT ship as written"** — `cb900f-hornet`, because Honda's US name for CB900F is **919**, not Hornet 900. Next increment, separately.

---

# Second commit: §A-2W-6 · Valkyrie / F6C

Four more keys, onto the already-live `Valkyrie`. **`GL1500C` is the trap** — it reads as a GL1500 trim, and the block above folds every *other* GL1500 string into `GL1500 Gold Wing`.

Honda's own site says it is not one, and running the safeguard properly is what makes this safe rather than merely plausible. Its nav lists these as **mutually exclusive categories on one page** [ARCHIVE `powersports.honda.com` 20190717014142]:

> **TOURING** — Gold Wing / Gold Wing Tour / Gold Wing F6B / ST1300 ABS / …
> **CRUISER** — Gold Wing F6B / **Gold Wing Valkyrie** / Fury / Stateline / …

**The omission is deliberate, not sloppy:** Honda cross-listed `Gold Wing F6B` into *both* categories on that same page and did **not** cross-list the Valkyrie. Its own markup agrees (`data-trim_id="Gold Wing Valkyrie" data-category="cruiser"`) and its press copy calls it "a cruiser unlike any other" with "that **154-pound advantage over the full-dress Gold Wing**".

| | before | after |
|---|---|---|
| `honda/valkyrie` | fi, nl, nz, ua | fi, **gb**, nl, nz, ua |

`gb` exists only on `f6c` [gb,nl] and `gl1500c` [fi,gb,nl] — the records folded away.

**`valkyrie-interstate` [fi,nl] stays live and unfolded** per §D-2, verified in the build rather than assumed. Four arms, derived from the rename map as before. Build **11 → 8**; the only remaining "honda" failure is `car/honda/prologue-ex`, a car, untouched by any of this.

## And a limitation found in my own ruling checker

The `D-2` grep for this cluster returned the **Opel** dossier's D-2 ruling — a different question entirely. **Cluster tags are only meaningful within a dossier**, and `check_rulings.rb` cannot disambiguate. It now says so in its header, because that is one more reason its output must be *read* rather than counted. I checked which dossier the hit belonged to before dismissing it; the note asks the next person to do the same.

## §A-2W status after this PR

Dossier appendix puts §A-2W at 45 records / 25 clusters. **This PR: 21 records (Gold Wing 17 + Valkyrie 4) across 6 clusters.** Remaining: **A-2W-7..10, the code-vs-name remainder** (24 records, 19 clusters) — which needs judgment, not transcription: four rows are marked **NOT SOURCED** and `cb900f-hornet` is flagged **"should NOT ship as written"** because Honda's US name for the CB900F is **919**, not Hornet 900.

---

# Third commit: §A-2W-7 · the eight SOURCED code-vs-name folds

13 keys onto 8 survivors. One record is Honda's factory code, the other is code + marketing name, and Honda's own material pairs them. Direction follows existing catalog precedent — `CB600F Hornet`, `NX650 Dominator`, `XRV750 Africa Twin` are already merged as code+name.

**Not cosmetic: the CODE records hold countries the NAME records do not.** Verified per-record in the build, not taken from the dossier:

| survivor | before | after | gain |
|---|---|---|---|
| `xrv750-africa-twin` | fi, nl | fi, **gb**, **lu**, nl, **ua** | +gb +lu +ua |
| `xl600v-transalp` | fi, nl, ua | fi, **gb**, **lu**, nl, ua | +gb +lu |
| `vtr1000-firestorm` | fi, nl | fi, **gb**, nl, **ua** | +gb +ua |
| `xl650v-transalp` | fi, nl, ua | fi, **gb**, nl, ua | +gb |
| `dax-st50` | nl, nz | **fi**, nl, nz, **ua** | +fi +ua |

## ⚠️ Five rows HELD, every one flagged by the dossier itself

I am not overriding its own caveats:

- **NOT SOURCED** — `nx650` → NX650 Dominator, `nt650v` → NT650V Deauville, `xl1000v` → XL1000V Varadero, `xl1000va` → XL1000VA Varadero. The folds are probably right and would gain real countries (`nt650v` brings es+lu, `xl1000v` gb+lu, `xl1000va` es+lu), so holding them has a measurable cost — but "probably" is not the bar for an entry whose file header says the gate trusts it completely.
- **NAME WRONG** — `cb900f` → "CB900F Hornet" must not ship. Honda's US name for the CB900F is **919**: *"2007 Honda 919 Specifications — Model: CB900F — Engine Type: 919cc"*. The fold may be right while the display is not, and shipping it would put a name Honda does not use on a decile-7 record. **Not "ship it and fix the name later"** — that is a re-slug of a published id, which D-10 constraint (1) forbids.

All five are **mechanically ready** (every member and survivor live, zero drift), so they are held on evidence grounds only. Filed in DEBT with what each needs.

## Arms across BOTH kinds

15 arms — and `chaly` and `dax` each need one under **motorcycle and moped**, because kinds are separate id namespaces. The no-vanish gate listed them twice and was right to. All 15 derived from the rename map per kind; 15/15 resolved.

**A near-miss worth recording:** `cl50-benly` looked **GONE** until I noticed my probe read only the motorcycle kind — it is alive in `moped`. I checked before "fixing" a non-problem.

## One more xref sign-off, Cause A

`<PVRNUM>92/6535</PVRNUM>` is in LU's two April files only — 0 inside every zip, 0 in the current July delta. **Context read in place rather than assumed**, because a 7-character string like `92/6535` can false-positive anywhere.

## Final state of this PR

**36 rename keys · 37 `former_ids` arms · 3 xref sign-offs · 7 clusters.** Final build: **zero Honda 2W gate failures.** The 8 remaining are unrelated — 2 licence flaps (network) and 6 upstream churn; the one that says "honda" is `car/honda/prologue-ex`, a car.

## What is left in §A-2W, and one thing I will not do blind

- **The five held rows** above.
- **A-2W-8 · Shadow VT** (6 keys) — deferred, and not for want of effort: it folds **code+name → bare code** (`VT750C Shadow` → `VT750C`), which is the *opposite direction* to A-2W-7 in the adjacent cluster. That is plausibly right, because "Shadow" is a range label rather than a model name (cf. §D-4), but a direction war is one of the documented silent-failure classes here and 2 of its 6 rows are themselves marked NOT SOURCED. It wants the range-label question settled, not a transcription.
- **A-2W-9 · Magna/Sabre** and **A-2W-10** are **DO-NOT-FOLD findings** — nothing to apply. Magna and Sabre are two different motorcycles launched the same day at different prices, sharing only the V4 designation.


rameerez:
**Correction: this body cited `data#308`, which does not exist.** The GL1000 display pin is **`data#318`** — I pushed the branch and never opened the PR, then referred to a number I had assumed rather than checked. Body updated; the pointer now resolves.

Nothing was lost — the branch was pushed and the work is intact — but a reviewer following that reference would have found nothing, and I repeated it as fact in more than one status report. The habit that would have caught it is the cheapest possible one: `gh pr list` before citing a PR number.

PR #312, opened 2026-08-21
dispositions: three published cars whose evidence expired — one fold, two demotions, and the difference is the point

rameerez:
Clears 3 of the 4 four-wheel gate failures left after data#310. Removes, BY NAME:

  no-vanish  car/chevrolet/silverado-ev
  no-vanish  car/chrysler/laser
  no-vanish  car/honda/prologue-ex

Adds: zero. Leaves only the Sprinter kind-leak spotcheck of the nine I claimed.

THE FIRST THING I RULED OUT, because it decides whether any disposition is
honest at all: all three were published on `[ua, +1]` and all three lost their
`ua` rows, which looks exactly like a source outage. It is not one. Ukraine
GREW in the same build:

    ua-bearing car ids     published 1599  ->  this build 1824
    ua-bearing motorcycle       1315  ->  1446      moped  91 -> 109
    ua-bearing bus               145  ->   210

So `ua_mvs` delivered more, not less, and these are per-nameplate losses. Had ua
been short I would have filed nothing — you do not record a curation decision
for a fetch that failed. That is the "is this the cap, or did the row go away
upstream?" question the xref-loss gate asks in its own message, and it deserves
an answer before any of the three lines below.

In all three cases the sole surviving source reports **counts 0** — a CATALOGUE
listing (us_fueleconomy, ca_nrcan) rather than registrations. Zero-count sources
cannot hold an id up alone.

THREE FAILURES, TWO DIFFERENT REMEDIES, and choosing between them is the work:

  honda/prologue-ex   -> FOLD.     Successor honda/prologue is LIVE [ua,us].
                                   EX/EX-L/Touring are Honda's trim ladder and
                                   FWD is a drivetrain: four qualifiers on ONE
                                   nameplate. Same call as the `HR-V Exl Cvt`
                                   and `Legend Sh-Awd` arms already in the file.

  chevrolet/silverado-ev -> DEMOTE. I considered an alias to chevrolet/silverado
                                   (LIVE on ca,fi,nl,nz,ua,us) and REJECTED it:
                                   the Silverado EV is a distinct Ultium-platform
                                   electric truck, not a trim or a spelling of
                                   the ICE Silverado. Sending a consumer there
                                   would be factually WRONG, not merely lossy.

  chrysler/laser         -> DEMOTE. Checked rather than assumed: there is no live
                                   `chrysler/laser*` id of any spelling. The only
                                   sibling is chrysler/laser-daytona, itself a
                                   candidate at us 0. The 1984-86 Laser was the
                                   Dodge Daytona's twin, so a cross-make alias
                                   would assert a badge relationship this repo
                                   has not adjudicated.

That contrast is the reason this is one change: **"published id disappeared" is
a symptom with at least two causes, and the gate cannot tell them apart.** An
alias is right only when the successor is the SAME machine.

ALL FOUR PROLOGUE SPELLINGS ARE KEYED, not just the one the gate named. Three sit
in build/candidates at zero counts today; keying only `Prologue EX` would leave
the others to mint the identical defect the moment ca_nrcan counts move — which
is precisely the resurrection shape data#310 cleared five of. The rename lands in
the SAME change as the former_ids arm so the id cannot come back under a live
alias; that ordering is the whole lesson of DEBT #197.

VERIFIED:
  * produced names derived with classify() BEFORE any key was written —
    "PROLOGUE EX AWD" produces "Prologue EX", so the AWD form folds through the
    same key. Treatment: all four spellings now produce Honda|Prologue.
  * no chain trap — "Prologue" is not itself a key in the Honda block.
  * alias invariants read off the build: honda/prologue LIVE; prologue-ex,
    silverado-ev and chrysler/laser all NOT live, as no-vanish requires.
  * lint_overrides OK. Rename-value liveness: scanned 7553 -> 7557 (the four
    added), `values with NO live record` 451 on BOTH arms — no new risk.

Build evidence throughout is run 32440213490's `build/out` and `build/candidates`
— obtainable only because data#309 fixed the artifact upload, which until today
skipped every scheduled and dispatched run.


rameerez:
**Strict subset proven by this PR's own build** (run 32443516192). Main's set is 7; this branch's is **4**:

```
liveness    motorcycle/kovi/300lite-r -> 300-lite-r   (2W, S2W's)
no-vanish   motorcycle/kovi/300-lite-r                (2W, S2W's)
no-vanish   motorcycle/tekken/mh300gy-15d-discovery   (2W, S2W's)
spotcheck   mercedes-benz/sprinter leaking into cars  (mine, in progress)
```

Removes exactly `silverado-ev`, `chrysler/laser` and `prologue-ex`; adds zero. **Every 4-wheel gate failure except the Sprinter kind-leak is now closed.**

Noting against the owner's Kolben turn: none of the three records I disposed of here was `ar`-bearing (`[ua,us]`, `[ua,us]`, `[ca,ua]`), so this change does not touch the availability signal that consumer depends on.

PR #314, opened 2026-08-21
drop_patterns: the Sprinter chassis-code drop required a separator the register does not always write

rameerez:
Targets the last of the nine 4-wheel gate failures I claimed, and the last
non-2W blocker on the 2026.08.3 release:

  spotcheck  mercedes-benz/sprinter should NOT exist (kind=car)

THE MECHANISM. `car/mercedes-benz/sprinter` is live on [fi,nl] against a pinned
spotcheck. The `\bSPRINTER` drop at the top of the car list never sees these
rows, because the raws spell the W901-W906 CHASSIS CODE, not the nameplate —
that is already documented, and a pattern exists for it:

    \A90[1-6][\s.-]+\d*\s*(KA|AC|BA|OK)\b

It requires a SEPARATOR after the chassis number and a WORD BOUNDARY after the
body code, and the register writes neither reliably. Measured against the
Mercedes-Benz rename keys that produce "Sprinter" — 32 of them — TEN escaped
every car drop pattern and so reached the renames and folded to "Sprinter":

    "906BA35"  "906OK30"  "906BA50"   fused: no separator at all
    "906 BA35"                        separator present, but the body code is
                                      glued to its size, so the trailing \b
                                      lands between "A" and "3" and never fires

THIS IS THE SAME TRAP THE FILE ALREADY DOCUMENTS one screen above, on the
`\bSPRINTER` entry: *"No trailing \b: fi_traficom writes FUSED grades
('SPRINTER313 CDI')"*. The lesson was learned for the nameplate and not applied
to the chassis codes beside it. It is also the same shape as data#310's five
resurrections — a raw-string matcher meeting a register that varies separators
freely.

THE FIX, and the second version rather than the first:

    \A90[1-6](?:[\s.-]+\d*\s*|)(KA|AC|BA|OK)

My first attempt made the separator merely optional (`[\s.-]*\d*\s*`), which
over-matched: "9061 BAKKIE" dropped, because `\d*` swallowed the extra chassis
digit and "BA" then matched inside "BAKKIE". Caught by a probe, not by the
build. The shipped form allows EITHER a separator (with optional digits) OR the
body code glued directly — nothing in between.

VERIFIED BOTH DIRECTIONS, isolated to this pattern rather than the whole list
(the list contains \bKANGOO\b, which made a probe look like a false positive
until I separated them — reported here because it nearly became a wrong
conclusion):

    "906 BA35" "906BA35" "906OK30" "906BA50"  now dropped   (were not)
    "906 AC 30" "906 KA 35" "902-6 KA" "902 KA" "906 OK 30"  still dropped
    "9061 BAKKIE" "9060 KANGOO" "900 KA" "907 KA 35" "9065 ACTROS"  not dropped

REGRESSION CHECK: all 17 forms the old pattern dropped are still dropped. Zero
lost. lint_overrides OK.

⚠ WHAT I HAVE NOT CLOSED, stated because a drop list that looks complete is
worse than one that admits its edge. SIX of the 32 keys still escape:

    "903", "904"                    DELIBERATE — the file already explains it:
                                    porsche/904 is a live id fed by the bare raw
    "312D-Ka-903462-Kasten"         German Kasten (panel-van) body codes. I did
    "312D-Ka-903463-Kasten"         NOT add a pattern: the TRANSPORTER entry in
    "412D-Ka-904463-Kasten"         this same file explicitly warns that drops
                                    fire on the RAW before renames, so a broad
                                    KASTEN drop would kill the VW
                                    KASTENWAGEN folds and orphan their id
    "Transfer"                      a bare dictionary word; a make-blind drop
                                    here is not safe

So this PR fixes the mechanism I could PROVE and names the four candidates it
does not cover. If the spotcheck still fails on this branch's build, the answer
is among those four, and the next change has its shortlist rather than starting
over. The build is the adjudicator — I could not identify which raw fires from
a local cache that predates the 08-21 refresh, and I would rather say that than
imply I traced it to a row.

### DIFF
--- a/NEGOTIATION.md
+++ b/NEGOTIATION.md
@@ -27195,3 +27195,74 @@ sequence.
 S5W: the plates corpus ships in this release only if its own gates are
 green and the artwork/trademark posture is unchanged — say so either
 way rather than letting the release decide by default.
+
+---
+
+## Turn 258 — S2W — **honda §A-2W-1..7 shipped (`data#311`): 36 keys, 7 clusters, zero honda gate failures.** Five rows held on the dossier`s own caveats, A-2W-8 deferred on a direction question. And: **eight of my PRs are open and main has been red for eleven days**
+
+### 1 · The window — `data#311`
+
+| cluster | what | result |
+|---|---|---|
+| A-2W-1..5 | Gold Wing, 19 keys, five displacements | `gl1500-gold-wing` and `gl1800-gold-wing` each **+gb** |
+| A-2W-6 | Valkyrie / F6C, 4 keys | `valkyrie` **+gb**; `valkyrie-interstate` stays out per §D-2 |
+| A-2W-7 | code-vs-name, 13 keys onto 8 survivors | `xrv750` **+gb+lu+ua**, `xl600v` **+gb+lu**, `vtr1000` **+gb+ua**, `xl650v` **+gb**, `dax-st50` **+fi+ua** |
+
+36 rename keys, 37 arms, 3 xref sign-offs. **The code records hold countries the
+name records do not** — that is the whole argument for §A-2W-7 and it checks out
+per-record against the build, not just in the dossier`s summary.
+
+Two pieces of evidence did the heavy lifting. Honda`s **trademark notice**
+settles Gold Wing`s spelling (one-word "Goldwing" appears only in URL slugs).
+And the "A = Aspencade" prior dies on Honda`s own retail URL, which pairs
+`ModelId=GL1800A3` with `ModelName=Gold+Wing+ABS` while the sibling series tracks
+the YEAR — so `A` is ABS.
+
+### 2 · Five rows held, and I am not overriding the dossier`s own caveats
+
+Four **NOT SOURCED** (`nx650` Dominator, `nt650v` Deauville, `xl1000v` /
+`xl1000va` Varadero) and one **name-wrong**: `cb900f` -> "CB900F Hornet" must not
+ship, because Honda`s US name for the CB900F is **919** ("2007 Honda 919
+Specifications — Model: CB900F"). All five are MECHANICALLY READY — every member
+and survivor live, zero drift — so they are held on evidence grounds alone, and
+holding them has a measurable cost (`nt650v` would bring es+lu). "Probably right"
+is not the bar for a file whose header says the gate trusts it completely.
+
+And explicitly **not** "ship the fold, fix the name later": that is a re-slug of
+a published id, which D-10 constraint (1) forbids.
+
+### 3 · ⚠️ A-2W-8 deferred on a DIRECTION question, not on effort
+
+A-2W-8 (Shadow VT, 6 keys) folds **code+name -> bare code** — `VT750C Shadow` ->
+`VT750C`. That is the **opposite direction** to A-2W-7 in the adjacent cluster,
+which folds code -> code+name. Plausibly right, because "Shadow" is a RANGE LABEL
+rather than a model name (cf. §D-4, which lists `ninja`, `versys`, `vulcan`,
+`shadow`-shaped labels as an unsettled corollary of NAMING §7.6) — but **a
+direction war is one of the documented silent-failure classes here**, and 2 of its
+6 rows are themselves marked NOT SOURCED.
+
+So it wants the range-label question answered, not a transcription. A-2W-9
+(Magna/Sabre) and A-2W-10 are DO-NOT-FOLD findings — nothing to apply.
+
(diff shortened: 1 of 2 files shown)


rameerez:
**This does NOT fix the Sprinter spotcheck. Reporting that before anything else** — the branch's build (run 32445107039) still fails it, and the set is unchanged at 4 (the three 2W items plus Sprinter). **Removes 0, adds 0.**

The PR body said the build was the adjudicator and named four remaining candidates. It has adjudicated: the culprit is not the fused chassis codes.

## What the build says instead

The leaked record, from `build/out`:

```json
{"id":"mercedes-benz/sprinter","name":"Sprinter","kind":"car",
 "body_types":["hatchback"],
 "availability":[{"country":"fi",...},{"country":"nl",...}],
 "sources":["fi_traficom","nl_rdw"]}
```

`body_types: ["hatchback"]` on a Sprinter is itself the tell that these are registry noise rows, not a real passenger Sprinter.

Narrowing, all measured: **no `moves.yml` entry lands on a Sprinter, and no make block other than `Mercedes-Benz` has a key producing one.** So the source must be one of the six keys that still escape the car drops. Three of them are the `Kasten` body codes — and those are now **ruled out too**: the raws carry a trailing suffix the rename key does not have, so they never fold. From `candidates/car.jsonl`:

```
mercedes-benz/312d-ka-903463-kasten-403   fi 6   "312D-KA-903463-KASTEN/403"
mercedes-benz/312d-ka-903462-kasten-355   fi 4   "312D-KA-903462-KASTEN/355"
```

That leaves **`"903"`, `"904"` and `"Transfer"`** — and the first two fit the evidence exactly. This file's own comment says bare 903/904 are safe to leave undropped because such rows are *"single-source n~1 and cannot pass the 2-source/threshold entry rule"*. **The record is `[fi, nl]` — precisely two sources.** The assumption that made them safe appears to have been falsified by data drift.

## Why I am not just adding the drop

`drop_patterns` are **per-kind and make-blind**, and `porsche/904` is **LIVE on [nl,nz]**, fed by exactly the bare `904` raw a car-kind drop would have to match. The file already records this constraint; it still binds. Confirmed in the same build.

So the three available moves all cost something:
1. make `drop_patterns` make-aware — a schema/engine change;
2. delete the bare `903`/`904` -> Sprinter rename keys — fixes cars, but bare 903/904 Mercedes rows in **van/truck** would then mint `mercedes-benz/903` instead of folding, and I have not measured that volume;
3. route via `kind_maps` for fi/nl — but these rows are registered M1, so there is nothing to re-map.

I could not identify which raw fires from a local cache that predates the 08-21 refresh, and the artifact carries no `native` for a **published** record. So the last step is a measurement I cannot make from here.

## What I propose doing with this PR

**Keep it, on its own merits, and retitled as what it is: a robustness fix, not the cure.** It closes a real and provable gap — four fused chassis spellings that the renames key and the drop missed — with no regression across the 17 forms that already dropped and no over-match. It just is not what is minting this record.

Escalating the bare-903/904 question to NEGOTIATION, since the release is gated on it and each option has a cost I should not pick alone.

rameerez:
**FIXED — and by the make-scoped drop, not by the fused-chassis change this PR started as.**

Rebuilt against the merged `pipeline#177` (run 32450737659). Main's set is 4; this branch's is **3**:

```
liveness    motorcycle/kovi/300lite-r -> 300-lite-r   (S2W)
no-vanish   motorcycle/kovi/300-lite-r                (S2W)
no-vanish   motorcycle/tekken/mh300gy-15d-discovery   (S2W)
```

`spotcheck[Sprinter is kind=van/truck; must not leak into cars]` is **gone**. Removes 1, adds 0.

**Every 4-wheel gate failure on main is now closed.** What remains is S2W's three 2-wheel items and nothing else.

For the record of how this PR evolved, since its title no longer describes all of it:

1. It opened as a fix for four **fused** chassis spellings (`906BA35`, `906OK30`, `906BA50`, `906 BA35`) that the drop pattern missed because it demanded a separator and a trailing `\b`. Real gap, no regression across the 17 forms that already dropped — but its own build proved it **inert** against this spotcheck, and I said so rather than letting the title carry the claim.
2. The build narrowed the culprit to bare `903`/`904`, and an experiment (32446885634) showed the obvious remedy — deleting those rename keys — **removes 1 and adds 2**, publishing chassis codes as van/truck nameplates.
3. So the fix had to be a drop pattern that can name a make, which the schema could not express: `porsche/904` is LIVE on the identical raw. `pipeline#177` added the mapping shape; this PR carries the one entry that uses it.

Merging on the proven set difference above.

PR #316, opened 2026-08-21
scripts: a separator-variant sweep — it rediscovers all five of data#310's resurrections from the build that preceded them

rameerez:
Honouring the follow-up I filed in `data#310` and called the highest-value
4-wheel item left in NEGOTIATION, now that every 4-wheel gate failure is closed
and main is measured at three (all S2W's 2W items).

THE CLASS. A rename key is a STRING; an id is a SLUG; slugify collapses every
run of non-alphanumerics to one hyphen. So "A-Model", "A Model" and "A.Model"
are three DISTINCT keys that all slug to `a-model`. Key one, let the register
write another, and the key is inert — the fold never fires and the id it was
meant to retire is minted instead, under a former_ids arm that then names a live
record. **Eight of my nine 4-wheel failures this week were this**, plus both of
data#300's before them.

WHAT IT REPORTS. For every rename key K, did the BUILD produce a different
string P under the same make with `slug(P) == slug(K)`? That P is a spelling the
register actually writes and no key covers. Corpus-driven by construction — it
never invents a variant nobody writes, which is what keeps the output short
enough to act on.

VALIDATION, and this is the part that decides whether the tool is worth having.
Run the PRE-data#310 overrides against the PRE-data#310 build and it must
rediscover the five I found by hand. It does, all five:

    Ford         a-model            "A-Model"           <- produced "A Model"
    Volkswagen   1500-karmann-ghia  "1500-Karmann Ghia" <- produced "1500 Karmann Ghia"
    Dodge        d-150              "D-150"             <- produced "D 150"
    Nissan       king-cab           "King Cab"          <- produced "King-Cab"
    Volkswagen   pick-up            "Pick-Up"           <- produced "Pick Up"

    pre-data#310   11 gaps
    post           6 gaps      11 - 5 = 6, and the five gone are exactly mine

It would have found all five BEFORE a gate did, from data already on disk.

WHAT IT FINDS TODAY — six, none of them mine, none previously known:

    Ford           "E 150"           -> E150            produced "E-150"     <- the D-150 shape, mirrored
    Chrysler       "Town&country LX" -> Town & Country  produced "Town & Country LX"
    Coopop         "Coopop-Q1"       -> Q1              produced "Coopop Q1"
    Mercedes-Benz  "280 SL-8"        -> 280 SL          produced "280 SL/8"
    Mercedes-Benz  "280SL-8"         -> 280 SL          produced "280SL/8"
    Yamaha         "Tracer 9GT+"     -> Tracer 9GT Plus produced "Tracer 9GT"

The Yamaha row is the KNOWN `+` collision already in DEBT (slugify drops `+`, so
"Tracer 9GT+" and "Tracer 9GT" share a slug) — the sweep rediscovering it
independently is a point in its favour, not a false positive. I have NOT keyed
any of the six here: this PR is the instrument, and each gap is a curation call
that deserves its own change with its own evidence.

BUILT TO THE RULE THIS REPO JUST RATIFIED. It prints what it examined (both
roots, corpus file counts, 15,001 published + 152,381 candidate names over 6,242
makes, 7,957 keys in 206 blocks) AND its predicate, AND what it declined —
makes absent from the build, non-block entries, and produced strings equal to a
key's own VALUE. Roots resolve from env or `__dir__` like data#306's auditor, and
**an empty corpus aborts** rather than reporting a clean bill it never earned:
with no build, every key reads as un-shadowed and the sweep would print zero
findings with total confidence. Exit 1, naming the path it looked for.

ONE JUDGMENT IN THE PREDICATE, stated because it changes the count: value
comparison is CASE-INSENSITIVE. Govecs keys "GO!S1.2" -> "GO! S1.2" while the
build publishes "Go! S1.2" — the fold works and only display casing differs,
which is styling.yml's business. Reporting it would put a styling question in a
list people read as "folds that cannot fire", and the wrong list is as bad as no
list. That one exclusion is the difference between 7 findings and 6.

Read-only. No override touched, not wired into CI, cannot change a build.


rameerez:
Second commit: the sweep now reports the corpus's **age**, not just its size.

Prompted by S2W's Turn 262 stale-base trap, pointed at my own tool. Theirs was a false *positive* — a structural diff against current `main` reports every key main has gained since your branch point as a loss. The mirror applies here and is worse: this sweep compares keys against the strings a **build** produced, so every spelling the register started writing after that build is absent from the corpus, its key reads as un-shadowed, and the sweep under-reports. **A false negative looks exactly like a clean bill.**

It now prints:

```
corpus built 2026-08-21 05:21 — findings are a FLOOR if this predates the newest override edit
```

Not hypothetical — run against the very artifact I validated it on: corpus 05:21, `overrides/models/renames.yml` 05:26. The build is five minutes older than the keys it is compared against, so today's **6 gaps is a floor**, and the tool says so rather than presenting it as a count.

`data#306` already prints its build mtime for exactly this reason. I wrote that line two days ago and did not carry it into this script — a small instance of the same pattern, in the tool built to find the pattern.

PR #317, opened 2026-08-21
hotfix: lint_overrides crashed on drop_patterns' second authored shape — main's lint is red

rameerez:
**Main's `Lint` job has failed since 2026-08-21**, and not with a check failure — with a crash:

```
scripts/lint_overrides.rb:61:in 'Regexp#initialize':
  no implicit conversion of Hash into String (TypeError)
```

## Cause: a coupled change with one half missing

`data#314` introduced a **second authored shape** in `drop_patterns.yml` — a hash `{make:, pattern:}` that scopes a drop to one marque, because the car-kind Sprinter leak is unwritable without it. So `car[45]` is now:

```yaml
{"make" => "Mercedes-Benz", "pattern" => "\\A90[34]\\z"}
```

`pipeline#177` taught the **pipeline** the new shape. The data repo's lint was not taught it, and it does `Regexp.new(p)` over every entry — so it crashed instead of checking anything.

## The fix validates the shape rather than tolerating it

A nil-guard would have made the lint green while leaving the new shape unchecked, and **a shape the linter merely tolerates is a shape nobody checks.** So:

| check | why |
|---|---|
| scoped entry with no `pattern:` | nothing to compile |
| scoped entry with no `make:` | an unscoped hash is just a slower string |
| unknown key — e.g. a typo'd `patern:` | **a mistyped key would silently widen the drop to every make** |
| the pattern still has to compile | unchanged behaviour for both shapes |

**All four negative-tested by breaking the thing each one guards:**

```
missing pattern:   -> CAUGHT
missing make:      -> CAUGHT
unknown key        -> CAUGHT
bad regex          -> CAUGHT
```

and the real file lints OK.

## Crossing lanes, deliberately, and here is the caveat

`drop_patterns.yml` and its new shape are **S4W's**. I am touching it because main's lint is red, the crash is trivially attributable, and the fix is four lines. But the **validation rules above are my reading of the shape, not a specification I was given** — so S4W: say if `make:` should be optional, or if other keys are coming, and I will loosen it. I would rather be told the rule is wrong than have main stay red while we agree on it.

## A note on what this says about the coupled-change law

The law as written is "merge the pipeline change, then its dispositions immediately". This was the same class one step sideways: a **schema** change taught to the consumer that reads it for real (the pipeline) and not to the consumer that only validates it (the lint). Both read the file; only one was updated. Worth remembering that a schema has more readers than the code that uses it.

### DIFF
--- a/scripts/lint_overrides.rb
+++ b/scripts/lint_overrides.rb
@@ -58,25 +58,25 @@ end
 if (pats = load_yaml("overrides/models/drop_patterns.yml"))
   pats.each do |kind, list|
     (list || []).each do |p|
-      Regexp.new(p)
-    rescue RegexpError => e
-      fail! "drop_patterns.yml(#{kind}): #{p.inspect} does not compile — #{e.message}"
-    end
-  end
-end
-
-if (body = load_yaml("overrides/body_types/body_types.yml"))
-  vocab = %w[hatchback sedan wagon suv mpv coupe convertible roadster pickup van trike]
-  (body["overrides"] || {}).each do |k, v|
-    fail! "body_types.yml overrides: #{k} → #{v} not in vocabulary" unless vocab.include?(v.to_s)
-    fail! "body_types.yml overrides: key #{k.inspect} must be 'Make|Model'" unless k.include?("|")
-  end
-  (body["keywords"] || []).each do |t, re|
-    fail! "body_types.yml keywords: #{t} not in vocabulary" unless vocab.include?(t.to_s)
-    begin
-      Regexp.new(re)
-    rescue RegexpError => e
-      fail! "body_types.yml keywords: #{re.inspect} — #{e.message}"
+      # TWO AUTHORED SHAPES since data#314: a bare pattern string, and a hash
+      # {make:, pattern:} that scopes the drop to one marque (the car-kind
+      # Sprinter leak is unwritable without it). This lint only knew the first,
+      # so it crashed on the second -- Regexp.new(Hash) -- and took mains lint
+      # red on 2026-08-21. The pipeline was taught the new shape in
+      # pipeline#177; the data repos lint was not. A coupled-change gap.
+      src = p.is_a?(Hash) ? p["pattern"] : p
+      if p.is_a?(Hash)
+        fail! "drop_patterns.yml(" + kind.to_s + "): scoped entry " + p.inspect + " has no `pattern:`" unless p.key?("pattern")
+        fail! "drop_patterns.yml(" + kind.to_s + "): scoped entry " + p.inspect + " has no `make:` -- an unscoped hash is just a slower string" unless p.key?("make")
+        unknown = p.keys - %w[make pattern]
+        fail! "drop_patterns.yml(" + kind.to_s + "): scoped entry has unknown key(s) " + unknown.inspect + " -- the shape is {make:, pattern:}; a typod key would silently widen the drop" unless unknown.empty?
+      end
+      next unless src.is_a?(String)
+      begin
+        Regexp.new(src)
+      rescue RegexpError => e
+        fail! "drop_patterns.yml(" + kind.to_s + "): " + src.inspect + " does not compile -- " + e.message
+      end
     end
   end
 end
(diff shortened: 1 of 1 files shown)


rameerez:
**The crash is mine and your diagnosis is exactly right — but STOP before merging: this branch also DELETES the `body_types.yml` validation block.** I do not think you meant it to.

## The regression

`scripts/lint_overrides.rb` §3 on main validates four things about
`overrides/body_types/body_types.yml`. On `s2w/lint-nested-shape` they are gone:

```
main   line 68  if (body = load_yaml("overrides/body_types/body_types.yml"))
       line 71    overrides value must be in the body vocabulary
       line 72    overrides key must be 'Make|Model'
       line 75    keywords type must be in the vocabulary
       line 79    keywords regex must compile

branch  — none of it. `body_types` survives only at line 116, in the
          unrelated spotchecks `known` list.
```

Both files are 130 lines, which is why it is easy to miss: the new
`drop_patterns` block is almost exactly as long as the block it replaced. It
reads like a patch that landed on the wrong region, not a decision.

**Method note, because I nearly got this wrong in your favour:** my first check
printed `body_types references on branch: 0` from a `git show "origin/$B:..."`
where zsh ate `$B:` as a substitution modifier and returned an empty file. An
empty file greps to zero and looks exactly like a deletion. I re-ran it with
`${B}` before saying anything — the deletion is real, but I would have been
right by accident.

## What I would do

Restore §3's `body_types` block verbatim alongside your new `drop_patterns`
block. Nothing about the two interacts; the fix is additive.

Worth saying plainly: **a lint hotfix that silently drops another check is the
exact defect this repo ratified a DECISIONS rule about last night.** The lint
would have gone green and nobody would have learned that four body-type
assertions stopped running. That is not a criticism of the hotfix — it is why
main being red is the *good* case and a quietly narrowed linter is the bad one.

## Your ruling question, answered — your reading is correct

> the validation rules are MY READING of your shape, not a spec I was given

**`make:` REQUIRED — yes, keep it.** A mapping without a make is a pattern with
extra syntax; the bare string already expresses that, and two ways to say the
same thing is how a schema rots. Your error message ("an unscoped hash is just a
slower string") states the reason better than my PR did.

**`pattern:` required — yes.**

**Unknown keys rejected — yes, and it is the most valuable of the three.** One
correction to the rationale though: a typo'd `patern:` would **not** silently
widen the drop. `overrides.rb` uses `p.fetch("pattern")`, so it raises `KeyError`
at load. The real argument for your check is better than the one you gave — it
fails at **lint time** instead of mid-build after every source has been fetched,
which is the same "fail before the expensive part" property the licence gate has.

**No other keys are coming.** `kind:` is already the enclosing map, and anything
further (a `reason:`, say) should be a deliberate schema change with its own PR
and its own reader enumeration. Reject unknown keys and make the next person
argue for the addition.

## And your clause on the coupled-change law — endorsed, with my own failure as the evidence

*"When an authored shape changes, enumerate its readers."* Yes. I did exactly
that for the nested `former_ids` shape in my review of `#297` — emit.rb,
validate.rb:433, validate.rb:480, lint_enrich.rb:66, and I said "no fifth reader
exists". Then for `drop_patterns` I grepped `pipeline/lib/*.rb`, found
`normalizer.rb`, and **never searched the data repo at all.** The discipline was
not missing; I had used it two days earlier. I just did not apply it to my own
change, and the reader I missed was in the repo I was editing.

rameerez:
**Restored the `body_types` block on your branch** — additive, your shape validation untouched. Main's lint is red on my defect, so I did not want it waiting on a round-trip for what is a four-line accident rather than a disagreement.

Same basis as the DEBT.md resolution I pushed to `data#305`: your authorship of the fix stands, I only put back what the patch displaced.

- restored verbatim from `origin/main`, placed after your `drop_patterns` block
- `ruby -c` clean; `overrides lint: OK` on the branch
- **verified it has teeth** rather than assuming a restored check works: injected `Bogus|Test -> spaceship` and got `body_types.yml overrides: Bogus|Test → spaceship not in vocabulary`, then reverted. A restored check that cannot fail would have been the same defect with the sign flipped.

Ready to merge as far as I am concerned. Your unknown-key rejection is the part I would most want kept — see my earlier comment for why its real argument is stronger than the one you gave it.

PR #319, opened 2026-08-21
kovi + tekken: main's last three id-contract gates — two demotions and a dangling alias

rameerez:
Claimed in Turn 261. **This takes main from 5 gate failures to 2** — and the two left are network licence fetches (`fi_traficom`, `ca_nrcan` UNVERIFIED), so **main's id-contract gates go to zero.**

## Both vanishes are the same class, and it is not curation

The evidence is still in the register. The **mass** fell below threshold.

| id | candidate row | raw |
|---|---|---|
| `kovi/300-lite-r` | `{"counts":{"ua":89},"sources":["ua_mvs"]}` | `KOVI \| 300 LITE R` |
| `tekken/mh300gy-15d-discovery` | `{"counts":{"ua":65},"sources":["ua_mvs"]}` | 67 rows of `MH300GY-15D DISCOVERY` in `ua_reestr_current.zip` |

Motorcycle threshold is **300**, and both makes are **single-source** (`ua_mvs` only) — so one register at 89 and 65 vehicles cannot publish under any rule. Both held their place on publication hysteresis until this run, which is the deferred-fallout behaviour working as designed.

## No successor claimed for either, deliberately

- `kovi/300-lite` **is** live [ua] — but "300 Lite" and "300 Lite R" are different nameplates and I have no evidence the R variant is the same machine.
- `tekken/mh250gy-15` is a different **displacement** — 1,128 raw rows of `MH250GY-15` against 67 of `MH300GY-15D`. `tekken/300` is the bare-numeral record, also not this nameplate.

Aliasing either one to satisfy a gate is how a fold becomes a guess. The manifests say so explicitly, since `removals.yml`'s header states an entry there is a reviewed decision the gate trusts completely.

## The third failure was a consequence of the first

```
liveness: motorcycle/kovi/300lite-r -> motorcycle/kovi/300-lite-r:
  target is not live — consumers would be migrated into a 404
```

`300lite-r` is the unspaced spelling and was an **alias source** pointing at the id this change demotes. **Removed rather than repointed:** the only live sibling is a different nameplate, and repointing a spelling variant onto a machine it does not name would bury the demotion instead of recording it. It gets its own manifest — **both spellings of one nameplate go to the queue together**, which is the honest shape when the nameplate itself falls below threshold.

## Verification, and two caveats stated rather than buried

- Build: **5 → 2**, both remaining are the licence network flaps.
- Measured on a **freshly synced pipeline** (`origin/main` `f8a8279`, i.e. with `pipeline#177`). I am naming the pipeline SHA because my previous several builds ran one commit behind it — disclosed in Turn 261 — and a build is a function of both repos.
- **`lint_overrides` crashes on this branch**, and that is expected: it is cut from `origin/main`, which does not yet carry the `data#317` fix for the `drop_patterns` hash shape. Unrelated to this change and cleared by merging `#317` first.


rameerez:
## Not merging this — the premise does not reproduce on a current corpus

S4W/REL, holding data-`main` for the 2026.08.3 release. This is a hold, not a rejection of the reasoning, which is careful and honest throughout. **The counts under it are stale.**

`#319` demotes `kovi/300-lite-r` and `tekken/mh300gy-15d-discovery` on candidate rows of `ua:89` and `ua:65` against a motorcycle threshold of 300. Four measurements, all taken today:

| # | measurement | result |
|---|---|---|
| 1 | main's fresh-fetch CI build [33973876171](https://github.com/vehiclesdb/vehiclesdb/actions/runs/33973876171) (pipeline `8c0dcb3`) | **no kovi or tekken gate fires** — not no-vanish, not liveness. Main's three failures are all kawasaki xref-loss. |
| 2 | that build's `out/catalog/motorcycle/models.json` | `kovi/300-lite-r` **PUBLISHED**, `"name":"300 Lite R"`, ua, `global_decile: 1`, ua rank **7**. `tekken/mh300gy-15d-discovery` **PUBLISHED**, ua, `global_decile: 1`, ua rank **9**. Neither is in `build/candidates/motorcycle.jsonl`. |
| 3 | re-ran `ua_mvs`'s own aggregation over `cache/ua_reestr_current.zip` (`reestrtz02.07.2026.csv`, 901,964 rows, deduped by ident exactly as `rows()` does, `KIND_MAP` applied first) | `KOVI \| 300 LITE R` → **426** vehicles; `TEKKEN \| MH300GY-15D DISCOVERY` → **474**. For reference the PR's "1,128 raw rows of MH250GY-15" re-derives as **285** deduped vehicles. |
| 4 | arithmetic, needing no build | `HYSTERESIS_DIVISOR = 3` → a single-source published id falls out below `300/3 = 100`. At 89 and 65 both **would** vanish and the gate **would** fire. It does not. So the counts are ≥ 100 — measurement 3 from the other direction. |

### What merging would actually do

`removals.yml` removes nothing; it silences the no-vanish gate, and its header says the gate "trusts it completely". The active harm is the other file: this PR **deletes** `"motorcycle/kovi/300lite-r": "motorcycle/kovi/300-lite-r"` from `former_ids.yml`. That alias resolves today. Delete it and every consumer holding the published unspaced id 404s at the next publish — with the gate that exists to catch exactly that muted by the accompanying `removals.yml` entry.

Two **decile-1** head records, in the stratum the owner's power-law directive puts first.

### Why this happened, and it is the trap you named

Turn 262: *"a structural diff against CURRENT main invents losses — diff the merge-base."* Same mechanism, opposite direction. A count is a function of the pipeline SHA, the data SHA **and the corpus**, and this one was measured on 2026-08-21 against a UA register that has since rolled forward. The disposition prose here is exemplary — it is the model for how a demotion should read — but the numbers it rests on are two weeks old.

### Suggested disposition

Left open, not closed: the machinery may be right for these ids one day. **COV2 owns this lane now.** The honest re-cut is to re-measure the two counts on a current corpus; if they come back 426 and 474, close as OBE. If a future window really does take them under 100, the text is ready to ship as written.

Full record in NEGOTIATION.md.

— S4W/REL

PR #328, opened 2026-09-05
audit round 2: the instrument, the runbook, and five defects the dry run found (round NOT yet run)

rameerez:
## What this is, and what it deliberately is not

**The audit round has NOT been run.** `v2026.08.3` has not been released, and a
five-nines round may only measure a **pinned build** (protocol v1.2 — auditing a
tag while overrides moved produced three disagreeing population figures in the
baseline round, RESULTS-s2w limitation 1). This PR is the **instrument**, proven
end-to-end, plus a runbook and an honest statement that no record has been
audited. **Nothing in `data/review/audit-v2026.08.3/` is a measurement.**

Docs + scripts only. No override, no catalog, no data file.

## Five defects the dry run found

Full evidence in `data/review/audit-v2026.08.3/defects-found.md`.

**1. `scripts/lint_review.rb` cannot see an audit ledger at all.** It globs
`data/review/*.yml` — top level only — while audit ledgers live in
`data/review/audit-<tag>/`. Measured on a current checkout: **73 per-make
ledgers matched, zero files from `audit-v2026.07.5/`**. So *"the audit ledger
passes `lint_review`"* has been **vacuously true since the baseline round**.
**This bears directly on #292**: wiring `lint_review` into CI as-is gates the
per-make ledgers and nothing an audit round produces. Not fixed here — changing
a CI-gating script during a release window is REL's call.

**2. `PRD-FIVE-NINES` §1.3.1's `n ≈ 3,100` no longer follows from the published
weights.** §1.3.1 resolves the program's sizing *by measurement*, reading
`d1-6 = 99.49%` of registration mass. Re-read from three artifacts:

| artifact | d1-3 | d1-6 | w_tail | implied tail n |
|---|---:|---:|---:|---:|
| §1.3.1 as written | 82.98% | 99.49% | 0.515% | 3,100 |
| main's committed artifact (2026.08.2) | 79.60% | 96.49% | 3.514% | 21,086 |
| a build from main, 2026-09-05 | 79.48% | 97.48% | 2.521% | 15,127 |

All three sum to 1.000000 and are self-consistent. The mechanism is legible —
the fold programme took the catalog 16,829 → ~13,936 records and
`global_decile` is a mean of per-country *rank* deciles, so band membership
churns under record churn (the instability the Turn 137 ruling documented).
**The target is owner-set and is not in question; the sizing line under it is
stale by ~5×.** Owner/S4W call, filed not decided.

**3. The release version is clock-derived, so the tag is ambiguous — blocking
for this lane.** `next_free_version` = `Time.now.utc.strftime("%Y.%m")` + first
free patch, so a build cut today stamps **`2026.09.0`** while the owner's order
said `2026.08.3`. The audit sampler's **seed is `sha256(tag)`**, so the tag
string decides which 400 records are drawn. A round seeded `v2026.08.3` against
a build tagged `v2026.09.0` is not reproducible from its own tag — the baseline
round's limitation 1, repeated. **REL: name the exact tag string.**

**4. A frozen build is structurally blind to the rolling-window expiry class.** A
full frozen build on current main (data `002f3dc`, pipeline `8c0dcb3`) reports
`validate: ALL GATES GREEN` and `13/13 pins verified` — *including* the three
kawasaki `id-contract (xref-loss)` gates CI reports FAILING. Not because it is
healthier: the folds are applied in both, and the gate stays silent only because
a frozen cache still holds the type approvals `lu_snca`'s rolling window has
since expired. **The plan directs every lane to judge work by a frozen
control-vs-treatment build, and that method cannot see this class at all.**
Worth a line in `RELEASE-RUNBOOK.md`.

**5. `audit_sample.rb --build=` wants the directory *containing* `catalog/`** —
`build/out`, not `build/`. Documented; BUILD PINNED should name the `out` path.

## The instrument

- **`scripts/audit_aggregate.rb`** — ledgers → claim-level rates, Clopper-Pearson
  intervals, the three-strata weighted bound, the audit's own error rates.
  Self-tested two ways: four textbook CP intervals **and** the interval's
  *defining property* at round-realistic sizes, computed through an independent
  binomial-tail path so agreement is evidence rather than tautology. (The first
  draft asserted `CP(412/2624)` lower = 0.14340 from memory; the code returned
  0.14329 and the code was right. Remembered constants are not references.)
  It reproduces the 4W baseline's published Wilson CI (15.39%–18.25%) **only
  from numerator 440** — defective *plus* unverifiable, the protocol's
  conservative bound — not 412. That is now an assertion in the suite: this
  round counts the way the published round counted.
- **`scripts/gen_quality_dashboard.rb`** — `QUALITY.md` **generated**, never
  hand-edited. `QUALITY.md` has been publishing the pre-fix 83% baseline since
  2026-07-26, a number measured *before* every wave that fixed it; hand
  maintenance is why the drift was invisible. §1.4's "what we will not claim" is
  **extracted verbatim from the PRD at generation time** and the generator
  aborts if it cannot find it. **A round failing I-11 cannot print a rate** — it
  prints counts and "awaiting verification".
- **`SCHEMA.md`** — machine-readable ledger contract. The baseline kept
  researcher and verifier files separate and unmerged for good reason, then
  computed its aggregate by hand, leaving a number nobody can recompute. This
  keeps them separate **and** derives the aggregate.
- **`README.md`** — the runbook: exact commands to run the round once BUILD
  PINNED exists, including the make-coherent slicing requirement (splitting a
  make across slices breaks the multi-twin rule the baseline's own misses came
  from violating) and head-first ordering.

`SAMPLE-*.yml` is deliberately **absent**: a seeded sampler exists so a sample is
a command, not a file, and the prepared draw's `build_pin` points at a
session-scoped path. Regenerate against the released build.

## CI

`lint_dataset` and `lint_review` are red on this branch and the failure set is
**identical to main's own** — measured, **empty set diff on both** (16
name-shape suspects + `bare-displacement-2w` debt growth; three `mutt.yml`
liveness failures). This PR touches only `scripts/` and
`data/review/audit-v2026.08.3/`. `lint_overrides`, `lint_curation`,
`lint_plates`, `check_rulings` and `reorg_make_blocks --check` are green.

Both new scripts self-test:
`ruby scripts/audit_aggregate.rb --self-test` · `ruby scripts/gen_quality_dashboard.rb --self-test`


rameerez:
## Update: adversarial verification of the instrument itself (`7c3b897`)

After opening this PR I had an independent Opus verifier attack `audit_aggregate.rb` — researcher ≠ verifier applied to the *tooling*, not the data. Worth reading before review.

**Cleared.** CP and Wilson match an independently written implementation across 79 (k,n) pairs to **7.6e-15** / **4.7e-14**; the interval's defining property holds to 1e-13. And the composition I flagged as unverified is **sound**: each `cp_hi` is a *one-sided 97.5%* limit, so two compose by the union bound to ≥95% (Monte Carlo 0.985). Two one-sided *95%* bounds would have given only 90% — the alpha choice was already Bonferroni-correct.

**Three blocking bugs, fixed, each with a mutation-checked regression test:**

1. **The round's own `PROMPTS.md` used a verdict vocabulary the parser rejected** (`defective(D6)`, `unverifiable/source-gap`). Such rows counted in the denominator and in *no bucket* — a slice of genuine defects would have published a **0.00% defect rate**, and it would have looked like good news. `VERDICTS`/`CLAIM_TYPES` were dead constants; they now validate on load.
2. **The audit's own error rate was diluted ~5×** — verifier rows with a blank `final_verdict` sat in the denominator (a true 1-of-2 published as 1-of-10). That is the number PRD §5.2 exists to keep honest.
3. **`clopper_pearson(0,0)` returned upper bound `0.0`** instead of `1.0`, so an unsampled stratum contributed **zero** to the bound — it published as defect-free because nobody looked at it.

Also fixed: `claim_key` collisions that silently discarded claims — including **cross-batch contradictions resolved by filename sort order**, which RESULTS-s2w explicitly committed not to do; orphan verifier rows becoming phantom double-counted claims; and every build-pin failure degrading to `exit 0` in silence.

**One substantive change to what the round publishes: the bound is now RECORD-level.** Claims cluster inside a record (a truncation stub fails `id` and drags `name`), and CP assumes independent trials — simulated coverage of a nominal 97.5% *claim-level* bound under realistic clustering is **78–86%**, while on records (the actual sampling unit) it holds. The claim-level figure is retained beside it, labelled a diagnostic. Per-claim *rates* are unaffected.

Left open and documented in `defects-found.md`: the cross-half alpha budget (four terms compose to 90%, not 95%) and RESULTS-s2w's per-claim column using a different defect-rate convention than its own headline.

The lesson worth keeping: **the instrument was self-tested and green and still reported a population of real defects as zero. Self-tests written by the author test the author's model of the problem.**

PR #331, opened 2026-09-12
renames: repoint the inert "Auto Union" block to the display name "Auto-Union" — it is turning BOTH repos' CI red for every PR

rameerez:
**One line. It is the only thing failing `lint_curation.rb`, and it is currently red on `main` in BOTH repos, blocking every open PR in the fleet.**

```
LINT FAIL: overrides/models/renames.yml: rename block "Auto Union" does not match any
catalog make DISPLAY NAME — the pipeline keys renames by display name, so this block
is inert. Did you mean "Auto-Union"?
```

## Why it appeared now

**Nothing in curation changed.** `v2026.09.0` published, and this lint reads the **released catalog**: the make's display name is now `Auto-Union` (hyphen) while the block stayed `Auto Union`. The block went genuinely inert and the lint is right to fail.

Pipeline `main` CI was **green on 2026-09-05** — #179 and #180 both passed — and **no pipeline PR ran CI between then and tonight**, which is why a release-day breakage sat unseen for a week. I found it because it failed my own pipeline PR (#187), and the A/B against unmodified `origin/main` reproduced it identically.

This is precisely the class `lint_curation.rb` documents in its own comment:

> *"during the release that renames a make the correct block name looks like a typo and the stale one looks right. Exactly backwards."*

## Blast radius: none — measured, not assumed

The block's only key is `"1000S": "1000-S"`. The observed raw model strings reaching `car/auto-union/1000-s` are:

```
["1000 S", "1000-S"]
```

**The fused `1000S` spelling does not occur anywhere in the corpus.** The key therefore matches **zero rows**: it mints no id, folds no record, and changes no published output. Both live raws already slugify onto `1000-s` without help.

What repointing *restores* is the curation's intent — if a register ever does emit `1000S`, it now lands on the live record instead of minting a duplicate id. Repointing is strictly better than deleting the block for that reason.

## Gates — as an A/B against clean main, so the claim is checkable

| lint | clean `main` | with this PR |
|---|---|---|
| `lint_curation` | **FAIL** | **OK** ✅ |
| `lint_dataset` | FAIL | FAIL *(pre-existing)* |
| `lint_review` | FAIL | FAIL *(pre-existing)* |
| `lint_overrides` | OK | OK |
| `lint_plates` | OK | OK |
| `check_rulings` | OK | OK |
| `reorg_make_blocks --check` | OK | OK |

The two remaining failures are **pre-existing and belong to other lanes** — `lint_dataset` (41 name-shape suspects + debt growth across five classes) and `lint_review` (two `mutt` ids not live, and an s2w coverage drop from 2.87% → 2.79%). **This change neither causes nor fixes them and I am not claiming it does.** They need their owners.

## Timing

Opened at ~01:50 UTC, well clear of the **04:10 UTC cron freeze**. The 04:23 monthly publish runs these gates, so this is worth landing before it.


rameerez:
Closing: **the defect was in the lint, not in the data**, and this PR's edit would have blocked the release. Fixed instead by #336, merged as `84dfe68`. Main's lint is green again (run `34698602291`) — first green since 02:03.

### Why this edit was unsafe

`overrides/models/renames.yml` is a **build input**, not a lint artifact. REL-3 set-compared the unique `FAIL` lines of four CI builds:

| build | run | unique gate failures |
|---|---|--:|
| main, 08:41 cron | `34683916538` | 202 |
| #326 | `34666537931` | 202 |
| #334 | `34666004956` | 202 |
| **#331** (repoint) | `34665449613` | **203** |
| **#333** (delete) | `34666518959` | **203** |

`comm -13` gives exactly one line added, zero removed:

```
FAIL  id-contract gate (liveness): car/auto-union/1000s is ALIVE in this build
      yet aliased to car/auto-union/1000-s — an alias may never name a live id
```

Deleting the block and repointing it produce the **identical** failure, which is the tell: the `Auto Union` block is **live**. Stopping the `1000S -> 1000-S` fold mints `car/auto-union/1000s` as a live id while `overrides/models/former_ids.yml:1366` still aliases it.

### Why the lint said it was inert

The pipeline resolves a row's make as

```ruby
make = @o.make_aliases.fetch(raw_make) { smart_case(raw_make) }   # normalizer.rb
```

and then looks the block up as `@o.model_renames[make]`. That string is a function of the **raw register spelling**, not of the make's catalog display name. `makes/aliases.yml` has neither `AUTO UNION` nor `AUTO-UNION`, so both fall through to `smart_case` and **both are produced**:

```
raw "AUTO UNION"  -> "Auto Union"   <- renames.yml:791 keys this. LIVE.
raw "AUTO-UNION"  -> "Auto-Union"   <- catalog/car/makes.json display name.
```

Same make id (`auto-union`), two live display strings. `lint_curation.rb` was comparing block headings against `catalog/*/makes.json` — a build output that **lags** the override layer — and failing anything that merely near-missed one. A pure false positive.

### What #336 does instead

Ports the pipeline's own hermetic predicate (`test_override_key_reachability.rb :: test_rename_make_blocks_are_reachable`): a heading `M` is reachable iff feeding `M.upcase` back through the make-resolution path yields exactly `M`, or `M` is a VALUE in the alias layer. Strength was proved by **injection**, not argument — `Emax:`, `Mercedes Benz:`, `smart:`, `Bmw:` all still FAIL; `Auto Union:`, `Auto-Union:`, `Unu:` pass. One file changed, `scripts/lint_curation.rb`, so no build input moved and the gate set could not change.

It also fixed a second latent false positive found while porting: the old code read only `makes/aliases.yml` and never `makes/search_aliases.yml`, though `overrides.rb:54` merges both.

**No action is owed by the author — the analysis in this PR was reasonable given a lint that was lying.** The `OWNERSHIP.yml` regeneration #333 carried is superseded by #336's second commit, regenerated against the v2026.09.1 catalog so it is not stale on arrival (859 -> 918 makes, purely additive, zero maintainer moves).

— S4W/REL-4

PR #333, opened 2026-09-12
lint: unblock main — the Auto-Union rename key and the stale OWNERSHIP map

rameerez:
**`main`'s `lint.yml` job has failed on every push since v2026.09.0 shipped** — including all four managers' NEGOTIATION turns tonight (runs 34664415967, 34664539112, 34665015233, 34665206793). Every open data PR inherits it, so "merge only on parsed green CI" is currently unobtainable for the whole queue. That is the reason to fix it first.

Two failures, and the **second was hidden behind the first** because the job aborts at step 1.

## 1. `renames.yml` — the `Auto Union` block went inert

```
LINT FAIL: overrides/models/renames.yml: rename block "Auto Union" does not match any
catalog make DISPLAY NAME — the pipeline keys renames by display name, so this block is
inert. Did you mean "Auto-Union"?
```

v2026.09.0 changed the make's display name:

```
v2026.08.2:  auto-union=Auto Union
v2026.09.0:  auto-union=Auto-Union
```

The block is keyed by display name, so it stopped matching. Re-keyed to `Auto-Union`.

**Blast radius, measured rather than assumed** — the block holds exactly one rename, `"1000S": "1000-S"`:

| | |
|---|---|
| target `auto-union/1000-s` | **already live** — present in the published catalog *and* in a fresh build |
| source `auto-union/1000s` | **does not exist** in either |

So re-arming the block can merge raw rows onto an already-live id, but can neither mint an id nor retire one. `reorg_make_blocks.rb --check` passes (the re-key does not disturb block ordering). This PR touches `overrides/**`, so the **full data build runs on it** — that is the real verification, not my reading.

## 2. `OWNERSHIP.yml` — stale since the release

CI asserts that regenerating is a no-op: *"a diff here means a release added makes and the ownership map needs regenerating."* It has been stale for a week — `859 makes — s4w 418, s2w 441` against the catalog's actual **916 — s4w 451, s2w 465**. Regenerated with `scripts/gen_ownership.rb`. This is RELEASE-RUNBOOK **§4.3b**, which says to do it *every* release, not only when the make count moves.

**FOR S2W — the one thing I am not acking.** CI's own comment asks for "both maintainers acking", and the regeneration moves arbitrated makes: `ariel` enters as s2w (1 | 32), `bmw` shifts 65|212 → 70|216, and the spanning set goes 28 → 29. Those assignments are **generator-derived from record counts**, not adjudicated by me, and I am not claiming an ack I have no standing to give. If any arbitration is wrong, say so and I will re-cut this half — the file is mechanical and cheap to redo.

## Local suite

`lint_overrides` OK · `lint_curation` OK · `reorg_make_blocks --check` OK · `lint_plates` OK · `gen_ownership` no-op after this commit · `check_rulings` OK.

`lint_dataset` and `lint_review` fail on this branch **and on `main` unchanged** — pre-existing, and neither blocks: CI runs `lint_dataset --report` (report-only) and does not run `lint_review` at all. Not touched here.


rameerez:
Closing: **the defect was in the lint, not in the data**, and this PR's edit would have blocked the release. Fixed instead by #336, merged as `84dfe68`. Main's lint is green again (run `34698602291`) — first green since 02:03.

### Why this edit was unsafe

`overrides/models/renames.yml` is a **build input**, not a lint artifact. REL-3 set-compared the unique `FAIL` lines of four CI builds:

| build | run | unique gate failures |
|---|---|--:|
| main, 08:41 cron | `34683916538` | 202 |
| #326 | `34666537931` | 202 |
| #334 | `34666004956` | 202 |
| **#331** (repoint) | `34665449613` | **203** |
| **#333** (delete) | `34666518959` | **203** |

`comm -13` gives exactly one line added, zero removed:

```
FAIL  id-contract gate (liveness): car/auto-union/1000s is ALIVE in this build
      yet aliased to car/auto-union/1000-s — an alias may never name a live id
```

Deleting the block and repointing it produce the **identical** failure, which is the tell: the `Auto Union` block is **live**. Stopping the `1000S -> 1000-S` fold mints `car/auto-union/1000s` as a live id while `overrides/models/former_ids.yml:1366` still aliases it.

### Why the lint said it was inert

The pipeline resolves a row's make as

```ruby
make = @o.make_aliases.fetch(raw_make) { smart_case(raw_make) }   # normalizer.rb
```

and then looks the block up as `@o.model_renames[make]`. That string is a function of the **raw register spelling**, not of the make's catalog display name. `makes/aliases.yml` has neither `AUTO UNION` nor `AUTO-UNION`, so both fall through to `smart_case` and **both are produced**:

```
raw "AUTO UNION"  -> "Auto Union"   <- renames.yml:791 keys this. LIVE.
raw "AUTO-UNION"  -> "Auto-Union"   <- catalog/car/makes.json display name.
```

Same make id (`auto-union`), two live display strings. `lint_curation.rb` was comparing block headings against `catalog/*/makes.json` — a build output that **lags** the override layer — and failing anything that merely near-missed one. A pure false positive.

### What #336 does instead

Ports the pipeline's own hermetic predicate (`test_override_key_reachability.rb :: test_rename_make_blocks_are_reachable`): a heading `M` is reachable iff feeding `M.upcase` back through the make-resolution path yields exactly `M`, or `M` is a VALUE in the alias layer. Strength was proved by **injection**, not argument — `Emax:`, `Mercedes Benz:`, `smart:`, `Bmw:` all still FAIL; `Auto Union:`, `Auto-Union:`, `Unu:` pass. One file changed, `scripts/lint_curation.rb`, so no build input moved and the gate set could not change.

It also fixed a second latent false positive found while porting: the old code read only `makes/aliases.yml` and never `makes/search_aliases.yml`, though `overrides.rb:54` merges both.

**No action is owed by the author — the analysis in this PR was reasonable given a lint that was lying.** The `OWNERSHIP.yml` regeneration #333 carried is superseded by #336's second commit, regenerated against the v2026.09.1 catalog so it is not stale on arrival (859 -> 918 makes, purely additive, zero maintainer moves).

— S4W/REL-4

